PR #14, opened 2022-01-20
chore: enable "lint" in CI checks

layershifter:
This PR enables `lint` command (or target in terms of NX) in CI pipelines.
Don't know how I overlooked it initially.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,6 +5,7 @@
   "overrides": [
     {
       "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
+      "excludedFiles": "*.fixture.js",
       "rules": {
         "@nrwl/nx/enforce-module-boundaries": [
           "error",
--- a/packages/react/src/RendererContext.tsx
+++ b/packages/react/src/RendererContext.tsx
@@ -16,14 +16,7 @@ export interface RendererProviderProps {
  * Verifies if an application can use DOM.
  */
 function canUseDOM(): boolean {
-  return (
-    typeof window !== 'undefined' &&
-    !!(
-      window.document &&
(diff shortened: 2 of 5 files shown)


.eslintrc.json
```diff
@@ -5,6 +5,7 @@
   "overrides": [
     {
       "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
+      "excludedFiles": "*.fixture.js",
```
layershifter, line 8: Fixtures should not be checked with `@nrwl/nx/enforce-module-boundaries` as it forces to use relative imports.

packages/react/src/RendererContext.tsx
```diff
@@ -16,14 +16,7 @@ export interface RendererProviderProps {
  * Verifies if an application can use DOM.
  */
 function canUseDOM(): boolean {
-  return (
-    typeof window !== 'undefined' &&
-    !!(
-      window.document &&
-      // eslint-disable-next-line deprecation/deprecation
```
layershifter, line 23: Not used `eslint-disable-next-line`

packages/react/src/makeStaticStyles.ts
```diff
@@ -3,7 +3,7 @@ import { makeStaticStyles as vanillaMakeStaticStyles } from '@griffel/core';
 import { useRenderer } from './RendererContext';
 import type { GriffelStaticStyles, MakeStaticStylesOptions } from '@griffel/core';
 
-export function makeStaticStyles<Selectors>(styles: GriffelStaticStyles | GriffelStaticStyles[]) {
```
layershifter, line 6: Unused generic

ling1726 (approved):

PR #20, opened 2022-01-25
docs: init docusaurus, add sample pages

layershifter:
This PR inits new `docs` application powered by Docusaurus.

- adds few sample pages
- routing is configured
- Nx project (`apps/docs/project.json`) was manually configured as [@nx-plus/docusaurus](https://github.com/ZachJW34/nx-plus/tree/master/libs/docusaurus) does not work with Nx 13 yet


apps/docs/project.json
```diff
@@ -0,0 +1,22 @@
...
+    "serve": {
+      "executor": "@nrwl/workspace:run-commands",
+      "options": {
+        "cwd": "apps/docs",
+        "commands": [{ "command": "docusaurus start" }]
+      }
+    }
+  },
+  "tags": []
+}
```
layershifter, line 22: @Hotell I tested both `build` & `serve` and it works. Anything missing?

apps/docs/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "@tsconfig/docusaurus/tsconfig.json",
+  "compilerOptions": {
+    "baseUrl": "."
+  }
+}
```
layershifter, line 6: Not sure that it's a good thing, but it's tsconfig suggested by docusaurus...

packages/react/README.md
```diff
@@ -164,8 +164,8 @@ import { makeStyles } from '@griffel/react';
 
 const useClasses = makeStyles({
   root: {
-    ':global(html[data-whatintent="mouse"])': { background: 'yellow' },
-    // outputs: html[data-whatintent="mouse"] .abcd { background: yellow }
+    ':global(html[data-whatintent="mouse"])': { backgroundColor: 'yellow' },
```
layershifter, line 167: Oops, we should not advertise shorthands in docs

apps/docs/docs/react/install.mdx
```diff
@@ -0,0 +1,27 @@
+---
```
layershifter, line 1: `apps/docs/docs/react/install.mdx` double "docs" in path does not sound nice. May be rename the project to `site`?

apps/docs/.eslintrc.json
```diff
@@ -0,0 +1,18 @@
+{
+  "extends": ["../../.eslintrc.json"],
+  "ignorePatterns": ["!**/*"],
+  "overrides": [
+    {
+      "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
+      "rules": {}
```
ling1726, line 7: is leaving `rules` empty just setting no lint rules at all ?

apps/docs/babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  presets: [require.resolve('@docusaurus/core/lib/babel/preset')],
```
ling1726, line 2: was the deep import preset auto generated ?

apps/docs/docs/react/install.mdx
ling1726: `docsite` ?

apps/docs/babel.config.js
layershifter: Yup

apps/docs/.eslintrc.json
layershifter: - it's for overrides, so does not affect base preset/rules
- it's autogenerated by Nx

apps/docs/docs/react/install.mdx
```diff
@@ -0,0 +1,27 @@
...
+sidebar_position: 1
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+# Install
+
+By default Griffel contains only runtime version, it's enough to install it and it works:
```
ling1726, line 10:
```suggestion
By default Griffel is a runtime css-in-js engine, you can simply install and use it in code directly:
```

This might be worded a bit clearer ?

apps/docs/docs/react/make-styles.md
```diff
@@ -0,0 +1,192 @@
...
+  color: blue;
+}
+.fnbrw4x:nth-child(2n) {
+  background-color: #fafafa;
+}
+```
+
+### `:global()` selector
+
+Another useful feature is `:global()` selector, it allows connecting local styles with global selectors.
```
ling1726, line 99:
```suggestion
Another useful feature is `:global()` selector, it associates local styles with global selectors.
```

Not really sure if this is better WDYT ?

apps/docs/docs/react/make-styles.md
```diff
@@ -0,0 +1,192 @@
...
+});
+```
+
+```css
+html[data-whatintent='mouse'] .f1xz3i88 {
+  background-color: yellow;
+}
+```
+
+### At-rules
```
ling1726, line 117: nit: these's no example output for at-rules

apps/docs/docs/react/make-styles.md
```diff
@@ -0,0 +1,192 @@
...
+
+`@media` & `@supports` queries are also supported:
```
ling1726, line 119: nit: should we have example output for @media and @support query example you have here ?

layershifter: Yeah, will add it 👍

ling1726 (approved):

layershifter:
Going to merge it, we can address issues/apply fixes in follow up PRs.

PR #22, opened 2022-01-27
docs: styling tweaks

layershifter:
## Color palette update

image

⬇️⬇️⬇️

image

## Fonts

Segoe UI fonts are used for text.

## Prism theme 

Now uses VSCode themes.

image

⬇️⬇️⬇️

image

## "code output" divider

image

⬇️⬇️⬇️

image

## Theme controls/Github link

image

⬇️⬇️⬇️

image


apps/website/src/components/OutputTitle/index.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import styles from './styles.module.css';
```
layershifter, line 2: These should `makeStyles()` call, but Docosaurus doesn't support CSS-in-JS hooks for SSR setup: https://github.com/facebook/docusaurus/issues/3236

apps/website/src/theme/Toggle/index.js
```diff
@@ -0,0 +1,97 @@
...
+      <GithubIcon />
+    </a>
+  );
+}
+
+export default function () {
+  return (
+    <div className={styles.container}>
+      <ThemeToggle />
+      <GithubLink />
```
layershifter, line 94: `GithubLink` should be just an item in navbar, but there is no way other way to React components to Navbar currently...

apps/website/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "@griffel/website",
```
layershifter, line 2: Sometimes docusaurus checks for updates and if `package.json` it fails any CLI command.

ling1726: not sure I understood your comment

apps/website/src/theme/Toggle/index.js
ling1726: Maybe rename the folder ? this isn't a `Toggle` component anymore is it ?

apps/website/package.json
layershifter: Yeah, missed a few words there, sorry. 

- Docusaurus has update notifier
- It reads `package.json` inside project

https://github.com/facebook/docusaurus/blob/770418f8d262c56cd3de5ba1087f972ab20da617/packages/docusaurus/bin/beforeCli.js#L84-L88

- If it's absent, it throws on any CLI command (`docusaurus build`, `docusaurus start`, etc.)

apps/website/src/theme/Toggle/index.js
layershifter: As I understood Docusaurus checks for folder existence and then replaces a component (_as this folder is registered nowhere_). If I will rename it, I will get default controls back.

ling1726: ah ok, now I understand

ling1726: Let's throw a comment there, so this does not get lost. Is there a docusaurus issue about the navbar ?

apps/website/src/theme/Toggle/index.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import { useColorMode } from '@docusaurus/theme-common';
+import useIsBrowser from '@docusaurus/useIsBrowser';
+import styles from './styles.module.css';
+
+const SunIcon = () => (
```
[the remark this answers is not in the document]
layershifter: Yup, moved them out

apps/website/src/theme/Toggle/index.js
layershifter: Added a comment. I failed to find an issue/feature request to link it 😿

apps/website/package.json
ling1726: small nit: you could leverage the `description` property to doc that requirement

layershifter: Hmm, good idea, pushed

ling1726 (approved):

PR #23, opened 2022-02-04
chore: Add storybook to @griffel/react

miroslavstastny:
Add storybook to @griffel/react to be used for dev inner loop.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
         name: Checkout [main]
         with:
           fetch-depth: 0
+          ref: ${{ github.ref }}
 
       - name: Derive appropriate SHAs for base and head for `nx affected` commands
         uses: nrwl/nx-set-shas@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,7 @@ jobs:
         with:
           fetch-depth: 0
           ref: ${{ github.ref }}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
 
       - name: Derive appropriate SHAs for base and head for `nx affected` commands
         uses: nrwl/nx-set-shas@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,8 +14,6 @@ jobs:
         name: Checkout [main]
         with:
           fetch-depth: 0
-          ref: ${{ github.ref }}
-          repository: ${{github.event.pull_request.head.repo.full_name}}
 
       - name: Derive appropriate SHAs for base and head for `nx affected` commands
         uses: nrwl/nx-set-shas@v2
@@ -48,6 +46,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
           fetch-depth: 0
 
       - name: Derive appropriate SHAs for base and head for `nx affected` commands
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,0 +1,11 @@
+module.exports = {
+  stories: [],
+  addons: ['@storybook/addon-essentials'],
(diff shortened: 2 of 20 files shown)


packages/react/.babelrc
```diff
@@ -1,6 +1,6 @@
 {
   "plugins": ["annotate-pure-calls"],
-  "presets": ["@babel/preset-react"],
+  "presets": ["@babel/preset-react", "@babel/typescript"],
```
miroslavstastny, line 3: Added to support typescript in storybook.
Probably not the correct way, will continue looking into this.

packages/react/.storybook/main.js
```diff
@@ -0,0 +1,25 @@
...
+
+  core: { ...rootMain.core, builder: 'webpack5' },
+
+  stories: [...rootMain.stories, '../src/**/*.stories.mdx', '../src/**/*.stories.@(ts|tsx)'],
+  addons: [...rootMain.addons, '@nrwl/react/plugins/storybook'],
+  // MX: requires tsconfig.json in repo root
+  // typescript: {
+  //   check: true,
+  //   configFile: 'tsconfig.base.json'
+  // },
```
miroslavstastny, line 14: Cruft, remove before merge.

packages/react/.babelrc
miroslavstastny: Should I just use `@nrwl/react/babel` as in `test` for `development` or always?

miroslavstastny: Moved to storybook config.

.github/workflows/ci.yml
```diff
@@ -46,6 +46,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
```
layershifter, line 49: This change is needed to make pipelines passing on forks. Without it cloning fails.

package.json
```diff
@@ -28,9 +31,15 @@
     "@nrwl/linter": "13.4.5",
     "@nrwl/node": "13.4.5",
     "@nrwl/react": "13.4.5",
+    "@nrwl/storybook": "^13.7.2",
```
layershifter, line 34:
```suggestion
    "@nrwl/storybook": "13.4.5",
```

Can you please downgrade `@nrwl/storybook` to the same version? As the package uses strict dependencies this adds a lot of duplicates:

image

packages/react/src/ComponentStyles.stories.tsx
```diff
@@ -0,0 +1,48 @@
+import * as React from 'react';
```
layershifter, line 1: As stories does not have 1:1 mapping to files, can we move them to `/stories` subdirectory? I.e. `packages/react/src/ComponentStyles.stories.tsx` => `packages/react/src/stories/ComponentStyles.stories.tsx`

.storybook/main.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
ling1726, line 1: is there a need to put the storybook config in the root of the repo ? will there be more storybooks in this monorepo ?

packages/react/.storybook/main.js
```diff
@@ -0,0 +1,21 @@
+const rootMain = require('../../../.storybook/main');
```
ling1726, line 1: looking at this config, it isn't required ? since nothing custom is actually being done here

.storybook/main.js
miroslavstastny: I get your point, but I am not sure. The structure is how it was created by the nx generator. Do you think we should diverge from that?

package.json
miroslavstastny: Downgraded, but even with that, there are +6000 lines in yarn.lock :/

layershifter: Can you please also run `yarn dedupe`? (it's built into Yarn Berry)

.storybook/main.js
ling1726: Not a blocker for me I guess, if this is how NX gets things done

ling1726 (approved):

PR #26, opened 2022-02-08
chore: Add DOMRenderer filter story

miroslavstastny:
Add DOMRenderer story to demo `unstable_filterCSSRule`.


packages/react/src/stories/index.stories.tsx
```diff
@@ -0,0 +1,7 @@
+export { Default } from './makeStyles.stories';
+export { ComponentStyles } from './ComponentStyles.stories';
+export { DOMRendererFilter } from './DOMRendererFilter.stories';
+
+export default {
+  title: 'makeStyles',
+};
```
miroslavstastny, line 7: Added in order to sort stories.

layershifter (approved):

PR #35, opened 2022-02-17
docs: Add a try it out playground page

ling1726:
Integrates sandpack to add a very basic playground to try out `makeStyles`.

Uses a custom renderer outputs css rules as strings. Then highlight and prettier is applied for a nicer reading experience.

More improvements to come in future PRs


apps/website/src/components/Playground/index.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { Sandpack } from '@codesandbox/sandpack-react';
+import { useColorMode } from '@docusaurus/theme-common';
+import AppCode from '!!raw-loader!./template/App.js';
+import StylesCode from '!!raw-loader!./template/Styles.js';
+
+export default function Playground() {
+  const { isDarkTheme } = useColorMode();
+  const sandpackTheme = isDarkTheme ? 'sandpack-dark' : 'github-light';
```
layershifter, line 9:
```suggestion
  const sandpackTheme = isDarkTheme ? 'dark' : 'github-light';
```

`dark` theme looks more matching to ours IMO.

apps/website/src/components/Playground/index.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
```
layershifter, line 1: Let's convert this to be TypeScript code. I pushed changes in config to this branch.

P.S. It's my mistake that I didn't do this in previous PR.

apps/website/src/components/Playground/index.js
```diff
@@ -0,0 +1,19 @@
...
+  const { isDarkTheme } = useColorMode();
+  const sandpackTheme = isDarkTheme ? 'sandpack-dark' : 'github-light';
+  return (
+    <Sandpack
+      template="react"
+      theme={sandpackTheme}
+      files={{ '/App.js': { code: AppCode, hidden: true }, '/Styles.js': { code: StylesCode, active: true } }}
+      customSetup={{ dependencies: { '@griffel/core': 'latest', 'highlight.js': 'latest', prettier: 'latest' } }}
+      options={{ editorHeight: 450 }}
+    />
```
layershifter, line 17: Let's switch to more complex config (we will need to do this sooner or later):

```tsx
    <SandpackProvider
      template="react"
      customSetup={{
        dependencies: { '@griffel/core': 'latest', 'highlight.js': 'latest', 'js-beautify': 'latest' },
        files: { '/App.js': { code: AppCode, hidden: true }, '/Styles.js': { code: StylesCode, active: true } },
      }}
    >
      <SandpackLayout theme={sandpackTheme}>
        <SandpackCodeEditor />
        <SandpackPreview showOpenInCodeSandbox={false} />
      </SandpackLayout>
    </SandpackProvider>
```

P.S. This also removes "export to CodeSandbox" button as it's non sense in this case.

apps/website/docusaurus.config.js
```diff
@@ -50,6 +51,12 @@ const config = {
         //   src: 'img/logo.svg',
         // },
         items: [
+          {
+            type: 'doc',
+            docId: 'try-it-out',
+            position: 'left',
+            label: 'Try it out',
+          },
```
layershifter, line 59: Let's move this after "React" section as docs are more important :)

layershifter (commented):
Really cool improvement ❤️

apps/website/docusaurus.config.js
ling1726: I actually disagree though, I think having hands on code experience is better for docs... it's easier for ppl to have a good positive experience 

but this can also be personal biais

apps/website/src/components/Playground/template/App.js
```diff
@@ -0,0 +1,41 @@
...
+
+export default function App() {
+  const [rules, setRules] = React.useState('');
+  const ref = React.useRef(null);
+  React.useEffect(() => {
+    const playgroundRenderer = {
+      id: 'playground',
+      insertCSSRules(cssRules) {
+        const raw = Object.values(cssRules).flat().join('\n');
+        const prettified = beautify.css_beautify(raw);
```
layershifter, line 17:
```suggestion
        const prettified = beautify.css(raw, { indent_size: 2 });
```

nit: let's use "2" as indent size

typings/tsconfig.json
```diff
@@ -12,6 +12,9 @@
   "references": [
     {
       "path": "./environment/tsconfig.json"
+    },
+    {
+      "path": "./static-assets/tsconfig.json"
```
layershifter, line 17:
```suggestion
```

As we agreed in offline to perform TS conversion separately, let's also restore this.

apps/website/docusaurus.config.js
layershifter: That's fair, but if people don't know how to install the package they don't care what CSS it produces 🐱

ling1726: but should the install be on the main landing page ?

apps/website/src/components/Playground/index.js
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <SandpackProvider
+      template="react"
+      customSetup={{
+        dependencies: { '@griffel/core': 'latest', 'highlight.js': 'latest', 'js-beautify': 'latest' },
+        files: { '/App.js': { code: AppCode, hidden: true }, '/Styles.js': { code: StylesCode, active: true } },
+      }}
+    >
+      <SandpackLayout theme={sandpackTheme}>
+        <SandpackCodeEditor showLineNumbers customStyle={{ height: PLAYGROUND_HEIGHT }} />
```
ling1726, line 21: unfortunately without the `editorHeight`  option of the default `<Sandpack />` component, it's really the only way to make the playground bigger, editing the layout component above does not do anything

apps/website/docusaurus.config.js
layershifter: - Yes, if we are React only
- No, if we will support something else except React

---

For JSS that has similar feature in docs it's also not first:

image

apps/website/src/components/Playground/index.js
layershifter: For now IMO it's ok. In followup PRs it would nice to introduce autosizing/resizing feature.

apps/website/docusaurus.config.js
ling1726: alright, changed the order

layershifter (approved):

PR #47, opened 2022-02-22
feat: Strip unsupported CSS properties at development time

### LINKED ISSUE #29: Do not add CSS shorthands in runtime
We don't allow to use CSS shorthands:
https://github.com/microsoft/griffel/blob/b4c7acc77e5e7b4883ca32249418a3d93dc43e77/packages/core/src/types.ts#L5-L49

We should implement something in runtime to prevent adding rules with shorthands. Currently shorthands usage produces TS errors, but rules are still inserted to DOM. That creates confusion and feeling that something is wrong with typings.

- This should be based on typings that we have
- The runtime check should be development (non-production) to avoid contributions to bundle size

#### Expected behavior

```tsx
import { makeStyles } from "@griffel/react";

  root: {
    // ⏭ "background" is not inserted to DOM & a warning is produced (currently inserted)
    // 💥 produces TS error (already done)
    background: "red"
  }
});
```

ling1726:
Updates the typings so that we can keep a collection of constants for
unsupported CSS longhand properties.

Use this collection in `resolveStyleRules` to strip rules that include
unsupported properties at developemnt time.

Fixes #29


packages/core/src/runtime/resolveStyleRules.ts
```diff
@@ -54,6 +55,14 @@ export function resolveStyleRules(
 ): [CSSClassesMap, CSSRulesByBucket] {
   // eslint-disable-next-line guard-for-in
   for (const property in styles) {
+    // eslint-disable-next-line no-prototype-builtins
+    if (process.env.NODE_ENV !== 'production' && UNSUPPORTED_CSS_PROPERTIES.hasOwnProperty(property)) {
```
ling1726, line 59: Do we want the babel preset to strip out these styles even in production time ? currently babel preset will work like normal in production builds.

At runtime we would like to avoid bundle size so it's fine, but at build time there's really no reason to keep unsupported css properties in the bundle.

packages/core/src/runtime/resolveStyleRules.ts
```diff
@@ -54,6 +55,14 @@ export function resolveStyleRules(
...
+      if (process.env.NODE_ENV !== 'test') {
```
ling1726, line 60: should a logging utility be included to keep tests clean (and assert that logging was called ) ?

packages/core/src/runtime/resolveStyleRules.ts
```diff
@@ -54,6 +55,14 @@ export function resolveStyleRules(
...
+        console.error('Unsupported CSS property', property, 'will not be resolved');
+      }
+      continue;
```
miroslavstastny, line 63: I think that stripping the unsupported CSS properties only in dev mode is a bad idea - that will lead to inconsistent CSS between dev and prod.

The final CSS should be the same in all three modes - dev, prod and babel.
We can conditionally warn or throw but not strip.

ling1726: That's a fair point, I think that the code to do this (if I'm doing it correctly) shouldn't have a large bundle size impact on the library, which was the original motivation in the issue to do it only at dev time.

What is your appetite to strip unsupported css in all modes ?

miroslavstastny: > What is your appetite to strip unsupported css in all modes ?

Works for me.

layershifter: > I think that stripping the unsupported CSS properties only in dev mode is a bad idea - that will lead to inconsistent CSS between dev and prod.

If we will keep this in production it will increase bundle size. It's the original motivation to have it only in `devMode`.

packages/core/src/types.ts
```diff
@@ -1,58 +1,16 @@
...
-  outline?: never;
-  overflow?: never;
-  padding?: never;
-  placeItems?: never;
-  placeSelf?: never;
-  textDecoration?: never;
-  textEmphasis?: never;
-  transition?: never;
-};
+export type GriffelStylesUnsupportedCSSProperties = Record<keyof typeof UNSUPPORTED_CSS_PROPERTIES, never>;
```
layershifter, line 6: As we are now `csstype@3` this could be improved:

```ts
type GriffelStylesUnsupportedCSSProperties = Record<
  keyof CSS.StandardShorthandProperties | keyof CSS.VendorShorthandProperties,
  never
>;
```

We should use this instead of our types.

packages/core/src/runtime/resolveStyleRules.ts
ling1726: I've added a new runtime bundle size fixture to this PR: here are the results with/without production check

with production check:

image


without production check:

image

packages/core/src/types.ts
ling1726: well it depends if we want to enable css property stripping in all environments... because in that case we should probably follow another source of truth... I'll try it out and see how it goes tho

packages/core/src/runtime/resolveStyleRules.ts
layershifter: > Do we want the babel preset to strip out these styles even in production time ? currently babel preset will work like normal in production builds.
> At runtime we would like to avoid bundle size so it's fine, but at build time there's really no reason to keep unsupported css properties in the bundle.

Correct. These properties should be always stripped by Babel plugin, it should also emit warnings.

packages/core/src/types.ts
ling1726: I can type `UNSUPPORTED_CSS_PROPERTIES` with that type, the only exception is `gridGap` which is in `CSS.ObsoleteProperties`, should I include that in the type union ?

packages/core/src/runtime/resolveStyleRules.ts
ling1726: I've removed the production check so that we're consistently stripping out the unsupported styes. I've kept the bundle size fixture

packages/core/src/runtime/resolveStyleRules.ts
```diff
@@ -54,6 +55,14 @@ export function resolveStyleRules(
 ): [CSSClassesMap, CSSRulesByBucket] {
   // eslint-disable-next-line guard-for-in
   for (const property in styles) {
+    // eslint-disable-next-line no-prototype-builtins
+    if (UNSUPPORTED_CSS_PROPERTIES.hasOwnProperty(property)) {
+      if (process.env.NODE_ENV !== 'test' && process.env.NODE_ENV !== 'production') {
```
layershifter, line 60: Can you please clarify why we should not warn in tests?

packages/core/src/types.ts
layershifter: > I can type `UNSUPPORTED_CSS_PROPERTIES` with that type, the only exception is `gridGap` which is in `CSS.ObsoleteProperties`, should I include that in the type union ?

Yes, please. We can include `gridGap` manually

packages/core/src/runtime/resolveStyleRules.test.ts
```diff
@@ -11,6 +12,19 @@ function getFirstClassName([resolvedClassesForSlot]: [CSSClassesMap, CSSRulesByB
...
+      'strips unsupported `%s` css property when not in production',
+      property => {
+        // Doesn't matter what the value is, just that the resulting objects are empty
+        const res = resolveStyleRules({ [property]: 'dummy' as unknown as undefined });
+        expect(res).toHaveLength(2);
+        expect(res[0]).toEqual({});
+        expect(res[1]).toEqual({});
+      },
+    );
+  });
```
layershifter, line 26: Can you please add a fixture to Babel plugin too?

ling1726: Done, I added the code in JS because our types are working.... alternatively I could have used ts with `@ts-nocheck`

packages/core/src/runtime/resolveStyleRules.ts
ling1726: it was polluting tests, but I can disable console just for that specific test in griffel

packages/core/src/types.ts
ling1726: This also resulted in extra vendor shorthands... do we want them ?

layershifter: > This also resulted in extra vendor shorthands... do we want them ?

Let's skip for now (until someone complains) 😅

packages/core/src/runtime/resolveStyleRules.ts
layershifter: Yes, please 👍

packages/core/src/runtime/resolveStyleRules.test.ts
layershifter: Can we use TS and `ts-expect-error` instead?

packages/core/src/runtime/resolveStyleRules.test.ts
```diff
@@ -99,7 +122,6 @@ describe('resolveStyleRules', () => {
           '--fooBar': 'var(--barBaz)',
 
           backgroundColor: 'red',
-          MozAnimation: 'initial',
```
layershifter, line 102: Let's use another vendor property for this test.

ling1726: used `MozBorderLeftColors`

packages/core/src/runtime/resolveStyleRules.test.ts
ling1726: I had to use it for every single property ☹️

packages/core/src/types.ts
ling1726: OK I removed all the vendor shorthands

packages/core/src/runtime/resolveStyleRules.ts
ling1726: babel-preset tests are kinda awkward to mock console, any ideas ?

packages/babel-preset/__fixtures__/unsupported-css-properties/code.ts
```diff
@@ -0,0 +1,100 @@
+import { makeStyles } from '@griffel/react';
+
+export const useStyles = makeStyles({
+  root: {
+    // @ts-expect-error unsupported css properties should be stripped
+    all: 'value',
```
layershifter, line 6: I suggest to trim this file and keep only common properties (`margin`, `padding`, `border`) as technically it's just integration test for `resolveStyles()`.

packages/core/bundle-size/makeStylesRuntime.fixture.js
```diff
@@ -0,0 +1,7 @@
+import { makeStyles, mergeClasses } from '@griffel/core';
+
+console.log(makeStyles, mergeClasses);
```
layershifter, line 3: 👍 

P.S. We need to follow up on integrating bundle size reporting.

packages/core/src/constants.ts
```diff
@@ -19,3 +20,54 @@ export const LOOKUP_DEFINITIONS_INDEX = 0;
 
 /** @internal */
 export const LOOKUP_DIR_INDEX = 1;
+
+/** @internal */
+export const UNSUPPORTED_CSS_PROPERTIES: Record<keyof CSS.StandardShorthandProperties, number> = {
```
layershifter, line 25:
```suggestion
/** 
  * We use "1" as values to cause less effect on bundle size
  * @internal
  */
export const UNSUPPORTED_CSS_PROPERTIES: Record<keyof CSS.StandardShorthandProperties, 1> = {
```

Let's use `1` instead of `number` as we don't really want to allow anything except "1". And add a comment ;)

packages/babel-preset/__fixtures__/unsupported-css-properties/code.ts
ling1726: it's a good call, I limited to what I think of as 'common' shorthands

packages/core/src/runtime/resolveStyleRules.ts
```diff
@@ -54,6 +55,14 @@ export function resolveStyleRules(
 ): [CSSClassesMap, CSSRulesByBucket] {
   // eslint-disable-next-line guard-for-in
   for (const property in styles) {
+    // eslint-disable-next-line no-prototype-builtins
+    if (UNSUPPORTED_CSS_PROPERTIES.hasOwnProperty(property)) {
+      if (process.env.NODE_ENV !== 'production') {
+        console.error('Unsupported CSS property', property, 'will not be resolved');
```
layershifter, line 61: Let's make this error more human 🐱 

image

```suggestion
        console.error(
          [
            `@griffel/react: You are using unsupported shorthand CSS property "${property}". ` +
              `Please check your "makeStyles" calls, there *should not* be following:`,
            ' '.repeat(2) + `makeStyles({`,
            ' '.repeat(4) + `[slot]: { ${property}: "${styles[property as keyof GriffelStyle]}" }`,
            ' '.repeat(2) + `})`,
            '',
            'Learn why CSS shorthands are not supported: https://aka.ms/griffel-css-shorthands',
          ].join('\n'),
        );
```

layershifter (approved):
LGTM, please address the last comment about error message and we are good to go 🚀

packages/core/src/runtime/resolveStyleRules.ts
ling1726: it's always easy to claim short error messages are easy to understand when you are writing them 😄

packages/core/src/runtime/resolveStyleRules.ts
layershifter: We had an offline call about it, implemented in [1d8a02d](https://github.com/microsoft/griffel/pull/47/commits/1d8a02d1d848b7dab1c24a9ec9f3105692fd8d4f)

PR #48, opened 2022-02-22
docs: Convert docs to typescript

ling1726:


apps/website/src/components/Playground/index.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import { SandpackProvider, SandpackLayout, SandpackCodeEditor, SandpackPreview } from '@codesandbox/sandpack-react';
 import { useColorMode } from '@docusaurus/theme-common';
-import AppCode from '!!raw-loader!./template/App.js';
-import StylesCode from '!!raw-loader!./template/Styles.js';
+import AppCode from '!!raw-loader!./template/App.tsx';
```
ling1726, line 4: it's possible to keep the playground in JS, it was just easier to turn them into TS files too... but I'm open to doing extra work to leave it as JS too

layershifter: I briefly checked and it seems that sandpack does not emit TS errors. So may be there is sense to keep them in JS. At least it feels that `react-preset` loads faster :)

ling1726: that's fine, we can still type check internally with jsdoc for these files

ling1726: I guess if we ever get more complicated requirements for live code editing we'll have to investigate docusaurus live editing

apps/website/src/theme/Toggle/index.tsx
```diff
@@ -49,7 +49,7 @@ function GithubLink() {
 
 export default function () {
   return (
-    <div className={styles.container}>
+    <div className={styles['container']}>
```
ling1726, line 52: @layershifter  is there a reason the tsconfig you originally proposed bans `styles.container` ?

layershifter: Not really, AFAIR it's defaults from Nx setup

layershifter (approved):

PR #55, opened 2022-02-25
fix(babel-preset): remove implicit dependency on TypeScript preset

layershifter:
This PR removes implicit dependency on `@babel/preset-typescript` for module evaluation.
If it's used - it should be specified obviously.

### DIFF
--- a/packages/babel-preset/src/transformPlugin.test.ts
+++ b/packages/babel-preset/src/transformPlugin.test.ts
@@ -15,6 +15,11 @@ pluginTester({
       plugins: ['typescript'],
     },
   },
+  pluginOptions: {
+    babelOptions: {
+      presets: ['@babel/typescript'],
+    },
+  },
   formatResult: code =>
     prettierFormatter(code, {
       config: {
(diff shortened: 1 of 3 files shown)


packages/babel-preset/src/transformPlugin.test.ts
```diff
@@ -15,6 +15,11 @@ pluginTester({
       plugins: ['typescript'],
     },
   },
+  pluginOptions: {
+    babelOptions: {
+      presets: ['@babel/typescript'],
+    },
+  },
```
layershifter, line 22: The preset is added as a config option of the plugin.

ling1726: I guess normally apps written in TS would have the preset for TS anyway in their babel setup?

ling1726 (approved):

packages/babel-preset/src/transformPlugin.test.ts
layershifter: Yes, that's correct. That's why it was there, but as this dependency is implicit it's not the best idea to have it in this way.

This will be documented later properly.

miroslavstastny (approved):

Issue #49, opened 2022-02-23
Cannot specify `zIndex` & `opacity` values due TS errors on nested rules


layershifter:
### Problem

```ts
import { makeStyles } from "@griffel/react";
import "./styles.css";

const useClasses = makeStyles({
  rootA: {
    ".foo .bar .baz": {
      opacity: 1
    }
  },
  rootB: {
    ".foo .bar .baz": {
      zIndex: 1
    }
  }
});
```

The following code should not produce TypeScript errors, but it does:

```
Type '{ opacity: number; }' is not assignable to type 'GriffelStylesCSSValue | GriffelStylesCSSObjectCustomL2 | undefined'.
  Type '{ opacity: number; }' is not assignable to type 'GriffelStylesCSSObjectCustomL2'.
    Type '{ opacity: number; }' is not assignable to type '{ [Property: string]: GriffelStylesCSSValue | GriffelStylesCSSObjectCustomL3 | undefined; }'.
      Property 'opacity' is incompatible with index signature.
        Type 'number' is not assignable to type 'GriffelStylesCSSValue | GriffelStylesCSSObjectCustomL3 | undefined'.ts(2322)
```

### Repro

https://codesandbox.io/s/gifted-newton-e6lmvo?file=/src/App.tsx

PR #57, opened 2022-03-02
test: add test for typings based with different TS versions

layershifter:
This PR adds tests for our typings are compatible with different versions of TypeScript including 3.9.
It's prerequisite for #49 as we want to ensure that fixes will not affect compatibility.

### Test steps

- Use existing artifacts produced by `build` step
- Pack `@griffel/core` package in a archive
- Create a temporary directory and stub project in it
- Install `@griffel/core` from an archive, install needed version of TypeScript 
- Run TS compiler 🚀

image

### Notes

- `types.test.ts` was moved to new package that contains test's code


e2e/typescript/src/test.mjs
```diff
@@ -0,0 +1,137 @@
...
+
+    tscBin = path.resolve(tempDir, 'node_modules', 'typescript', 'bin', 'tsc');
+    console.log(
+      logSymbols.info,
+      'Using TypeScript',
+      (await sh(`${tscBin} --version`, tempDir, true)).replace('Version', '').trim(),
+    );
+  } catch (e) {
+    console.error(logSymbols.error, 'Something went wrong setting up the test:');
+    console.error(/** @type {Error} */ e?.stack || e);
```
miroslavstastny, line 107:
```suggestion
    console.error(/** @type {Error} */ e?.stack ?? e);
```

miroslavstastny (approved):

PR #58, opened 2022-03-02
docs: add home page

layershifter:
This PR adds a draft homepage that we can use initially.

image


apps/website/src/pages/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+import * as React from 'react';
+
+import styles from './index.module.css';
+
+const Homepage: React.FC = () => {
+  return (
+    <Layout>
+      <div className={styles['container']}>
+        <div className={styles['containerLayer']} />
+        <div className={cx('hero hero--dark', styles['hero'])}>
```
layershifter, line 15: These components are from https://infima.dev/ included with Docusaurus.

ling1726 (approved):

PR #63, opened 2022-03-06
docs: Add templates to try-it-out page

ling1726:
Adds templates to quickly toggle between in the sidebar. Uses a storybook-like loading pattern, where creating a new module inside the templates folder will auto populate the sidebar and new template in the docsite.

Also supports loading metadata from within the template module. The metadata object is not stripped out, but could be done pretty easily in the future through babel standalone in the browser or at build time


apps/website/sidebars.js
```diff
@@ -1,8 +1,56 @@
...
+                babel.types.isProperty(property) &&
+                babel.types.isIdentifier(property.key) &&
+                (babel.types.isStringLiteral(property.value) || babel.types.isBooleanLiteral(property.value))
+              ) {
+                meta[property.key.name] = property.value.value;
+              }
+            });
+          }
+        },
+      });
```
layershifter, line 36: nit: It's preferable to use paths over nodes

```js
VariableDeclarator(path) {
          const idPath = path.get('id');
          const initPath = path.get('init');

          if (idPath.isIdentifier({ name: 'meta' }) && initPath.isObjectExpression()) {
            initPath.get('properties').forEach(propertyPath => {
              if (propertyPath.isObjectProperty()) {
                const keyPath = propertyPath.get('key');
                const valuePath = propertyPath.get('value');

                if (keyPath.isIdentifier() && (valuePath.isStringLiteral() || valuePath.isBooleanLiteral())) {
                  meta[keyPath.node.name] = valuePath.node.value;
                }
              }
            });
          }
        },
```

apps/website/src/components/Playground/code/templates/border.js
```diff
@@ -0,0 +1,21 @@
+//@ts-check
```
layershifter, line 1: Let's use `checkJs: true` in `apps/website/tsconfig.json` instead

apps/website/src/components/Playground/index.tsx
```diff
@@ -1,14 +1,30 @@
...
+import path from 'path-browserify';
+
+const ctx = require.context('!!raw-loader!./code/templates', false, /\.js$/);
+
+const templates: Record<string, string> = ctx.keys().reduce((acc, modulePath) => {
+  if (modulePath.includes('app')) {
+    return acc;
+  }
+
+  const templateName = path.parse(modulePath).name;
```
layershifter, line 16: AFAIR Webpack normalizes paths, so we can replace this module with `templateName.split('/').slice(-1)[0]`

layershifter (commented):
Before merging this PR we need to fix type-checking in `/Users/olfedias/WebstormProjects/griffel/apps/website/tsconfig.json`

`types` should in `compilerOptions`, https://www.typescriptlang.org/tsconfig#types

```diff
    "noFallthroughCasesInSwitch": true,
+   "types": ["environment", "static-assets", "@docusaurus/module-type-aliases", "@docusaurus/theme-classic"]
  },
- "types": ["environment", "static-assets", "@docusaurus/module-type-aliases", "@docusaurus/theme-classic"],
```

For `require.context()` you will also need to install `@types/webpack-env` and add it to `compilerOptions.types`

apps/website/src/components/Playground/index.tsx
ling1726: had to strip the extension as well, but yeah that works

apps/website/sidebars.js
```diff
@@ -1,8 +1,56 @@
 // @ts-check
 
+const path = require('path');
+const fs = require('fs');
+const babel = require('@babel/core');
+
+function generateTryItOutSidebar() {
+  const playgroundTemplatePath = path.join(__dirname, '/src/components/Playground/code/templates');
+  const templateFiles = fs.readdirSync(playgroundTemplatePath).filter(filename => filename !== 'app.js');
```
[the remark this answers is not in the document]
ling1726: it was actually already the case, I just removed the filtering

apps/website/src/components/Playground/index.tsx
```diff
@@ -13,7 +12,8 @@ const templates: Record<string, string> = ctx.keys().reduce((acc, modulePath) =>
     return acc;
   }
 
-  const templateName = path.parse(modulePath).name;
+  const templateName = modulePath.split('/').slice(-1)[0].split('.')[0];
+  console.log(templateName);
```
layershifter, line 16:
```suggestion
```

Oops

layershifter (approved):

PR #65, opened 2022-03-08
chore: add explicit dependencies

layershifter:
This PR:
- adds dependencies for each package explicitly, this allows `beachball` (releasing tool) to properly populate & bump dependencies
- to ensure that all dependencies are declared I added `import/no-extraneous-dependencies`:
  image
- to ensure that dependencies are in sync I added `syncpack` check

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
   "scripts": {
     "build": "nx affected:build",
     "change": "beachball change",
+    "check-dependencies": "syncpack list -pP",
     "lint": "nx affected:lint",
     "start": "nx run @griffel/website:serve",
     "storybook": "nx run @griffel/react:storybook",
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -13,6 +13,7 @@
   "scripts": {
     "build": "nx affected:build",
     "change": "beachball change",
+    "check-dependencies": "syncpack list -pP",
```
miroslavstastny, line 16: There is a list of script an engineer is supposed to run before pushing - lint, type-check, test... now this one.
What is the best way to know about that list? Not a big fan of precommit hook but can we have something like `ci` of `verify` which I can run (or to know the complete list)?

miroslavstastny (approved):

PR #70, opened 2022-03-09
fix: introduce modern typings for TS >4

### LINKED ISSUE #378: bug(core): Typescript error using animationName inside of a nested selector
repro: https://codesandbox.io/s/elegant-moon-odwjen?file=/src/App.tsx

When using `animationName` inside a nested selector, there are unexpected typescript errors:

```
(property) "&[data-foo]": {
    animationName: {
        from: {
            opacity: number;
        };
        to: {
            opacity: number;
        };
    }[];
}
Type '{ animationName: { from: { opacity: number; }; to: { opacity: number; }; }[]; }' is not assignable to type 'string | number | (string | number)[] | GriffelStylesCSSObjectCustomL2'.
  Type '{ animationName: { from: { opacity: number; }; to: { opacity: number; }; }[]; }' is not assignable to type 'string'.ts(2322)
```

layershifter:
Fixes #378.

This PR introduces modern typings for TypeScript 4.1 and higher as this version supports recursive types references. This allow to simplify typings and remove weird workarounds. Also newer typings make TS errors more precise (check changes in the fixtures) 👍 

To keep compatibility with older versions of TS older typings are kept as legacy. To resolve them `typesVersions` field is used.

### DIFF
--- a/e2e/typescript/src/assets/fixture.ts
+++ b/e2e/typescript/src/assets/fixture.ts
@@ -62,6 +62,13 @@ assertType({
   ':hover': { '--color': 'red' },
 });
 
+assertType({
+  ':hover': { content: "'foo'" },
+});
+assertType({
+  ':hover': { animationName: { from: {}, to: {} } },
+});
+
 // Custom selectors
 //
 
@@ -86,6 +93,9 @@ assertType({
 
 assertType({
   '.bar': { '--color': 'red' },
+});
+assertType({
+  '.bar': { animationName: { from: {}, to: {} } },
 });
 
 // Nested custom selectors
@@ -192,7 +202,6 @@ assertType({
   '@media screen and (max-width: 992px)': {
     ...typedMixin,
   },
-  // @ts-expect-error FIXME - this is supposed to pass!
   '@media(forced-colors: active)': {
     ...typedMixin,
     color: 'var(--customColor)',
@@ -201,6 +210,13 @@ assertType({
     ...typedMixin,
   },
   ':before': {
+    ...typedMixin,
+    color: 'var(--customColor)',
+  },
+  '& .foo': {
+    ...typedMixin,
+  },
+  '& .bar': {
     ...typedMixin,
     color: 'var(--customColor)',
   },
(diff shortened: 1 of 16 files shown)


e2e/typescript/src/assets/fixture.ts
```diff
@@ -103,20 +109,30 @@ assertType({
...
+  overflow: '1',
+});
+assertType({
+  // @ts-expect-error "paddingLeft" cannot be numeric value
+  paddingLeft: 5,
+});
+assertType({
+  // @ts-expect-error "0" is invalid value for "color"
+  color: 0,
+});
```
layershifter, line 135: Just formatting change to detect errors on an exact line.

packages/core/project.json
```diff
@@ -31,6 +31,14 @@
...
       }
     },
+    "build": {
+      "executor": "@nrwl/workspace:run-commands",
+      "dependsOn": [{ "target": "prebuild", "projects": "self" }],
+      "options": {
+        "commands": [{ "command": "node tools/createLegacyTypesForCore.mjs" }],
+        "outputPath": ["dist/packages/core/types-legacy"]
+      }
+    },
```
layershifter, line 41: Runs the new script that copies all `.d.ts` files to `types-legacy` and then updates imports.

e2e/utils/src/createTempDir.ts
```diff
@@ -3,7 +3,7 @@ import * as logSymbols from 'log-symbols';
 
 export function createTempDir(prefix: string) {
   // "Unsafe" means delete even if it still has files inside (our desired behavior)
-  const tempDir = tmp.dirSync({ prefix: 'nextjs', unsafeCleanup: true }).name;
+  const tempDir = tmp.dirSync({ prefix, unsafeCleanup: true }).name;
```
layershifter, line 6: Oops

packages/style-types/src/legacy/index.ts
```diff
@@ -0,0 +1,6 @@
+export type { GriffelAnimation, GriffelStyle } from './makeStyles';
```
layershifter, line 1: Copied from previous typings.

ling1726 (approved):

PR #71, opened 2022-03-09
chore: add compliance checks

layershifter:
This PR adds new compliance checks required by security policies. They will run before the actual release.

### DIFF
--- a/azure-pipelines.compliance.yml
+++ b/azure-pipelines.compliance.yml
@@ -1,0 +1,63 @@
+# These steps have to run on a windows machine,
+# and therefore unfortunately can't be integrated in the regular steps
+
+parameters:
+  - name: serviceTreeID
+    type: string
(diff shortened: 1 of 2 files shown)


miroslavstastny (approved):

azure-pipelines.compliance.yml
```diff
@@ -0,0 +1,63 @@
...
+parameters:
+  - name: serviceTreeID
+    type: string
+    default: 'PLEASE USE YOUR SERVICE TREE ID FOR THE REPO'
+
+jobs:
+  - job: Compliance
+    displayName: Compliance checks
+    pool:
+      name: 1ES-Teams-Win2019-Security
```
layershifter, line 13:
```suggestion
    pool: uifabric-windows-2019-small
```

layershifter: We should use a different pool

PR #72, opened 2022-03-10
chore: improve React 18 support

### LINKED ISSUE #40: react: Better support for React 18
RC for React 18 is already published:

image

---

> **NOTE: Make sure you read the section "When to Insert <style> on The Client". If you currently inject style rules "during render", it could make your library VERY slow in concurrent rendering.**
> https://github.com/reactwg/react-18/discussions/110

In short we should use `useInsertionEffect` to apply our styles to DOM.

For reference:
- changes in Emotion https://github.com/emotion-js/emotion/pull/2600
- issue with bundling https://github.com/emotion-js/emotion/pull/2651

- [x] Types in React 18 does not allow `children` by default in `React.FC`, check components and update (#214)

layershifter:
Fixes #40.

This PR improves compatibility of `@griffel/react` by using a custom insertion factory that uses [`useInsertionEffect`](https://react.dev/reference/react/useInsertionEffect) when it's available.

### DIFF
--- a/packages/react/src/insertionFactory.ts
+++ b/packages/react/src/insertionFactory.ts
@@ -7,16 +7,14 @@ export const insertionFactory: GriffelInsertionFactory = () => {
   const insertionCache: Record<string, boolean> = {};
 
   return function insert(renderer: GriffelRenderer, cssRules: CSSRulesByBucket) {
-    if (useInsertionEffect) {
-      // Even if `useInsertionEffect` is available, we can't use it in SSR as it will not be executed
-      if (canUseDOM()) {
-        // eslint-disable-next-line react-hooks/rules-of-hooks
-        useInsertionEffect(() => {
-          renderer.insertCSSRules(cssRules!);
-        }, [renderer, cssRules]);
+    // Even if `useInsertionEffect` is available, we can use it on a client only as it will not be executed in SSR
+    if (useInsertionEffect && canUseDOM()) {
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      useInsertionEffect(() => {
+        renderer.insertCSSRules(cssRules!);
+      }, [renderer, cssRules]);
 
-        return;
-      }
+      return;
     }
 
     if (insertionCache[renderer.id] === undefined) {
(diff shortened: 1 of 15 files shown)


packages/react/src/useInsertionEffect.ts
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react';
+
+// @ts-expect-error
+export const useInsertionEffect = React['useInsertion' + 'Effect']
```
spmonahan, line 4: Looks like this string concat is done to prevent issues with transpiling/bundling. Would be nice to have a comment explaining this, it wasn't obvious to me until I read [this emotion PR](https://github.com/emotion-js/emotion/pull/2651).

packages/core/src/makeStyles.ts
```diff
@@ -43,10 +46,8 @@ export function makeStyles<Slots extends string | number>(stylesBySlots: StylesB
       }
     }
 
-    if (insertionCache[rendererId] === undefined) {
-      renderer.insertCSSRules(cssRules!);
-      insertionCache[rendererId] = true;
-    }
+    insertStyles(renderer, dir, cssRules!);
```
ling1726, line 49: nit: maybe at this point the `computeClasses` function name above was never quite clear. This step does not only compute the classes but also inserts them to DOM

packages/react/src/insertionFactory.ts
```diff
@@ -0,0 +1,27 @@
...
+        return;
+      }
+    }
+
+    if (insertionCache[renderer.id] === undefined) {
+      renderer.insertCSSRules(cssRules!);
+      insertionCache[renderer.id] = true;
+    }
+  };
+};
```
ling1726, line 27: nit: you only need one level

```suggestion
  return function insert(renderer: GriffelRenderer, cssRules: CSSRulesByBucket) {
    if (useInsertionEffect && canUseDOM) {
      // Even if `useInsertionEffect` is available, we can't use it in SSR as it will not be executed
      // eslint-disable-next-line react-hooks/rules-of-hooks
      useInsertionEffect(() => {
        renderer.insertCSSRules(cssRules!);
      }, [renderer, cssRules]);

      return;
    }

    if (insertionCache[renderer.id] === undefined) {
      renderer.insertCSSRules(cssRules!);
      insertionCache[renderer.id] = true;
    }
  };
};
```

ling1726 (approved):

PR #74, opened 2022-03-11
fix: Improve typings for styles object

miroslavstastny:
Disallow invalid values on CSS properties. This previously passed type check even though `outline-box` is not a valid value for `box-sizing`:

```js
assertType({
  boxSizing: 'outline-box', // ✅  passed before, now correctly fails
});
```

Another example is an object value assigned to a CSS property:
```js
assertType({
  zIndex: { color: 'red' }, // ✅  passed before, now correctly fails
});
```

Typecheck fails on correct line:
```js
assertType({
  boxSizing: 'outline-box',
  zIndex: 1, // ✅  this previously failed for zIndex, but the error is the outline-box value
});
```

### Limitations

The typings allow any value (string, number, CSS object) to be assigned to an unknown property or a CSS variable:
```js
assertType({
  unknownProperty: 'allowed', // 🔴 passes
  '--variableWithObject': { zIndex: 1}, // 🔴 passes
  '--variableWithNumber': 42 // 🔴 passes
});
```

### DIFF
--- a/change/@griffel-react-41ab2ed0-7da1-4045-86ef-915d5b5a7cd0.json
+++ b/change/@griffel-react-41ab2ed0-7da1-4045-86ef-915d5b5a7cd0.json
@@ -1,7 +1,0 @@
-{
-  "type": "patch",
-  "comment": "fix: Improve typings for styles object",
-  "packageName": "@griffel/react",
-  "email": "miroslav.stastny@microsoft.com",
-  "dependentChangeType": "patch"
-}
--- a/change/@griffel-react-41ab2ed0-7da1-4045-86ef-915d5b5a7cd0.json
+++ b/change/@griffel-react-41ab2ed0-7da1-4045-86ef-915d5b5a7cd0.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix: Improve typings for styles object",
+  "packageName": "@griffel/react",
+  "email": "miroslav.stastny@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/e2e/typescript/src/assets/fixture.ts
+++ b/e2e/typescript/src/assets/fixture.ts
@@ -4,6 +4,9 @@ function assertType(style: GriffelStyle): GriffelStyle {
   return style;
 }
 
+// Animation
+assertType({ animationName: 'foo' });
+
 // Basic styles
 //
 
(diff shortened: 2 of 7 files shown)


change/@griffel-react-41ab2ed0-7da1-4045-86ef-915d5b5a7cd0.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix: Improve typings for styles object",
+  "packageName": "@griffel/react",
+  "email": "miroslav.stastny@microsoft.com",
+  "dependentChangeType": "patch"
+}
```
miroslavstastny, line 7: @layershifter - is it correct that the change entry was generated for anything else than `@griffel/core`?

packages/core/src/types.ts
```diff
@@ -55,7 +55,7 @@ type GriffelStylesCSSObjectCustomL5 = {
 } & Partial<GriffelStylesUnsupportedCSSProperties> &
   GriffelStylesStrictCSSObject;
 
-export type GriffelStyle = GriffelStylesCSSObjectCustomL1;
+export type GriffelStyle = GriffelStylesStrictCSSObject | GriffelStylesCSSObjectCustomL1;
```
miroslavstastny, line 58: I originally wanted to remove the strict object here. Unfortunately, without that object spreading (mixins) does not pass the type check in TS<4.2:

```ts
type T1 = {
  [Property: string]: string | number
}

type T2 = {
  n: number
}

type Compound = T1 & Partial<T2>

function assertType(t: Compound): Compound {
  return t;
}

const mixin: Compound = {
  color: 'red'
}

assertType({...mixin, background: 'blue'}) // 🔴 Type '{ background: string; n?: number | undefined; }' is not assignable to type 'T1' in TS<4.2
```
[Typescript playground](https://www.typescriptlang.org/vo/play?ts=3.9.7#code/C4TwDgpgBAKgjFAvFA3gWAFBSgbQAoBOA9pAaAFxQDOwBAlgHYDmAupTfc1AD5QMCuAWwBGEApgC+mTKEiwATElSZsDSgJFjJ0jLOgBhIoLBF+DACZL4UAGRQ8AQzJ0HAGwA8MeQD4dAMzMAY2A6IgYoByoqMWAYcAgACmBKQ2NTCwBKFKMTM0t0LCgCCGB+AnDgAG5tDExAsJooQToAD0ZstLylAux61yICSgByYvMhmsxI6LI4yASUADol5raGABooYQdAgGsmYjzh4Vd+CCGJDKgAeiuoQF4NwBZd2HioIZRN7b2Di3ZaRiZKnwAPzqISiAg8KB5CB+RgQcyAiRDKB0Kh8IjACJROhMBgOY7QYBEKB6V7wZGMWAAZXcABYFvIgA)

packages/core/src/types.ts
```diff
@@ -30,35 +30,30 @@ type GriffelStylesCSSPseudos = {
 // updated or removed.
 //
 
-type GriffelStylesCSSObjectCustomL1 =
-  | ({
-      [Property: string]: string | undefined | GriffelStylesCSSObjectCustomL2;
-    } & Partial<GriffelStylesUnsupportedCSSProperties>)
```
miroslavstastny, line 36: Removed `Partial<GriffelStylesUnsupportedCSSProperties>` as that is already a part of `GriffelStylesStrictCSSObject`

change/@griffel-react-41ab2ed0-7da1-4045-86ef-915d5b5a7cd0.json
layershifter: No, can you please delete these changefiles?

e2e/typescript/src/assets/fixture.ts
```diff
@@ -103,20 +117,61 @@ assertType({
...
+  // @ts-expect-error Object is not assignable to CSS property
+  opacity: { color: 'red' },
+});
+
+// Mixins with invalid values
+//
+
+// @ts-expect-error Object in "zIndex" is not assignable to CSS property
+assertType({
+  ...{ zIndex: { color: 'red' } },
```
miroslavstastny, line 167: We should also add a mixin which is typed as `GriffelStyle` (similar to what we have in babel fixtures)

miroslavstastny: added

layershifter (approved):

ling1726 (approved):

PR #79, opened 2022-03-22
docs: show real inserted styles should be used in playground

ling1726:
Modifies the playground so that it inserts all the css rules into a style tag, and then outputs the real styles.

Obviously validation would be great, but a bit more complicated since [`stylis`](https://github.com/thysultan/stylis) already strips most invalid css styles. I'll try to tackle this in another PR. My idea would be to use this maybe [w3c-css-validator](https://www.npmjs.com/package/w3c-css-validator)

Currently, invalid CSS is just inserted as an empty rule, it's some validation but should be improved.

### DIFF
--- a/apps/website/src/components/Playground/code/templates/selectors.js
+++ b/apps/website/src/components/Playground/code/templates/selectors.js
@@ -9,19 +9,19 @@ export default makeStyles({
 
   attribute: {
     '&[data-attribute]': {
-      color: 'red',
+      color: 'magenta',
     },
   },
 
   class: {
-    '&. child-class': {
-      color: 'red',
+    '&.my-class': {
+      color: 'olive',
     },
   },
 
   childClass: {
-    '&.my-class': {
-      color: 'red',
+    '& .my-class': {
+      color: 'pink',
     },
   },
 });
--- a/apps/website/src/components/Playground/code/app.js
+++ b/apps/website/src/components/Playground/code/app.js
@@ -11,18 +11,32 @@ export default function App() {
   const ref = React.useRef(null);
 
   React.useEffect(() => {
+    const styleTag = document.createElement('style');
+    document.head.append(styleTag);
+
     /** @type import('@griffel/core').GriffelRenderer */
     const playgroundRenderer = {
       id: 'playground',
       insertCSSRules(cssRules) {
-        const raw = Object.values(cssRules)
-          .reduce(
-            (acc, bucketRules) =>
-              acc.concat(bucketRules.map(cssRule => (Array.isArray(cssRule) ? cssRule[0] : cssRule))),
-            [],
-          )
-          .join('\n');
+        Object.values(cssRules)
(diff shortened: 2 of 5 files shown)


apps/website/src/components/Playground/template/App.js
```diff
@@ -12,12 +12,30 @@ export default function App() {
   const [rules, setRules] = React.useState('');
   const ref = React.useRef(null);
   React.useEffect(() => {
+    const styleTag = document.createElement('style');
+    styleTag.id = 'playground';
```
layershifter, line 16: Is this `id` used somehow?

apps/website/src/components/Playground/template/App.js
```diff
@@ -12,12 +12,30 @@ export default function App() {
...
+      insertCSSRules(cssRuleBuckets) {
+        const cssRules = Object.values(cssRuleBuckets)
+          .reduce((acc, val) => {
+            return acc.concat(val);
+          }, [])
+          .filter(Boolean);
+
+        cssRules.forEach(cssRule => {
+          styleTag.sheet?.insertRule(cssRule);
+        });
```
layershifter, line 31:
```suggestion
        Object.values(cssRuleBuckets)
          .flatMap(rules => rules)
          .forEach(cssRule => {
            styleTag.sheet?.insertRule(cssRule);
          });
```

Is it better?

apps/website/src/components/Playground/template/App.js
```diff
@@ -12,12 +12,30 @@ export default function App() {
...
+
+        if (!styleTag.sheet) {
+          return;
+        }
+
+        const raw = Array.from(styleTag.sheet.cssRules)
```
layershifter, line 37:
```suggestion
        const raw = Array.from(styleTag.sheet.cssRules).reverse()
```

Can we add `.reverse()` so it will follow the definition order?

image

ling1726: yeah I didn't notice that, good idea

apps/website/src/components/Playground/template/App.js
ling1726: not really, it's just for debugging purposes. I can hide it behind a  NODE_ENV check ?

layershifter: If it's not needed I would rather delete it...

apps/website/src/components/Playground/code/templates/selectors.js
```diff
@@ -9,19 +9,19 @@ export default makeStyles({
...
     '&[data-attribute]': {
-      color: 'red',
+      color: 'magenta',
     },
   },
 
   class: {
-    '&. child-class': {
-      color: 'red',
+    '&.my-class': {
```
layershifter, line 17: Aha, an invalid selector 👀

layershifter (approved):

PR #80, opened 2022-03-22
feat: add CSS extraction

layershifter:
This PR is the last one to support CSS extraction.

The PR:
- add sorting of CSS rules and implements the same logic as #141 with Stylis
- adds new fixtures and tests
- cleanup `@internal` modifiers and adds necessary exports from `@griffel/core`

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -138,7 +138,7 @@ export type CSSClasses = /* ltrClassName */ string | [/* ltrClassName */ string,
 export type CSSClassesMap = Record<PropertyHash, CSSClasses>;
 export type CSSClassesMapBySlot<Slots extends string | number> = Record<Slots, CSSClassesMap>;
 
-export interface CSSRulesByBucket {
+export type CSSRulesByBucket = {
   // default
   d?: CSSBucketEntry[];
   // link
@@ -161,7 +161,7 @@ export interface CSSRulesByBucket {
   t?: CSSBucketEntry[];
   // @media rules
   m?: CSSBucketEntry[];
-}
+};
 
 export type CSSBucketEntry = string | [string, Record<string, unknown>];
 
--- a/packages/webpack-extraction-plugin/__fixtures__/webpack/style-buckets/code.ts
+++ b/packages/webpack-extraction-plugin/__fixtures__/webpack/style-buckets/code.ts
@@ -13,7 +13,14 @@ const styles = __styles(
     h: ['.color-yellow:hover { color: yellow; }'],
     a: ['.color-black:active { color: black; }'],
     k: ['@keyframes foo { from{ transform:rotate(0deg); } to { transform:rotate(360deg); } }'],
-    t: ['@media (forced-colors: active) { .color-magenta { color:magenta; } }'],
+    m: [
+      [
+        '@media (forced-colors: active) { .color-magenta { color: magenta; } }',
+        {
+          m: '(forced-colors: active)',
+        },
+      ],
+    ],
   },
 );
 
--- a/packages/webpack-extraction-plugin/__fixtures__/webpack/style-buckets/code.ts
+++ b/packages/webpack-extraction-plugin/__fixtures__/webpack/style-buckets/code.ts
@@ -1,0 +1,20 @@
+import { __styles } from '@griffel/react';
+
+const styles = __styles(
+  {},
+  // Classes in this test are intentionally not realistic to simplify snapshots
+  {
+    d: ['.color-red { color: red; }', '.animation-name { animation-name: foo; }'],
+    l: ['.color-orange:link { color: orange; }'],
+    v: ['.color-purple:visited { color: purple; }'],
+    w: ['.color-pink:focus-within { color: pink; }'],
+    f: ['.color-blue:focus { color: blue; }'],
+    i: ['.color-light-blue:focus-visible { color: salmon; }'],
+    h: ['.color-yellow:hover { color: yellow; }'],
+    a: ['.color-black:active { color: black; }'],
+    k: ['@keyframes foo { from{ transform:rotate(0deg); } to { transform:rotate(360deg); } }'],
+    t: ['@media (forced-colors: active) { .color-magenta { color:magenta; } }'],
+  },
+);
+
+console.log(styles);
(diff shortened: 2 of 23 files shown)


packages/webpack-extraction-plugin/tsconfig.spec.json
```diff
@@ -0,0 +1,21 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "jsx": "react",
+    "outDir": "../../dist/out-tsc",
+    "module": "commonjs",
+    "types": ["jest", "node", "environment"]
+  },
+  "include": [
+    "__fixtures__/*/code.ts",
```
layershifter, line 10: This should be tweak to track all TS files except `output.ts`

packages/webpack-extraction-plugin/virtual-loader/index.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Moves CSSloader to the end of the loader queue so it runs first.
+ */
+virtualLoader.pitch = function () {
+  if (this.loaders[0].pitch !== virtualLoader.pitch) {
+    // If the first loader isn't this one - skip.
+    return;
+  }
+
+  // The first loader is Compiled's css-loader - we need to shift
```
layershifter, line 23: - Update this copy-paste  
- Ensure that it's needed

packages/webpack-extraction-plugin/__fixtures__/webpack/style-buckets/code.ts
```diff
@@ -0,0 +1,20 @@
...
+    d: ['.color-red { color: red; }', '.animation-name { animation-name: foo; }'],
+    l: ['.color-orange:link { color: orange; }'],
+    v: ['.color-purple:visited { color: purple; }'],
+    w: ['.color-pink:focus-within { color: pink; }'],
+    f: ['.color-blue:focus { color: blue; }'],
+    i: ['.color-light-blue:focus-visible { color: salmon; }'],
+    h: ['.color-yellow:hover { color: yellow; }'],
+    a: ['.color-black:active { color: black; }'],
+    k: ['@keyframes foo { from{ transform:rotate(0deg); } to { transform:rotate(360deg); } }'],
+    t: ['@media (forced-colors: active) { .color-magenta { color:magenta; } }'],
```
ling1726, line 16:
```suggestion
    m: ['@media (forced-colors: active) { .color-magenta { color:magenta; } }'],
```

shouldn't this be bucket `m` ?

packages/webpack-extraction-plugin/src/sortCSSRules.ts
```diff
@@ -0,0 +1,82 @@
...
+  );
+}
+
+export function sortCSSRules(css: string, compareMediaQueries: GriffelRenderer['compareMediaQueries']): string {
+  return serialize(
+    compile(css)
+      .map(element => ({
+        ...element,
+        bucketName: getStyleBucketNameFromElement(element),
+        metadata: getElementMetadata(element),
```
ling1726, line 63: I think this metadata should follow the same object pattern that is used in the DOM renderer.

```ts
{ m: '(max-height: 1px)' }
```

That way adding metadata for core rules in the future will be straightfoward to translate to the extraction logic.

Alternative some kind of metadata parser pattern would be good because you can share that with the extraction implementation and core. However with currently only one rule with one metadata property it's not currently necessary

packages/webpack-extraction-plugin/src/sortCSSRules.ts
```diff
@@ -0,0 +1,82 @@
...
+        reference: getElementReference(element),
+      }))
+      .filter(
+        (elementA, index, self) => self.findIndex(elementB => elementA.reference === elementB.reference) === index,
+      )
+      .sort((nodeA, nodeB) => {
```
ling1726, line 69: I wonder if we can create a util for this in core. since `renderToStyleElement` has a similar ordering requirement. I like that it's only done in one sort pass here

packages/webpack-extraction-plugin/__fixtures__/webpack/style-buckets/code.ts
layershifter: Yeah, it should. I also discovered that meta was not handled properly, see 123c2e3

packages/core/src/types.ts
```diff
@@ -138,7 +138,7 @@ export type CSSClasses = /* ltrClassName */ string | [/* ltrClassName */ string,
 export type CSSClassesMap = Record<PropertyHash, CSSClasses>;
 export type CSSClassesMapBySlot<Slots extends string | number> = Record<Slots, CSSClassesMap>;
 
-export interface CSSRulesByBucket {
+export type CSSRulesByBucket = {
```
layershifter, line 141: `interface` produces any on `Object.values()` 🥹

[TS playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgILIN4ChnJge3wC5kBnMKUAcwG0BdHZAIzihPMpFoYF8ssE+EOWRwS6ALyZGBYshoByWQEYFAGmRLCAJgV01jFm3kKjq3v0HCwzccinZcAeifIAAmFIBaCAA8ADhAIYD5QUPhQyADq4VzI-uGBUGAAnoZwAF4kikZ6WHwCQiIAbvbIAPJMAFZBYAB0xXAANgCuEKQAFHAAlFguyADKABb4LU0AJswocCCiYXApAPzIAEItNsBgCqSiIGlWJcplxTQADAzFynXApAP4ALYQYEPUHd0A3EA)

packages/webpack-extraction-plugin/src/sortCSSRules.ts
layershifter: Well, we don't have it typed, so for now I will leave it as it. As you said there is a single usage currently...

packages/webpack-extraction-plugin/src/sortCSSRules.ts
layershifter: I will do this in a follow up PR.

packages/webpack-extraction-plugin/src/sortCSSRules.ts
```diff
@@ -0,0 +1,82 @@
...
+  return serialize(
+    compile(css)
+      .map(element => ({
+        ...element,
+        bucketName: getStyleBucketNameFromElement(element),
+        metadata: getElementMetadata(element),
+        reference: getElementReference(element),
+      }))
+      .filter(
+        (elementA, index, self) => self.findIndex(elementB => elementA.reference === elementB.reference) === index,
```
ling1726, line 67: with the amount of rules that there could be, I really thing we should use something with a contant time lookup to avoid the duration of this step exploding

layershifter: Done in 2b72bf8

ling1726 (approved):

PR #82, opened 2022-03-22
feat: export shaker evaluator

layershifter:
This PR re-exports the evaluator from `@linaria/shaker` and updates README file for `@griffel/babel-preset`.

We had the same experience with both existing internal customers of Webpack loader - it's required to tweak evaluation settings. For that we need to configure the evaluator, but it's not re-exported we need to add a dependency to third party module and update it separately 😯

### DIFF
--- a/packages/babel-preset/README.md
+++ b/packages/babel-preset/README.md
@@ -101,7 +101,7 @@ module.exports = {
       {
         evaluationRules: [
           {
-            action: require('@linaria/shaker').default,
+            action: require('@griffel/babel-preset').shakerEvaluator,
           },
           {
             test: /[/\\]node_modules[/\\]/,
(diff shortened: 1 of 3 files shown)


packages/babel-preset/README.md
```diff
@@ -145,37 +145,25 @@ const useStyles = __styles(/* resolved styles */);
...
 $ DEBUG=linaria\* LINARIA_LOG=debug yarn build
 ```
 
-Using Lage from the root of the repo
-
-```sh
-$ DEBUG=linaria\* LINARIA_LOG=debug yarn lage build --to <package-name>
-```
-
-Using `yarn workspace` from the root of the repo
```
layershifter, line 165: We don't have `lage` in this repo, third parties may not have it. It's not relevant

packages/babel-preset/README.md
```diff
@@ -101,7 +101,7 @@ module.exports = {
       {
         evaluationRules: [
           {
-            action: require('@linaria/shaker').default,
+            action: require('@griffel/babel-preset').shakerEvaluator,
```
ling1726, line 104: should we point to linaria documentation how to configure this ?

ling1726 (approved):

packages/babel-preset/README.md
layershifter: Well, there is no way to configure `@linaria/shaker` so I am not sure what to link there...

PR #84, opened 2022-03-22
feat: add support for @layer

micahgodbolt:
This is part of my FHL project, so this is an exercise if nothing else. This PR adds support for @layer queries in griffel. How @layer works within an atomic CSS project is certainly a longer conversation. But here's the work to support it in a basic state.


packages/core/src/runtime/utils/hashClassName.ts
```diff
@@ -6,10 +6,11 @@ export interface HashedClassNameParts {
   value: string;
   pseudo: string;
   media: string;
+  layer: string;
   support: string;
 }
 
-export function hashClassName({ media, property, pseudo, support, value }: HashedClassNameParts): string {
+export function hashClassName({ media, layer, property, pseudo, support, value }: HashedClassNameParts): string {
```
miroslavstastny, line 13: Why is the `layer` added as an argument and then not used in the hash?

packages/core/src/runtime/compileCSS.ts
```diff
@@ -8,6 +8,7 @@ export interface CompileCSSOptions {
 
   pseudo: string;
   media: string;
+  layer: string;
```
miroslavstastny, line 11: @layershifter Don't we also need to update https://github.com/microsoft/griffel/blob/bc30fac684d0a8f8a9d92f84847bf111b87b4f71/packages/core/src/renderer/rehydrateRendererCache.ts#L6
and corresponding UT (`rehydrateCache() avoids double insertion`)?

change/@griffel-react-6d0f6f43-4864-4ba4-b292-8e46b9d1d20f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
miroslavstastny, line 2:
```suggestion
  "type": "none",
```
beachball will rollup the changelog from core

packages/core/src/runtime/utils/hashClassName.ts
micahgodbolt: No reason. I started adding 'layer' everywhere I saw media.

micahgodbolt: Seems i missed one :)

micahgodbolt: added it into hashing

packages/core/src/runtime/compileCSS.ts
micahgodbolt: added layer into line 6

layershifter:
@micahgodbolt there are some actions need for this PR 💡

- #52 introduced refactor to `resolveStyles`. While your PR does not have merge conflicts, fixes are needed.
  Just update `compileCSS()` calls in `packages/core/src/runtime/resolveStyleRules.ts` to include `layer`. 
- `layer` was included to hashing (in `bc764ce`) you need to update snapshots to use updated classes

👆 this should fix CI failures.

packages/core/src/runtime/resolveStyleRules.test.ts
```diff
@@ -321,6 +321,72 @@ describe('resolveStyleRules', () => {
...
+          color: green;
+        }
+        @layer color {
+          .fe3e8s9 {
+            color: red;
+          }
+        }
+      `);
+    });
+
```
layershifter, line 341:
```suggestion

    it('handles layer queries with commas & dots', () => {
      expect(
        resolveStyleRules({
          '@layer theme, layout, utilities': { color: 'red' },
          '@layer framework.utilities': { color: 'red' },
        }),
      ).toMatchInlineSnapshot(`
        @layer theme,layout,utilities {
          .f16sacd2 {
            color: red;
          }
        }
        @layer framework.utilities {
          .faxdetk {
            color: red;
          }
        }
      `);
    });
```

Let's also add a test for commas & dots case.

layershifter:
@micahgodbolt I am entirely sure what we should do there, but let's put `@layer` also to separate style bucket.

#### `packages/core/src/runtime/getStyleBucketName.ts`

```diff
-export function getStyleBucketName(pseudo: string, media: string, support: string): StyleBucketName {
+export function getStyleBucketName(pseudo: string, layer: string, media: string, support: string): StyleBucketName {
  // We are grouping all the at-rules like @media, @supports etc under `t` bucket.
- if (media || support) {
+ if (layer || media || support) {
```

#### `packages/core/src/runtime/getStyleBucketName.test.ts`

```diff
-    expect(getStyleBucketName(' :link', '', '')).toBe('l');
-    expect(getStyleBucketName(' :hover ', '', '')).toBe('h');
-    expect(getStyleBucketName(':link', '', '')).toBe('l');
-    expect(getStyleBucketName(':visited', '', '')).toBe('v');
-    expect(getStyleBucketName(':focus-within', '', '')).toBe('w');
-    expect(getStyleBucketName(':focus', '', '')).toBe('f');
-    expect(getStyleBucketName(':focus-visible', '', '')).toBe('i');
-    expect(getStyleBucketName(':hover', '', '')).toBe('h');
-    expect(getStyleBucketName(':active', '', '')).toBe('a');
-    expect(getStyleBucketName(':active', '(max-width: 100px)', '')).toBe('t');
-    expect(getStyleBucketName(':active', '', '(display: table-cell)')).toBe('t');

+    expect(getStyleBucketName(' :link', '', '', '')).toBe('l');
+    expect(getStyleBucketName(' :hover ', '', '', '')).toBe('h');

+    expect(getStyleBucketName(':link', '', '', '')).toBe('l');
+    expect(getStyleBucketName(':visited', '', '', '')).toBe('v');
+    expect(getStyleBucketName(':focus-within', '', '', '')).toBe('w');
+    expect(getStyleBucketName(':focus', '', '', '')).toBe('f');
+    expect(getStyleBucketName(':focus-visible', '', '', '')).toBe('i');
+    expect(getStyleBucketName(':hover', '', '', '')).toBe('h');
+    expect(getStyleBucketName(':active', '', '', '')).toBe('a');

+    expect(getStyleBucketName(':active', 'theme', '', '')).toBe('t');
+    expect(getStyleBucketName(':active', '', '(max-width: 100px)', '')).toBe('t');
+    expect(getStyleBucketName(':active', '', '', '(display: table-cell)')).toBe('t');
```

Calls of `getStyleBucketName` in `packages/core/src/runtime/resolveStyleRules.ts` should be also updated :)

layershifter:
Non actionable:
We need to do better testing for hydration, but `@layer` is not supported in JSDOM (similar to https://github.com/jsdom/jsdom/issues/2026) 😿

micahgodbolt:
Do you need me to dig into the new bucket work? I'm on shield this week, so it might be a bit before I can dig in.

layershifter:
> Do you need me to dig into the new bucket work? I'm on shield this week, so it might be a bit before I can dig in.

Please apply changes in https://github.com/microsoft/griffel/pull/84#issuecomment-1090133133 and we should be good to go 🐱

micahgodbolt:
> > Do you need me to dig into the new bucket work? I'm on shield this week, so it might be a bit before I can dig in.
> 
> Please apply changes in [#84 (comment)](https://github.com/microsoft/griffel/pull/84#issuecomment-1090133133) and we should be good to go 🐱

@layershifter are those suggestions the exact diffs you want to see made? Can you use the built in "add a suggestion" feature, and then I can just click to accept the change.

layershifter:
> @layershifter are those suggestions the exact diffs you want to see made?

Yes

> Can you use the built in "add a suggestion" feature, and then I can just click to accept the change.

Unfortunately, these files are not present in PR I don't know how to create suggestions 🤔

micahgodbolt:
@layershifter thanks for pointing out the need to change getStyleBucketname. I've updated the function and tests

layershifter:
> Calls of `getStyleBucketName` in `packages/core/src/runtime/resolveStyleRules.ts` should be also updated :)

@micahgodbolt I think that we are missing this step 👆 There are few similar failures:

```
../core/src/runtime/resolveStyleRules.ts:216:25 - error TS2345: Argument of type '{ value: string; property: string; pseudo: string; media: string; support: string; }' is not assignable to parameter of type 'HashedClassNameParts'.
  Property 'layer' is missing in type '{ value: string; property: string; pseudo: string; media: string; support: string; }' but required in type 'HashedClassNameParts'.

216         ? hashClassName({
                            ~
217             value: rtlDefinitions.map(v => (v?.value ?? '').toString()).join(';'),
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
221             support,
    ~~~~~~~~~~~~~~~~~~~~
222           })
    ~~~~~~~~~~~

  ../core/src/runtime/utils/hashClassName.ts:9:3
    9   layer: string;
        ~~~~~
    'layer' is declared here.

../core/src/runtime/resolveStyleRules.ts:233:31 - error TS2554: Expected 4 arguments, but got 3.

233       const styleBucketName = getStyleBucketName(pseudo, media, support);
```

micahgodbolt:
Passing! Thanks for all the feedback.

change/@griffel-core-9132796e-b0a7-46a9-b70f-72bef61024b9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "add support for @layer",
```
layershifter, line 3:
```suggestion
  "comment": "feat: add support for @layer",
```

change/@griffel-react-6d0f6f43-4864-4ba4-b292-8e46b9d1d20f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "docs: add support for @layer",
+  "packageName": "@griffel/react",
+  "email": "mgodbolt@microsoft.com",
+  "dependentChangeType": "patch"
```
layershifter, line 6:
```suggestion
  "dependentChangeType": "none"
```

layershifter (approved):
❤️

PR #88, opened 2022-03-30
Browser devtools: generate data

YuanboXue-Amber:
This PR aims to create data for Griffel brower devtool.

👉 TODO before closing PR - verify no changes are introduced in production bundle

## Basic concept of Griffel browser devtool 
Styles are applied to a DOM element through classNames. The classNames are returned by either `makeStyles` or `mergeClasses` call. The merging of classNames can be considered as a tree structure.

Example:
```tsx
const useButtonStyles = makeStyles({
  red: {
    color: 'red',
  },
  green: {
    color: 'green',
  },
  backgroundBlue: {
    backgroundColor: 'blue',
  },
});

const TwoButtons = ({ green, backgroundBlue }) => {
  const classes = useButtonStyles();
  const mergedClasses1 = mergeClasses(classes.red, green && classes.green);
  const mergedClasses2 = mergeClasses(mergedClasses1, backgroundBlue && classes.backgroundBlue);
  return (
    <>
      <button {...props} className={mergedClasses1} />
      <button {...props} id="blue-button" className={mergedClasses2} />
    </>
  );
};
```
Styles for button with id 'blue-button' are constructed like so:
Screenshot 2022-03-30 at 18 37 55

This PR will create the above tree structure, and devtool can rely on it for the UI.

#### Reason:
In Griffel, a sequence hash prefixed with `__` is used to represent a set of atomic styles. A sequence hash is also applied to DOM element's className. Devtool can rely on this sequence hash in DOM className to resolve the styles. 

Sequence hash is computed based on atomic styles only, without information of how styles are merged. Therefore it is not enough to construct the tree. Consider the above example: 

When `useButtonStyles` is called, hashes will be computed for all styles. 
`color: green` will have a sequence hash like __xxxxxxx. `color: red` will have its own hash as well.

When `mergeClasses1` is calculated, the 2 styles being merged are `color: red` and `color: green`; 
Griffel will compute sequence hash of the merged result: `color: green`, which is the same as the previously computed hash in `useButtonStyles` call. 

Given the sequence hash of `color: green`, there's no way to know if it is computed in `makeStyles` call or `mergeClasses` call.

#### When devtool is enabled,
A suffix is added to sequence hash, representing the order of classNames merging.

When `useButtonStyles` is called, `color: green` will have a sequence hash like __xxxxxxx_0000000.
When `color: red` and `color: green` are merged, the generated `color: green` will be represented by a new sequence hash: __xxxxxxx_yyyyyyy, where yyyyy is a hash of all merged sequences (in this case, sequence hash of `color: red`, and sequence hash of `color: green`).

Given an updated sequence hash __xxxxxxx_yyyyyyy, it is possible to create the tree structure representing styles merging.

### Step2, build a tree

Given a DOM element's className, we can compute a tree structure.
The leaf nodes are styles in `makeStyles` calls. The parent nodes are `mergeClasses`.
Leaf nodes contains also atomic css rules and its related className.

Generated data for the above example on button with id 'blue-button':
Screenshot 2022-04-09 at 17 24 20


packages/core/src/renderer/createDOMRenderer.ts
```diff
@@ -69,6 +72,8 @@ export function createDOMRenderer(
     },
   };
 
+  process.env.NODE_ENV !== 'production' && injectDevTools(window);
```
YuanboXue-Amber, line 75: @layershifter is this where we want to injectDevTools so it works with iframes?

YuanboXue-Amber:
CI is failing because of failed unit tests due to sequence hash update:
image

Do we want to run test against with/without debug sequence?

packages/core/src/constants.ts
```diff
@@ -10,6 +10,15 @@ export const SEQUENCE_HASH_LENGTH = 7;
 /** @internal */
 export const SEQUENCE_PREFIX = '___';
 
+/** @internal */
+export const DEBUG_SEQUENCE_SEPARATOR = '_';
+
+/** @internal */
+export const SEQUENCE_SIZE =
+  process.env.NODE_ENV === 'production'
```
miroslavstastny, line 18: To enable debug, we should check, in addition to NODE_ENV, a local storage flag.
See https://github.com/microsoft/fluentui/blob/f9ca45d012d2e77eea728e19651b144452562004/packages/fluentui/styles/src/debugEnabled.ts for inspiration

packages/core/src/devtools/types.ts
```diff
@@ -0,0 +1,21 @@
+import { SequenceHash } from '../types';
+
+declare global {
+  interface Window {
+    __MAKESTYLES_DEVTOOLS__: {
```
miroslavstastny, line 5: nit: 
```suggestion
    __GRIFFEL_DEVTOOLS__: {
```

packages/core/src/devtools/store.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const sequenceDetails: Record<SequenceHash, { slotName: string }> = {};
+
+const cssRules: string[] = [];
+
+export const MK_DEBUG = {
+  getChildrenSequences: (debugSequenceHash: SequenceHash): SequenceHash[] => {
+    for (const key in mergeClassesCachedResults) {
+      const sequenceClassName = mergeClassesCachedResults[key].split(' ')?.[0];
+      if (sequenceClassName === debugSequenceHash) {
```
miroslavstastny, line 13:
```suggestion
Object.keys(mergeClassesCachedResults)
  .find(key => mergeClassesCachedResults[key].startsWith(debugSequenceHash))
```
```

File: packages/core/src/renderer/createDOMRenderer.ts (hunk 1 above)
Comment (miroslavstastny) [inline packages/core/src/renderer/createDOMRenderer.ts:75]:
```suggestion
  process.env.NODE_ENV !== 'production' && document && injectDevTools(document.defaultView);
```

packages/core/src/constants.ts
layershifter: We should check how it works with SSR and you can't get `localStorage` from the server :)

packages/core/src/renderer/createDOMRenderer.ts
YuanboXue-Amber: applied :)

packages/core/src/devtools/store.ts
YuanboXue-Amber: applied :)

packages/core/src/devtools/types.ts
YuanboXue-Amber: applied :)

packages/core/src/constants.ts
YuanboXue-Amber: I applied the localStorage solution like the example from Miro. 
But IDK how to handle it with SSR. Will check or shout for help on Monday

YuanboXue-Amber: Updated it to use debug hash always for non prod :)

packages/core/src/runtime/utils/hashSequence.ts
```diff
@@ -17,6 +17,15 @@ function padEndHash(value: string): string {
...
+  classes: string,
+  dir: 'ltr' | 'rtl',
+  sequenceIds: (SequenceHash | undefined)[] = [],
+): SequenceHash {
+  return process.env.NODE_ENV !== 'production'
+    ? SEQUENCE_PREFIX +
+        padEndHash(hash(classes + dir)) +
+        DEBUG_SEQUENCE_SEPARATOR +
+        padEndHash(hash(sequenceIds.join('')))
+    : SEQUENCE_PREFIX + padEndHash(hash(classes + dir));
```
layershifter, line 30: nit, let's make this more readable

```suggestion
  if (process.env.NODE_ENV === 'production') {
    return SEQUENCE_PREFIX + padEndHash(hash(classes + dir));
  }

  return (
    SEQUENCE_PREFIX +
    padEndHash(hash(classes + dir)) +
    DEBUG_SEQUENCE_SEPARATOR +
    padEndHash(hash(sequenceIds.join('')))
  );
```

packages/core/src/makeStyles.test.ts
```diff
@@ -22,7 +22,7 @@ describe('makeStyles', () => {
         color: 'red',
       },
     });
-    expect(computeClasses({ dir: 'ltr', renderer }).root).toEqual('___afhpfp0 fe3e8s9');
+    expect(computeClasses({ dir: 'ltr', renderer }).root).toEqual('___afhpfp0_0000000 fe3e8s9');
```
layershifter, line 25: Let's a test to test hash generation in production:

```diff
describe('makeStyles', () => {
  let renderer: GriffelRenderer;

  beforeEach(() => {
+   process.env.NODE_ENV = 'test';
    renderer = createDOMRenderer(document);
  });
```

```ts
  it('in production mode hashes do not include debug information', () => {
    process.env.NODE_ENV = 'production';
    const computeClasses = makeStyles({
      root: { color: 'red' },
    });

    expect(computeClasses({ dir: 'ltr', renderer }).root).toEqual('___afhpfp0 fe3e8s9');
  });
```

packages/core/src/mergeClasses.test.ts
```diff
@@ -9,6 +9,8 @@ const options: MakeStylesOptions = {
   renderer: createDOMRenderer(document),
 };
 
+const EMPTY_STRING_HASH = '0000000';
```
layershifter, line 12: Not sure that these changes in tests are good. I propose:

- run them in production mode
  ```ts
  beforeEach(() => {
    process.env.NODE_ENV = 'test';
  });
  ```
- have a separate test for dev

packages/core/src/renderer/createDOMRenderer.ts
```diff
@@ -1,3 +1,6 @@
+import { injectDevTools } from '../devtools/injectDevTools';
+import { isDevToolsEnabled } from '../devtools/isDevToolsEnabled';
+import { MK_DEBUG } from '../devtools/store';
```
layershifter, line 3: Let's create `./devtools/index.ts` that exports _public API_ to be used in other lib files.

```suggestion
import { injectDevTools, isDevToolsEnabled, MK_DEBUG } from '../devtools';
```

packages/core/src/renderer/createDOMRenderer.ts
```diff
@@ -69,6 +73,11 @@ export function createDOMRenderer(
     },
   };
 
+  process.env.NODE_ENV !== 'production' &&
+    isDevToolsEnabled &&
+    target?.defaultView &&
+    injectDevTools(target.defaultView);
```
layershifter, line 79: nit: Can you please use `if` conditions instead of ternaries?

packages/core/src/makeStyles.ts
```diff
@@ -25,10 +27,22 @@ export function makeStyles<Slots extends string | number>(stylesBySlots: StylesB
     if (isLTR) {
       if (ltrClassNamesForSlots === null) {
         ltrClassNamesForSlots = reduceToClassNameForSlots(classesMapBySlot, dir);
+
+        process.env.NODE_ENV !== 'production' &&
+          isDevToolsEnabled &&
+          Object.entries(ltrClassNamesForSlots!).forEach(([slotName, sequenceHash]) => {
+            MK_DEBUG.addSequenceDetails(sequenceHash as SequenceHash, slotName);
+          });
```
layershifter, line 35:
```suggestion
   MK_DEBUG.addSequenceDetails(ltrClassNamesForSlots);
```

Is there sense to change signature of `addSequenceDetails` to handle this instead? (few lines below the loop happens for `rtlClassNamesForSlots`)

packages/core/src/devtools/isDevToolsEnabled.ts
```diff
@@ -0,0 +1,13 @@
+const isEnabled = () => {
+  let griffelDevtoolsEnabled = false;
+  if (process.env.NODE_ENV !== 'production') {
```
layershifter, line 3: Is there sense in this check? This should be never in production bundle, right?

packages/core/src/devtools/isDevToolsEnabled.ts
```diff
@@ -0,0 +1,13 @@
...
+    try {
+      griffelDevtoolsEnabled = !!window.localStorage['griffelDebug'];
```
layershifter, line 5: I am not entirely sure that we will use `localStorage` for this. Is there sense for now to export `false`?

```ts
export const isDevToolsEnabled = false
```

And once we will have an extension we can use the real condition.

packages/core/src/devtools/types.ts
```diff
@@ -0,0 +1,23 @@
...
+declare global {
+  interface Window {
+    __GRIFFEL_DEVTOOLS__: {
+      getInfo: (el: HTMLElement) => DebugResult | undefined;
+    };
+  }
+}
+
+export type DebugAtomicClassName = { className: string; overriddenBy?: string };
+export type DebugCSSRules = Record<string, string>; // className -> cssRule
```
layershifter, line 12:
```suggestion
export type DebugCSSRules = Record<string /* className */, string /* cssRule */>;
```

nit

packages/core/src/devtools/store.ts
```diff
@@ -0,0 +1,39 @@
+import { SequenceHash } from '../types';
+import { SEQUENCE_PREFIX, SEQUENCE_SIZE } from '../constants';
+import { mergeClassesCachedResults } from '../mergeClasses';
+
+const sequenceDetails: Record<SequenceHash, { slotName: string }> = {};
+
+const cssRules: string[] = [];
+
+export const MK_DEBUG = {
```
layershifter, line 9:
```suggestion
export const debugData = {
```

Let's use something more meaningful for name (opened for other options)

packages/core/src/devtools/store.test.ts
```diff
@@ -0,0 +1,66 @@
+import { SEQUENCE_PREFIX } from '../constants';
+import { makeStyles } from '../makeStyles';
+import { mergeClasses } from '../mergeClasses';
+import { createDOMRenderer } from '../renderer/createDOMRenderer';
+import { MakeStylesOptions } from '../types';
+import { MK_DEBUG } from './store';
+
+jest.mock('./isDevToolsEnabled.ts', () => ({
```
layershifter, line 8:
```suggestion
jest.mock('./isDevToolsEnabled', () => ({
```

nit: It should work even without an extension

packages/core/src/devtools/injectDevTools.ts
```diff
@@ -0,0 +1,84 @@
...
+  if (!window || typeof window === 'undefined') {
+    return;
+  }
+
+  // eslint-disable-next-line no-prototype-builtins
+  if (window.hasOwnProperty('__GRIFFEL_DEVTOOLS__')) {
+    return;
+  }
+
+  const devtools = {
```
layershifter, line 65: Let's make it stricter

```suggestion
  const devtools: typeof window['__GRIFFEL_DEVTOOLS__'] = {
```

packages/core/src/devtools/injectDevTools.ts
```diff
@@ -0,0 +1,84 @@
...
+      if (rootDebugSequenceHash === undefined) {
+        return undefined;
+      }
+
+      return getDebugTree(rootDebugSequenceHash);
+    },
+  };
+
+  Object.defineProperty(window, '__GRIFFEL_DEVTOOLS__', {
+    configurable: true,
```
layershifter, line 79:
```suggestion
    enumerable: false,
    configurable: false,
```

We probably don't want to make it `configurable`

packages/core/src/devtools/injectDevTools.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  return node;
+};
+
+export function injectDevTools(window: (Window & typeof globalThis) | null) {
+  if (process.env.NODE_ENV === 'production') {
+    return;
+  }
+
```
layershifter, line 55:
```suggestion
```

Do we really need this check there?

packages/core/src/devtools/injectDevTools.ts
```diff
@@ -0,0 +1,84 @@
...
+  if (!window || typeof window === 'undefined') {
+    return;
+  }
+
+  // eslint-disable-next-line no-prototype-builtins
+  if (window.hasOwnProperty('__GRIFFEL_DEVTOOLS__')) {
+    return;
+  }
```
layershifter, line 63:
```suggestion
    if (!window || window.hasOwnProperty('__GRIFFEL_DEVTOOLS__')) {
    return;
  }
```

This can be a single check, right?

packages/core/src/devtools/injectDevTools.ts
```diff
@@ -0,0 +1,84 @@
+import { DEFINITION_LOOKUP_TABLE, SEQUENCE_PREFIX } from '../constants';
+import { LookupItem, SequenceHash } from '../types';
+import { MK_DEBUG } from './store';
+import { DebugSequence } from './types';
+import { getDebugClassNames } from './utils';
+
+const getDebugTree = (debugSequenceHash: SequenceHash, parentNode?: DebugSequence) => {
+  const lookupItem: LookupItem | undefined = DEFINITION_LOOKUP_TABLE[debugSequenceHash];
```
layershifter, line 8: Let';s move this function to a separate file. Then you can separate tests in `packages/core/src/devtools/injectDevTools.test.ts`

- `injectDevTools` should a property to a passed `window`
- `getDebugTree` does traversal

layershifter (commented):
Great work, mostly nits 🚀

packages/core/src/devtools/injectDevTools.ts
```diff
@@ -0,0 +1,30 @@
+import { SEQUENCE_PREFIX } from '../constants';
+import { getDebugTree } from './getDebugTree';
+
+export function injectDevTools(document: Document) {
+  const window = document.defaultView;
+  if (!window || Object.prototype.hasOwnProperty.call(window, '__GRIFFEL_DEVTOOLS__')) {
```
layershifter, line 6:
```suggestion
  if (!window || window.__GRIFFEL_DEVTOOLS__)) {
```

Should it be really `.hasOwnProperty`?

layershifter, line 6:
```suggestion
  if (!window || window.__GRIFFEL_DEVTOOLS__)) {
```

Should it be really `.hasOwnProperty`?

packages/core/src/makeStyles.ts
```diff
@@ -32,6 +33,11 @@ export function makeStyles<Slots extends string | number>(stylesBySlots: StylesB
       }
     }
 
+    if (process.env.NODE_ENV !== 'production' && isDevToolsEnabled) {
+      const classNamesForSlots = isLTR ? ltrClassNamesForSlots : rtlClassNamesForSlots;
+      debugData.addSequenceDetails(classNamesForSlots!);
+    }
+
```
layershifter, line 40: I suggest to move it to bottom:

```ts
    const classNamesForSlots = isLTR
      ? (ltrClassNamesForSlots as Record<Slots, string>)
      : (rtlClassNamesForSlots as Record<Slots, string>);

    if (process.env.NODE_ENV !== 'production' && isDevToolsEnabled) {
      debugData.addSequenceDetails(classNamesForSlots);
    }

    return classNamesForSlots;
```

packages/core/src/mergeClasses.ts
```diff
@@ -139,7 +137,11 @@ export function mergeClasses(): string {
   let atomicClassNames = reduceToClassName(resultDefinitions, dir!);
 
   // Each merge of classes generates a new sequence of atomic classes that needs to be registered
-  const newSequenceHash = hashSequence(atomicClassNames, dir!);
+  const newSequenceHash = hashSequence(
+    atomicClassNames,
+    dir!,
+    process.env.NODE_ENV === 'production' ? [] : sequencesIds,
```
layershifter, line 143:
```suggestion
    sequencesIds,
```

nit: just checked this part, it seems that we can just pass it always without checks

layershifter, line 143:
```suggestion
    sequencesIds,
```

nit: just checked this part, it seems that we can just pass it always without checks

packages/core/src/mergeClasses.test.ts
```diff
@@ -125,12 +154,16 @@ describe('mergeClasses', () => {
       const rtlClasses1 = computeClasses({ ...options, dir: 'rtl' });
       const rtlClasses2 = computeClasses({ ...options, dir: 'rtl' });
 
-      expect(mergeClasses(rtlClasses1.start, rtlClasses2.start)).toBe(rtlClasses1.start);
-      expect(mergeClasses(rtlClasses1.start, rtlClasses2.start)).toBe(rtlClasses2.start);
-
-      expect(mergeClasses(rtlClasses1.start, rtlClasses2.start, rtlClasses1.end, rtlClasses2.end)).toBe(
-        '___1soy3ld f93e62u fo2qazs',
+      expect(mergeClasses(rtlClasses1.start, rtlClasses2.start)).toMatchInlineSnapshot(
+        removeSequenceHash(rtlClasses1.start),
```
layershifter, line 158:
```suggestion
       rtlClasses1.start,
```

packages/core/src/mergeClasses.test.ts
```diff
@@ -125,12 +154,16 @@ describe('mergeClasses', () => {
...
       );
+      expect(mergeClasses(rtlClasses1.start, rtlClasses2.start)).toMatchInlineSnapshot(
+        removeSequenceHash(rtlClasses2.start),
```
layershifter, line 161:
```suggestion
       rtlClasses2.start,
```

packages/core/src/mergeClasses.test.ts
```diff
@@ -142,7 +175,7 @@ describe('mergeClasses', () => {
       const sequence1 = mergeClasses('ui-button', classes.block);
       const sequence2 = mergeClasses(sequence1, classes.grid);
 
-      expect(sequence2).toBe(`ui-button ${classes.grid}`);
+      expect(sequence2).toMatchInlineSnapshot(`ui-button ${removeSequenceHash(classes.grid)}`);
```
layershifter, line 178:
```suggestion
      expect(sequence2).toMatchInlineSnapshot(`ui-button ${classes.grid}`);
```

packages/core/src/mergeClasses.test.ts
```diff
@@ -43,13 +62,17 @@ describe('mergeClasses', () => {
 
     const resultClassName = makeStyles({ root: { display: 'grid', paddingLeft: '5px' } })(options).root;
 
-    expect(mergeClasses(classes.block, classes.flex, classes.grid, classes.padding)).toBe(resultClassName);
+    expect(mergeClasses(classes.block, classes.flex, classes.grid, classes.padding)).toMatchInlineSnapshot(
+      removeSequenceHash(resultClassName),
```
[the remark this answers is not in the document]
layershifter: As we call `removeSequenceHash()` in `mergeClassesSequenceHashRemoval` we don't need to call it in `toMatchInlineSnapshot`

layershifter (approved):
Please address minor changes and we are ready to go 🚀

packages/core/src/mergeClasses.production.test.ts
```diff
@@ -0,0 +1,145 @@
+import { createDOMRenderer } from './renderer/createDOMRenderer';
```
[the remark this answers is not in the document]
YuanboXue-Amber: Oops, right

PR #93, opened 2022-04-19
feat: add @fluentui/react-components to handled imports by default

layershifter:
This PR adds `@fluentui/react-components` to handled imports by default in Babel plugin & Webpack loader.

Also fixes an issue with handling of `require()` calls: `modules` option was ignored before.


packages/babel-preset/src/transformPlugin.ts
```diff
@@ -100,7 +109,10 @@ export const transformPlugin = declare<Partial<BabelPluginOptions>, PluginObj<Ba
 
   const pluginOptions: Required<BabelPluginOptions> = {
     babelOptions: {},
-    modules: [{ moduleSource: '@griffel/react', importName: 'makeStyles' }],
+    modules: [
+      { moduleSource: '@griffel/react', importName: 'makeStyles' },
+      { moduleSource: '@fluentui/react-components', importName: 'makeStyles' },
```
ling1726, line 114: does this really make sense to couple this explicitly to Fluent ?

layershifter: Well, as Fluent is currently a single real consumer - yes...

ling1726: I mean the moduleSource is configurable for that reason, but I don't really have too strong opinion on it 🤷

ling1726 (approved):

packages/babel-preset/src/transformPlugin.ts
layershifter: From the internal chat:

> has there been any thought into adding imports from @fluentui/react-components as the "default" instead of just from griffel/react? we ended up forgetting to add this moduleSource property when we moved over to react-components so our build styles have been broken for a week+. we (I) definitely should have made a note in our work item but obviously just ended up forgetting

I think that it's okay for now and in v2 we can reconsider this. At least, there is no harm caused by this change.

PR #98, opened 2022-04-21
feat(Devtools): UI to display monolithic css rules per makeStyle slot 

YuanboXue-Amber:
Add real UI for griffel devtools.
Screenshot 2022-04-21 at 11 35 28

- show all makeStyles calls with slot name and monolithic css rules
    - atomic css rules from devtools result are processed into monolithic css objects using `stylis`
- displayed overridden css rules; clicking on them highlights the overriding rules

TODOs on devtools UI
- Show code highlights for css rules
- Add search function
- Click slot name expand/collapse rules within
- Show className for each rule
- Collapse shorthands in devtools UI

TODOs on devtools data:
- currently devtools is always disabled from @griffel/core, find out a way to enable it that is compatible with SSR
- Add sourcemap to devtools data; jump to source code from devtools UI

### DIFF
--- a/packages/devtools/src/FlattenView.tsx
+++ b/packages/devtools/src/FlattenView.tsx
@@ -1,0 +1,55 @@
+import * as React from 'react';
+import { makeStaticStyles, makeStyles, shorthands } from '@griffel/react';
+
+import { SlotCSSRules } from './SlotCSSRules';
+import { getRulesBySlots } from './utils';
+import { ViewContext } from './ViewContext';
+
+import type { DebugResult } from '@griffel/core';
+import type { SlotInfo } from './types';
+
+const View: React.FC<{ slots: SlotInfo[] }> = ({ slots }) => {
+  const [highlightedClass, setHighlightedClass] = React.useState('');
+  const contextDefault = React.useMemo(() => ({ highlightedClass, setHighlightedClass }), [highlightedClass]);
+
+  return (
+    <ViewContext.Provider value={contextDefault}>
+      {slots.map(({ slot, rules }) => (
+        <SlotCSSRules key={slot} slot={slot} atomicRules={rules} />
+      ))}
+    </ViewContext.Provider>
+  );
+};
+
+const useStyles = makeStyles({
+  root: {
+    paddingBottom: '10px',
+  },
+  info: {
+    ...shorthands.margin(0, '5px'),
+  },
+});
+
+const useStaticStyles = makeStaticStyles({
+  pre: {
+    fontFamily: 'Menlo, Consolas, "dejavu sans mono", monospace',
+    margin: 'initial',
+    whiteSpace: 'normal',
+    fontSize: '11px',
+    lineHeight: '14px',
+  },
+});
+
+export const FlattenView = ({ debugResultRoot }: { debugResultRoot: DebugResult }) => {
+  const slots = React.useMemo(() => getRulesBySlots(debugResultRoot), [debugResultRoot]);
+
+  useStaticStyles();
+  const classes = useStyles();
+
+  return (
+    <div className={classes.root}>
+      <div className={classes.info}>direction: {debugResultRoot.direction}</div>
+      <View slots={slots} />
+    </div>
+  );
+};
--- a/packages/devtools/src/FlattenView.tsx
+++ b/packages/devtools/src/FlattenView.tsx
@@ -40,7 +40,10 @@ const useStaticStyles = makeStaticStyles({
   },
 });
 
-export const FlattenView = ({ debugResultRoot }: { debugResultRoot: DebugResult }) => {
+type FlattenViewProps = { debugResultRoot: DebugResult };
+
+export const FlattenView: React.FC<FlattenViewProps> = props => {
+  const { debugResultRoot } = props;
   const slots = React.useMemo(() => getRulesBySlots(debugResultRoot), [debugResultRoot]);
 
   useStaticStyles();
--- a/packages/devtools/src/FlattenView.tsx
+++ b/packages/devtools/src/FlattenView.tsx
@@ -10,10 +10,10 @@ import type { SlotInfo } from './types';
 
 const View: React.FC<{ slots: SlotInfo[] }> = ({ slots }) => {
   const [highlightedClass, setHighlightedClass] = React.useState('');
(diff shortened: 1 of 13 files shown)


packages/devtools/src/FlattenView.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import type { DebugResult } from '@griffel/core';
+import type { SlotInfo } from './types';
+
+const View: React.FC<{ slots: SlotInfo[] }> = ({ slots }) => {
+  const [highlightedClass, setHighlightedClass] = React.useState('');
+  const contextDefault = React.useMemo(() => ({ highlightedClass, setHighlightedClass }), [highlightedClass]);
+
+  return (
+    <ViewContext.Provider value={contextDefault}>
```
layershifter, line 16:
```suggestion
  const contextValue = React.useMemo(() => ({ highlightedClass, setHighlightedClass }), [highlightedClass]);

  return (
    <ViewContext.Provider value={contextValue}>
```

_nit_

packages/devtools/src/FlattenView.tsx
```diff
@@ -0,0 +1,55 @@
...
+  pre: {
+    fontFamily: 'Menlo, Consolas, "dejavu sans mono", monospace',
+    margin: 'initial',
+    whiteSpace: 'normal',
+    fontSize: '11px',
+    lineHeight: '14px',
+  },
+});
+
+export const FlattenView = ({ debugResultRoot }: { debugResultRoot: DebugResult }) => {
```
layershifter, line 43:
```suggestion
type FlattenViewProps = { debugResultRoot: DebugResult };

export const FlattenView: React.FC<FlattenViewProps> = props => {
  const { debugResultRoot } = props;
```

_nit_

packages/devtools/src/RulesView.tsx
```diff
@@ -0,0 +1,95 @@
...
+  },
+});
+
+type MonolithicRulesViewProps = {
+  rules: MonolithicRules;
+  // specify if there should be spacing between rules with different selectors
+  noLineSpacing?: boolean;
+};
+
+export const MonolithicRulesView: React.FC<MonolithicRulesViewProps> = ({ rules, noLineSpacing }) => {
```
layershifter, line 73:
```suggestion
export const MonolithicRulesView: React.FC<MonolithicRulesViewProps> = (props) => {
  const { rules, noLineSpacing } = props
```

_nit_

layershifter, line 73: nit: Can either rename a file to `MonolithicRulesView.tsx` or a component to `RulesView.tsx`?

packages/devtools/src/getMonolithicCSSRules.ts
```diff
@@ -0,0 +1,55 @@
...
+  const nestedSelector = stylis.tokenize(classNameSelector).slice(1).join('');
+  monolithicRules[nestedSelector] = monolithicRules[nestedSelector] ?? [];
+
+  (children as stylis.Element[]).forEach(child => {
+    (monolithicRules[nestedSelector] as RuleDetail[]).push({
+      css: child.value,
+      className: stylis.tokenize(classNameSelector)[0].slice(1),
+      overriddenBy,
+    });
+  });
```
layershifter, line 18:
```suggestion
  if (Array.isArray(children)) {
    children.forEach(child => {
      (monolithicRules[nestedSelector] as RuleDetail[]).push({
        css: child.value,
        className: stylis.tokenize(classNameSelector)[0].slice(1),
        overriddenBy,
      });
    });
  } else {
    throw new Error('Unsupported input from "element.children"');
  }
```

I would use an explicit check to avoid typecasts

packages/devtools/src/getMonolithicCSSRules.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+function parseStylisElement(monolithicRules: MonolithicRules, element: stylis.Element, overriddenBy?: string) {
+  const { type } = element;
+  if (type.startsWith('@')) {
+    parseAtElement(monolithicRules, element, overriddenBy);
+  } else if (type === 'rule') {
+    parseRuleElement(monolithicRules, element, overriddenBy);
+  }
```
layershifter, line 43:
```suggestion
  if (type.startsWith('@')) {
    parseAtElement(monolithicRules, element, overriddenBy);
  } else if (type === 'rule') {
    parseRuleElement(monolithicRules, element, overriddenBy);
  } else {
    throw new Error('Unsupported type of entries');
  }
```

Let's throw in `else` branch

packages/devtools/src/RulesView.tsx
YuanboXue-Amber: I'm going for filename MonolithicRulesView.tsx

packages/devtools/src/FlattenView.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { makeStaticStyles, makeStyles, shorthands } from '@griffel/react';
+
+import { SlotCSSRules } from './SlotCSSRules';
+import { getRulesBySlots } from './utils';
+import { ViewContext } from './ViewContext';
+
+import type { DebugResult } from '@griffel/core';
+import type { SlotInfo } from './types';
+
+const View: React.FC<{ slots: SlotInfo[] }> = ({ slots }) => {
```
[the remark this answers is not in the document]
YuanboXue-Amber: Heyy I don't see any reason why not. This was a leftover from my prototyping. I'll use just one component for it

layershifter (approved):
I approved, but please check comments 😉

packages/devtools/src/getMonolithicCSSRules.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export function getMonolithicCSSRules(rules: AtomicRules[]): MonolithicRules {
+  const monolithicRules = {};
```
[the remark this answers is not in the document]
YuanboXue-Amber: I did some updates following this style, please take a look at commit [refactor getMonolithicCSSRules](https://github.com/microsoft/griffel/pull/98/commits/c583ddc4fd1fddbb5334edc5da7c517832a3edd3)

layershifter (approved):

Issue #78, opened 2022-03-17
Styles that need shorthands are not being errored when they are inside of selectors


khmakoto:
Styles like `padding` and `margin` which need the use of a shorthand to be expanded into their `left`, `right`, `top`, `bottom` equivalents error when you try to use them in a `makeStyles` call. However, if they are used inside of a selector that's not part of `CSS.Pseudos` then they do not error out and the styles are ultimately ignored.

We should fix the types so that these styles always error out to indicate the user that they need to change them to shorthands.

Example:

```ts
const useStyles = makeStyles({
  base: {
    ':hover': {
      // will error out
      padding: "1px"
    },

    [`& .${testClassName}`]': {
      // will not error out
      padding: "1px"
    }
  }
});
```

miroslavstastny:
Unfortunately we have not found a way how to avoid this in Typescript.
Currently Griffel already warns on the `padding` in development mode and never adds it to DOM (see #47).
We plan to add Lint rules to improve the build time check in #76.

layershifter:
We cannot do anything against it unfortunately. But the ESLint plugin with a matching rule was implemented in #99, you can find an example how it could be consumed [in docs](https://github.com/microsoft/griffel/tree/main/packages/eslint-plugin) or in the real PR (microsoft/fluentui#22961).

Issue #30, opened 2022-02-15
Document why CSS shorthands are forbidden


layershifter:
`README` file for [`@griffel/react`](https://github.com/microsoft/griffel/tree/main/packages/react) should contain limitations section.

It should include explainer why CSS shorthands are forbidden and reference shorthand functions.

layershifter:
> No support for shorthands came as a big shock. I'd definitely want to clearly understand why shorthands are forbidden before depending on the version of `@fluentui/react-components` that uses griffel.

Hey, thanks for feedback. We understand consequences and choose predictability as we cannot expand CSS shorthands properly. You can more context in the [matching RFC](https://github.com/microsoft/fluentui/blob/master/rfcs/react-components/convergence/no-css-shorthands-in-make-styles.md).

> This adds a huge amount of verbosity to one of my projects in a way that'd impact readability of the code. For example, border: '1px solid var(--colorNeutralForeground)' must become:

Totally agree, writing 12 CSS rules for a single border is not okay, that's why we have already introduced [some helpers](https://github.com/microsoft/griffel/tree/main/packages/core/src/shorthands):

```tsx
import { shorthands } from '@griffel/react';
// 👇 These helpers are also re-exported in Fluent UI itself
// import { shorthands } from '@fluentui/react-components';

makeStyles({
  root: {
    backgroundColor: 'red',
    display: 'flex',
    ...shorthands.border('1px', 'solid', 'red'),
    ...shorthands.margin('5px'),
    ...shorthands.padding('10px', '2px')
  }
})
```

The problem is that these functions are not documented (and reasoning, too). We will work on it.

thure:
Thank you for sharing the RFC, I understand the motivation for this design choice better now.

Offering a shorthands helper will certainly go a long way toward ameliorating the ergonomic consequences of this design choice, I'm glad to hear those will be made available alongside `makeStyles`.

PR #99, opened 2022-04-21
feat: create ESLint plugin for Griffel

### LINKED ISSUE #76: Lint rules for GriffelStyle
The shape of `GriffelStyle` is currently guarded only by TS typings. Does are sometimes difficult to understand and not always 100% accurate.
It would be great to add lint rules to improve strictness and readability of the error messages.

layershifter:
This PR adds `@griffel/eslint-plugin` with a single ESLint rule to enforce usage of CSS longhands.

Fixes #76.

## Rule Details

Griffel does not support usage of CSS shorthands, [microsoft/griffel#30](https://github.com/microsoft/griffel/issues/30#issuecomment-1045959297).
This is enforced by typings, but there are cases when they don't work as expected [microsoft/griffel#78](https://github.com/microsoft/griffel/issues/78).

Examples of **incorrect** code for this rule:

```js

  root: {
    background: 'red',
    padding: '10px 20px',
  },
});
```

image

Examples of **correct** code for this rule:

```js

  root: {
    backgroundColor: 'red',
    ...shorthands.padding('10px', '20px'),
  },
});
```

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -65,4 +65,5 @@ export type {
   GriffelRenderer,
 } from './types';
 
+// Private exports, are used by devtools
 export type { DebugCSSRules, DebugSequence, DebugResult } from './devtools';
--- a/packages/eslint-plugin/src/rules/no-shorthands.ts
+++ b/packages/eslint-plugin/src/rules/no-shorthands.ts
@@ -1,0 +1,122 @@
+import { ESLintUtils, TSESTree } from '@typescript-eslint/utils';
+import * as CSS from 'csstype';
+
+import { createRule } from '../utils/createRule';
+import { isIdentifier, isObjectExpression, isProperty } from '../utils/helpers';
+
+export const RULE_NAME = 'no-shorthands';
+
+// This collection is a map simply for faster access when checking if a CSS property is unsupported
+// @griffel/core has the same definition, but ESLint plugin should not depend on it
+const UNSUPPORTED_CSS_PROPERTIES: Record<keyof CSS.StandardShorthandProperties, true> = {
+  all: true,
+  animation: true,
+  background: true,
+  backgroundPosition: true,
+  border: true,
+  borderBlock: true,
+  borderBlockEnd: true,
+  borderBlockStart: true,
+  borderBottom: true,
+  borderColor: true,
+  borderImage: true,
+  borderInline: true,
+  borderInlineEnd: true,
+  borderInlineStart: true,
+  borderLeft: true,
+  borderRadius: true,
+  borderRight: true,
+  borderStyle: true,
+  borderTop: true,
+  borderWidth: true,
+  columns: true,
+  columnRule: true,
+  flex: true,
+  flexFlow: true,
+  font: true,
+  gap: true,
+  grid: true,
+  gridArea: true,
+  gridColumn: true,
+  gridRow: true,
+  gridTemplate: true,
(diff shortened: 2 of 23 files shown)


packages/eslint-plugin/src/utils/helpers.ts
```diff
@@ -0,0 +1,11 @@
...
+
+type IsHelper<NodeType extends AST_NODE_TYPES> = (node: TSESTree.Node | null | undefined) => node is TSESTree.Node & {
+  type: NodeType;
+};
+
+export const isIdentifier: IsHelper<AST_NODE_TYPES.Identifier> = ASTUtils.isIdentifier;
+export const isObjectExpression: IsHelper<AST_NODE_TYPES.ObjectExpression> = ASTUtils.isNodeOfType(
+  AST_NODE_TYPES.ObjectExpression,
+);
+export const isProperty: IsHelper<AST_NODE_TYPES.Property> = ASTUtils.isNodeOfType(AST_NODE_TYPES.Property);
```
layershifter, line 11: Had to type this manually otherwise TypeScript goes wild:

image

packages/core/src/index.ts
```diff
@@ -65,4 +65,5 @@ export type {
   GriffelRenderer,
 } from './types';
 
+// Private exports, are used by devtools
```
ling1726, line 68: maybe prefix with `@internal` and then strip on build and use `stripInternal` in the tsconfig ?

you'll be able to use it with aliases in the monorepo but the types will be stripped on build

packages/eslint-plugin/src/rules/no-shorthands.ts
```diff
@@ -0,0 +1,122 @@
...
+
+function findShorthandProperties(
+  node: TSESTree.ObjectExpression,
+  isRoot = false,
+  result: TSESTree.Identifier[] = [],
+): TSESTree.Identifier[] {
+  for (const propertyNode of node.properties) {
+    if (isProperty(propertyNode)) {
+      if (isIdentifier(propertyNode.key) && !isRoot) {
+        if (Object.prototype.hasOwnProperty.call(UNSUPPORTED_CSS_PROPERTIES, propertyNode.key.name)) {
```
ling1726, line 69: nit: these nested if blocks can just be reduced to a single one right ?

packages/eslint-plugin/src/utils/helpers.ts
```diff
@@ -0,0 +1,11 @@
+import { AST_NODE_TYPES, ASTUtils, TSESTree } from '@typescript-eslint/utils';
```
ling1726, line 1: is it necessary to use typescript-eslint to create the lint rule ?

layershifter: AFAIR there are no similar utils in ESLint itself. I was using [eslint-plugin-testing-library](https://github.com/testing-library/eslint-plugin-testing-library) as a reference.

packages/eslint-plugin/src/rules/no-shorthands.ts
layershifter: Yes, but IMO it's more readable

packages/core/src/index.ts
layershifter: Didn't know about `stripInternal`, it's a good idea. I will try it in another PR.

ling1726 (approved):

PR #100, opened 2022-04-22
feat: adds inset shorthand

### LINKED ISSUE #73: Add shorthand function for `inset`
Right now there's no shorthand in `@griffer/core` for [`inset`](https://developer.mozilla.org/en-US/docs/Web/CSS/Inset)

Setting `top`, `bottom`, `left`, and `right` for an absolute positioned element is a fairly common pattern. It would be great to expose a shorthand equivalent for this newer CSS rule.

### How it should work?
`inset` is CSS shorthand that is expanded to the `top`, `bottom`, `left`, and `right` properties. It follows the same format as `margin` and `padding`, so the logic from those shorthand functions can be reused.

bsunderhus:
Resolves #73

### DIFF
--- a/change/@griffel-core-e9730721-98ad-4971-8976-6f934c35cf1f.json
+++ b/change/@griffel-core-e9730721-98ad-4971-8976-6f934c35cf1f.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Adds inset shorthand",
+  "packageName": "@griffel/core",
+  "email": "bsunderhus@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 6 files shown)


change/@griffel-core-e9730721-98ad-4971-8976-6f934c35cf1f.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Adds inset shorthand",
```
layershifter, line 3:
```suggestion
  "comment": "nit: add shorthand function for inset",
```

layershifter (approved):
👍

PR #102, opened 2022-04-22
docs: add limitations & shorthands, minor improvements

### LINKED ISSUE #30: Document why CSS shorthands are forbidden

layershifter:
Fixes #30.

This PR:
- adds new page to website about `shorthands` & section to README
- adds limitations page that describes why CSS shorthands are forbidden
- minor improvements

### DIFF
--- a/apps/website/docs/react/api/shorthands.md
+++ b/apps/website/docs/react/api/shorthands.md
@@ -4,6 +4,12 @@ sidebar_position: 3
 
 # shorthands
 
+:::note
+
+Check [limitations](/react/guides/limitations) to understand why these helpers are needed.
+
+:::
+
 `shorthands` provides a set of functions to mimic [CSS shorthands](https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties) and improve developer experience as [CSS shorthands are not supported](/react/guides/limitations#css-shorthands-are-not-supported) by Griffel. For example:
 
 ```tsx
--- a/apps/website/docs/react/api/shorthands.md
+++ b/apps/website/docs/react/api/shorthands.md
@@ -1,0 +1,176 @@
+---
+sidebar_position: 3
+---
+
+# shorthands
+
+`shorthands` provides a set of functions to mimic [CSS shorthands](https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties) and improve developer experience as [CSS shorthands are not supported](/react/guides/limitations#css-shorthands-are-not-supported) by Griffel. For example:
+
+```tsx
+import { makeStyles, shorthands } from '@griffel/react';
+
+const useClasses = makeStyles({
+  root: {
+    // ❌ This is not supported, TypeScript compiler will throw, styles will not be inserted to DOM
+    padding: '2px 4px 8px 16px',
+    // ✅ Use shorthand functions to avoid writting CSS longhands
+    ...shorthands.padding('2px', '4px', '8px', '16px'),
+  },
+});
+```
+
+:::caution
+
+The most of the functions follow syntax in matching CSS properties, but each value should a separate argument:
+
+```js
+// ❌ Will produce wrong results
+shorthands.padding('2px 4px');
+// ✅ Correct output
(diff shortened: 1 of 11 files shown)


apps/website/docs/react/api/shorthands.md
```diff
@@ -0,0 +1,176 @@
+---
+sidebar_position: 3
+---
+
+# shorthands
```
bsunderhus, line 5: would be nice to have a link to `limitations.md` file

bsunderhus (approved):

apps/website/docs/react/guides/limitations.md
```diff
@@ -3,3 +3,88 @@ sidebar_position: 2
...
+
+```html
+<!-- Case 1: ❌ Wrong usage -->
+<div class="a b c">Hello world!</div>
+<!-- Case 2: ✅ Correct usage -->
+<div class="a c">Hello world!</div>
+<div class="b c">Hello world!</div>
+```
+
+- In "Case 1" both classes are applied to an element: it's wrong as result is **not deterministic** and depends on classes order in CSS definitions (i.e. [order of appearance](https://www.w3.org/TR/css-cascade-3/#cascade-order)), [demo on CodeSandbox](https://codesandbox.io/s/css-insertion-order-matters-mgt6y)
```
miroslavstastny, line 49:
```suggestion
- In "Case 1" both classes are applied to an element: it's wrong as result is **nondeterministic** and depends on classes order in CSS definitions (i.e. [order of appearance](https://www.w3.org/TR/css-cascade-3/#cascade-order)), [demo on CodeSandbox](https://codesandbox.io/s/css-insertion-order-matters-mgt6y)
```

apps/website/docs/react/guides/limitations.md
```diff
@@ -3,3 +3,88 @@ sidebar_position: 2
...
+  root: {
+    background: 'green',
+    backgroundColor: 'red',
+  },
+});
+```
+
+:::caution
+
+In this example the problem is the same: both classes will be applied, result depends on order of appearance.
```
miroslavstastny, line 84:
```suggestion
In this example the problem is the same: both classes will be applied, result depends on the order of appearance.
```

apps/website/docs/react/guides/limitations.md
```diff
@@ -3,3 +3,88 @@ sidebar_position: 2
...
+
+:::
+
+There is an option to expand CSS shorthands to longhands, but it's not reliable and does not with static rules i.e. you can't expand rules with CSS variables without knowing their value. The single predictable solution is disallow usage of CSS shorthands.
```
miroslavstastny, line 88:
```suggestion
There is an option to expand CSS shorthands to longhands, but it's not reliable and does not work with static rules i.e. you can't expand rules with CSS variables without knowing their value. The single predictable solution is disallow usage of CSS shorthands.
```

apps/website/docs/react/guides/limitations.md
```diff
@@ -3,3 +3,88 @@ sidebar_position: 2
...
+
+You can get more information on the original RFC, [microsoft/fluentui#20573](https://github.com/microsoft/fluentui/pull/20573).
```
miroslavstastny, line 90: I would add a sentence describing the solution, something like this:

For this reason Griffel disallows CSS shorthand properties in the input style object. Instead of shorthand properties, use `shorthands` helper functions which do the shorthand to longhand expansion statically. [link to shorthands docs]

miroslavstastny (approved):
We should have docsite container deploy as part of the PR CI :-(

apps/website/docs/react/guides/limitations.md
layershifter: Added 👍

apps/website/docs/react/api/shorthands.md
layershifter: Added 👍

PR #103, opened 2022-04-22
Fix transform plugin so it can handle more than one matched import sources

statm:
Currently, the transform plugin only keeps tab of the last import statement where the package name (`moduleSource`) is a match. For example, in:

```javascript
import { dupImport1 } from 'custom-package';
import { createStyles } from 'custom-package';
import { dupImport2 } from 'another-custom-package';
import { otherImport } from 'another-package';
```

with config:
```javascript
{
  "modules": [
    { "moduleSource": "custom-package", "importName": "createStyles" },
    { "moduleSource": "another-custom-package", "importName": "createMyStyles" }
  ]
}
```

The plugin will only focus on the 3rd import line (`import { dupImport2 } from 'another-custom-package';`) and then decide that there's no placement needed, thus missing the 2nd import where `createStyles` actually exists.

This PR fixes this issue by changing the `importDeclarationPath` into an `importDeclarationPaths` array, so we don't miss any imports to replace.

### DIFF
--- a/packages/babel-preset/__fixtures__/duplicated-imports/code.ts
+++ b/packages/babel-preset/__fixtures__/duplicated-imports/code.ts
@@ -1,9 +1,11 @@
 import { dupImport1 } from 'custom-package';
 import { createStyles } from 'custom-package';
-import { dupImport2 } from 'custom-package';
+import { dupImport2 } from 'another-custom-package';
+import { otherImport } from 'another-package';
 
 dupImport1;
 dupImport2;
+otherImport;
 
 export const useStyles = createStyles({
   root: { color: 'red' },
--- a/packages/babel-preset/__fixtures__/duplicated-imports/code.ts
+++ b/packages/babel-preset/__fixtures__/duplicated-imports/code.ts
@@ -1,24 +1,12 @@
 // @ts-expect-error This module does not exist, but will be resolved via aliases
 import { createStylesA } from 'custom-package';
 // @ts-expect-error This module does not exist, but will be resolved via aliases
-import { createStyles } from 'custom-package';
-// @ts-expect-error This module does not exist, but will be resolved via aliases
-import { createStylesB } from 'another-custom-package';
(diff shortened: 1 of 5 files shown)


packages/babel-preset/__fixtures__/duplicated-imports/code.ts
```diff
@@ -0,0 +1,12 @@
+import { dupImport1 } from 'custom-package';
+import { createStyles } from 'custom-package';
+import { dupImport2 } from 'another-custom-package';
+import { otherImport } from 'another-package';
+
+dupImport1;
+dupImport2;
+otherImport;
```
layershifter, line 8: Can you please turn these into real cals to be sure that transforms will be applied? 🐱 

```suggestion
export const useClassesA = createStylesA({
  root: { color: 'yellow' },
})
export const useClassesB = createStylesB({
  root: { color: 'blue' },
})
export const useClassesC = createStylesC({
  root: { color: 'orange' },
})
```

layershifter:
There are CI failures on non-existing imports:

```
__fixtures__/duplicated-imports/code.ts:3:28 - error TS2307: Cannot find module 'another-custom-package' or its corresponding type declarations.

3 import { dupImport2 } from 'another-custom-package';
```

You can use `@ts-expect-error` to get rid of them:

https://github.com/microsoft/griffel/blob/222a470836840b134075f231d19202aaea35b93f/packages/babel-preset/__fixtures__/import-custom-name/code.ts#L1-L2

packages/babel-preset/__fixtures__/duplicated-imports/code.ts
statm: Addressed as suggested. Thanks!

layershifter (approved):
@statm thanks for the contribution ❤️ I reviewed changes on fresh head and pushed some minor changes.

P.S. There is another problem (#104) that I found it during the review. Please let us know if it affects you.

PR #108, opened 2022-04-26
feat(Devtools): enable it via content_scripts

YuanboXue-Amber:
When griffel devtools start, content-script.js will run and add `__GRIFFEL_DEVTOOLS__` as `enabled` in sessionStorage.
Griffel core can use it to determine if devtools should be injected.

Reason for `sessionStorage`: content script runs in an isolated world. Changes like `document.XXX = yyy` in content script will not be reflected on the page (page will still have `document.XXX = undefined`)

### DIFF
--- a/packages/core/jest.config.js
+++ b/packages/core/jest.config.js
@@ -6,5 +6,4 @@ module.exports = {
   },
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx'],
   coverageDirectory: '../../coverage/packages/core',
-  globals: { window: {} },
 };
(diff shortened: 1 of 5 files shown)


packages/core/jest.config.js
```diff
@@ -6,4 +6,5 @@ module.exports = {
   },
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx'],
   coverageDirectory: '../../coverage/packages/core',
+  globals: { window: {} },
```
YuanboXue-Amber, line 9: Reason for adding jest globals:
without it there's error
image
And I think it's better to keep jest running in 'node' env

layershifter: Just use a `typeof window` check first 😉 As we will get the same error with SSR.

```
export const isDevToolsEnabled: boolean = typeof window !== 'undefined' && window?.sessionStorage?.getItem('__GRIFFEL_DEVTOOLS__');
```

YuanboXue-Amber: Ahh nice :)

miroslavstastny (approved):

layershifter (approved):

PR #110, opened 2022-04-26
Fix: Multiple selectors and RTL flipping

### LINKED ISSUE #105: Multiple selectors are not properly scoped
When writing multiple selectors separated by commas only the first selector is properly scoped with a class name.

[Here's a link to an annotated repro case in CodeSandbox](https://codesandbox.io/s/griffel-focus-within-tmfh2e?file=/src/App.tsx).

## Current Behavior

A Griffel styles like

```js
makeStyles({
  root: {
    ":active,:focus-within": {
      ...shorthands.borderColor("red")
    }
  }
});
```

Produces a selector like this in DevTools:

```
.f1mk9nzz:active, :focus-within {
    border-right-color: red;
}
```

Currently only the `:active` selector is scoped with a generated class name.

## Expected Behavior

Both the `:active` and `:focus-within` selectors should be scoped with the same generated class name. So we should see something like this in DevTools:

```
.f1mk9nzz:active, .f1mk9nzz:focus-within {
    border-right-color: red;
}
```

Alternatively, if this syntax is not supported I should get an error in VSCode explaining to me that I cannot write styles in this way.

### Notes

In Codesandbox I get an error (red squiggle under the selectors) when writing `":active,:focus-within"` indicating to me that perhaps this way of writing styles is not supported by Griffel. However, I don't get this error when writing styles like this in VS Code so I'm not certain. I'm using `griffel@1.0.3` in both Codesandbox and VSCode.

spmonahan:
Fixes #105 

## Current Behavior

When writing styles with multiple selectors and rules that will be flipped for RTL only the first selector is properly scoped to the generated class name.

```js
import { makeStyles } from '@griffel/core';

export default makeStyles({
  root: {
    ':active,:focus-within': {
      paddingLeft: '10px'
    }
  },
});
```

Generates

```
.f14f5aie:active,
.f14f5aie:focus-within {
  padding-left: 10px;
}

.f1sheuf0:active,
// 💣💥`:focus-with` rule applies globally! 😱
:focus-within {
  padding-right: 10px;
}
```

## New Behavior

```js
import { makeStyles } from '@griffel/core';

export default makeStyles({
  root: {
    ':active,:focus-within': {
      paddingLeft: '10px'
    }
  },
});
```

Generates

```
.f14f5aie:active,
.f14f5aie:focus-within {
  padding-left: 10px;
}

.f1sheuf0:active,
// `:focus-within` is properly scoped 👍👍
.f1sheuf0:focus-within {
  padding-right: 10px;
}
```

### DIFF
--- a/change/@griffel-core-94a0ec59-abfa-4d0f-a8a8-843026c57848.json
+++ b/change/@griffel-core-94a0ec59-abfa-4d0f-a8a8-843026c57848.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix issue with comma-separated selectors and RTL",
+  "packageName": "@griffel/core",
+  "email": "seanmonahan@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/core/src/runtime/compileCSS.ts
+++ b/packages/core/src/runtime/compileCSS.ts
@@ -102,7 +102,7 @@ export function compileCSS(options: CompileCSSOptions): [string /* ltr definitio
     cssRule = `${classNameSelector}{${normalizedPseudo} ${cssDeclaration}};`;
 
     if (rtlProperty) {
-      cssRule = `${cssRule}; ${rtlClassNameSelector}${normalizedPseudo} ${rtlCSSDeclaration};`;
+      cssRule = `${cssRule}; ${rtlClassNameSelector}{${normalizedPseudo} ${rtlCSSDeclaration}};`;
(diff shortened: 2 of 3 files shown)


miroslavstastny (approved):

bsunderhus (approved):

layershifter (approved):

layershifter:
Was fixed in `@griffel/react@1.0.4`.

PR #113, opened 2022-05-03
feat(devtools): add filter

YuanboXue-Amber:
Add filter for devtools:
Screenshot 2022-05-03 at 16 06 15

### DIFF
--- a/packages/devtools/src/utils.ts
+++ b/packages/devtools/src/utils.ts
@@ -1,5 +1,5 @@
 import type { DebugResult } from '@griffel/core';
-import { AtomicRules, SlotInfo } from './types';
+import { SlotInfo } from './types';
 
 export function getRulesBySlots(node: DebugResult, result: SlotInfo[] = []): SlotInfo[] {
   if (node.children.length === 0 && node.slot) {
@@ -29,18 +29,13 @@ export function filterSlots(slots: SlotInfo[], searchTerm: string) {
     return slots;
   }
 
-  const result: SlotInfo[] = [];
+  return slots.reduce<SlotInfo[]>((acc, { slot, rules }) => {
+    const filteredRules = rules.filter(rule => rule.cssRule.includes(searchTerm));
 
-  slots.forEach(({ slot, rules }) => {
-    const filteredRules: AtomicRules[] = [];
-    rules.forEach(rule => {
-      if (rule.cssRule.includes(searchTerm)) {
-        filteredRules.push(rule);
-      }
-    });
     if (filteredRules.length) {
-      result.push({ slot, rules: filteredRules });
+      return [...acc, { slot, rules: filteredRules }];
     }
-  });
-  return result;
+
+    return acc;
+  }, []);
 }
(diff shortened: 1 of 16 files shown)


packages/devtools/src/utils.ts
```diff
@@ -23,3 +23,15 @@ export function getRulesBySlots(node: DebugResult, result: SlotInfo[] = []): Slo
     return [...acc, ...getRulesBySlots(child)];
   }, result);
 }
+
+export function filterSlots(slots: SlotInfo[], searchTerm: string) {
+  if (!searchTerm.length) {
+    return slots;
+  }
+
+  const clone = JSON.parse(JSON.stringify(slots)) as SlotInfo[];
```
ling1726, line 32: the JSON functions are some of the most expensive ones in the browser, why do you need to deep clone the slot info ?

YuanboXue-Amber: heyy, the `slots` should not be changed because it contains all style information for the selected component. Search filters the slots and returns a subset. If I don't use a deep clone, changes will happen directly on 'slots', and when user clears filter I would have to reconstruct 'slots'.
I'm not happy with this implementation TBH, open for better suggestion.

YuanboXue-Amber: Just when I write reply to you I though about creating a new array as filtered result, instead of filtering on `slots` array. It is much faster. For 1000 simple search, JSON approach was 6.66ms and this one was 2.91ms

packages/devtools/src/utils.ts
```diff
@@ -23,3 +23,24 @@ export function getRulesBySlots(node: DebugResult, result: SlotInfo[] = []): Slo
...
 }
+
+export function filterSlots(slots: SlotInfo[], searchTerm: string) {
+  if (!searchTerm.length) {
+    return slots;
+  }
+
+  const result: SlotInfo[] = [];
+
+  slots.forEach(({ slot, rules }) => {
```
ling1726, line 34: nit: might be worth using a `map().filter(Boolean)` combination to be more 'pure function' programming

ling1726 (approved):

packages/devtools/src/utils.ts
ling1726: yeah, it is never better to serialize and deserialize JSON for a deep clone

packages/devtools/src/utils.ts
```diff
@@ -23,3 +23,24 @@ export function getRulesBySlots(node: DebugResult, result: SlotInfo[] = []): Slo
     return [...acc, ...getRulesBySlots(child)];
   }, result);
 }
+
+export function filterSlots(slots: SlotInfo[], searchTerm: string) {
```
layershifter, line 27: @YuanboXue-Amber you don't like `.reduce()`, right? 🐱 

```ts
  return slots.reduce<SlotInfo[]>((acc, { slot, rules }) => {
    const filteredRules = rules.filter(rule => rule.cssRule.includes(searchTerm));

    if (filteredRules.length) {
      return [...acc, { slot, rules: filteredRules }];
    }

    return acc;
  }, []);
```

YuanboXue-Amber: Haha yeah I didn't think of this is such a nice use case for reduce. Changes applied :)

layershifter (approved):

PR #114, opened 2022-05-03
feat(devtools): click slot name collapse/expand styles within

YuanboXue-Amber:
Clicking on slot name collapse/expand styles within:
Screenshot 2022-05-03 at 16 12 22

### DIFF
--- a/packages/devtools/src/SlotCSSRules.tsx
+++ b/packages/devtools/src/SlotCSSRules.tsx
@@ -27,7 +27,7 @@ export const SlotCSSRules: React.FC<{ slot: string; atomicRules: AtomicRules[] }
   const classes = useStyles();
 
   const [expanded, setExpanded] = React.useState(true);
-  const handleSlotNameClick = () => setExpanded(v => !v);
+  const toggleExpanded= () => setExpanded(v => !v);
 
   const { setHighlightedClass } = useViewContext();
   const undoHighlight = () => setHighlightedClass('');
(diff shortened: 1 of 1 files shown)


packages/devtools/src/SlotCSSRules.tsx
```diff
@@ -25,15 +26,22 @@ export const SlotCSSRules: React.FC<{ slot: string; atomicRules: AtomicRules[] }
   const rules = React.useMemo(() => getMonolithicCSSRules(atomicRules), [atomicRules]);
   const classes = useStyles();
 
+  const [expanded, setExpanded] = React.useState(true);
+  const handleSlotNameClick = () => setExpanded(v => !v);
```
ling1726, line 30:
```suggestion
  const toggleExpanded= () => setExpanded(v => !v);
```

ling1726 (approved):

packages/devtools/src/SlotCSSRules.tsx
```diff
@@ -25,15 +26,22 @@ export const SlotCSSRules: React.FC<{ slot: string; atomicRules: AtomicRules[] }
...
 
   return (
     <>
-      <pre className={classes.slotName}>{slot}</pre>
-      <div className={classes.rules} onClick={undoHighlight}>
-        <MonolithicRulesView rules={rules} />
-      </div>
+      <pre className={classes.slotName} onClick={toggleExpanded}>
+        {slot}
+      </pre>
```
layershifter, line 39: I think that are missing there an indicator to indicate if it's opened:

image

YuanboXue-Amber: good idea, let me create a separate PR for this

YuanboXue-Amber: PR #116

PR #115, opened 2022-05-04
feat(devtools): add tooltip to show className

YuanboXue-Amber:
Hover on each rule show a tooltip with className.
Tooltip is positioned using floating-ui.

Screenshot 2022-05-04 at 13 01 35

Screenshot 2022-05-04 at 12 24 01


YuanboXue-Amber:
I'm not sure if it is clear enough that the tooltip shows className.  Wondering if people would be confused when seeing them for the first time 🤔

layershifter:
Closing for housekeeping, we can reopen if it's still valuable.

PR #118, opened 2022-05-17
fix: update isDevToolsEnabled check

layershifter:
There are two issues with the existing check:
- it's not properly tree shaken with `webpack@4` & `terser-webpack-plugin@4`
  image
- it throws in incognito mode & with blocked third party cookies
  ```
  Uncaught DOMException: Failed to read the 'sessionStorage' property from 'Window': Access is denied for this document.
      at http://192.168.0.150:3000/iframe:18:28
      at http://192.168.0.150:3000/iframe:25:9
  ```

### DIFF
--- a/packages/core/src/devtools/isDevToolsEnabled.ts
+++ b/packages/core/src/devtools/isDevToolsEnabled.ts
@@ -1,8 +1,8 @@
-export const isDevToolsEnabled: Boolean = (() => {
+export const isDevToolsEnabled: boolean = (() => {
   // Accessing "window.sessionStorage" causes an exception when third party cookies are blocked
   // https://stackoverflow.com/questions/30481516/iframe-in-chrome-error-failed-to-read-localstorage-from-window-access-deni
   try {
-    return Boolean(typeof window !== 'undefined' && window?.sessionStorage?.getItem('__GRIFFEL_DEVTOOLS__'));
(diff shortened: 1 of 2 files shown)


packages/core/src/devtools/isDevToolsEnabled.ts
```diff
@@ -1,3 +1,9 @@
-export const isDevToolsEnabled = Boolean(
-  typeof window !== 'undefined' && window?.sessionStorage?.getItem('__GRIFFEL_DEVTOOLS__'),
-);
+export const isDevToolsEnabled: Boolean = (() => {
+  // Accessing "window.sessionStorage" causes an exception when third party cookies are blocked
+  // https://stackoverflow.com/questions/30481516/iframe-in-chrome-error-failed-to-read-localstorage-from-window-access-deni
+  try {
+    return Boolean(typeof window !== 'undefined' && window?.sessionStorage?.getItem('__GRIFFEL_DEVTOOLS__'));
```
ling1726, line 5:
```suggestion
    return Boolean(typeof window !== 'undefined' && window.sessionStorage?.getItem('__GRIFFEL_DEVTOOLS__'));
```

You already have an undefined guard for `window`

YuanboXue-Amber (approved):

ling1726 (approved):

PR #127, opened 2022-06-01
Akima lunar/126/add shorthands support for flex

### LINKED ISSUE #126: Add shorthands support for flex
# **[FEATURE]** Add `shorthands` support for `flex` property

Expand `...shorthands.flex(0, 0, 'auto')` into:

```js
{
  flexGrow: 0,
  flexShrink: 0,
  flexBasis: 'auto'
}
```

AkimaLunar:
# Adds `shorthands` support for the `flex` property

Closes #126 

### Usage

```js
shorthands.flex('auto')
shorthands.flex(1, '2.5rem')
shorthands.flex(0, 0, 'auto')
shorthands.flex([0, 0, 'auto'], [0, 0, '100vw'])
```

---

**Question**: Do you log into console when there are issues with expanding shorthands? I thought it would be nice to add a console message when `shorthands.flex` input doesn't match the format.


packages/core/src/shorthands/flex.ts
```diff
@@ -0,0 +1,163 @@
...
+  }
+};
+
+/**
+ * mergeFallbacks takes an array of FlexStyles and merges then into one object where each key can have an array of
+ * fallbacks.
+ * @param fallbackArray
+ * @returns FlexStyle with fallback arrays if needed.
+ */
+const mergeFallbacks = (fallbackArray: FlexInput[]) =>
```
ling1726, line 136: Can you provide some explanation about the fallback ? I don't see any mention of fallbacks for flex in the mdn docs

packages/core/src/index.ts
```diff
@@ -2,38 +2,40 @@
...
+  borderTop,
   borderWidth,
+  flex,
   gap,
-  margin,
-  padding,
-  overflow,
   inset,
+  margin,
   outline,
```
bsunderhus, line 17: This PR should not reorganize import order, this has nothing to do with `flex` shorthand implementation.

packages/core/src/index.ts
```diff
@@ -2,38 +2,40 @@
...
```
bsunderhus, line 36: This PR should not reorganize export order, this has nothing to do with `flex` shorthand implementation.

bsunderhus (dismissed):

packages/core/src/shorthands/flex.ts
AkimaLunar: I was following [this API design (CSS Fallback Properties)](https://griffel.js.org/react/api/make-styles#css-fallback-properties) so `shorthands.flex` can allow the user to provide multiple values as fallbacks. Here is the usage I intended:

_Input_
```ts
flex([0, 0, 'auto'], [0, 0, '100vw'])
```

_Output_
```css
.a {
    flex-grow: 0;
    flex-shrink: 1;
    flex-basis: 'auto';
    flex-basis: '100vw';
}
```

If you don’t think it is necessary, I can remove it. It doesn’t look like other `shorthands` utilities have support for that. If that is the way you want to go, I will only keep `generateFlex` and rename it `flex`.

---

P.S. It would be good to add support for fallbacks though. Probably better though a generic abstraction that could be applied to all other “expanders”.

layershifter: As you noticed we have support for fallback values, but it's not a goal of shorthands functions to do this.

I suggest to limit it to form of `flex(0, 0, ['auto', '100vw'])`, it will simplify implementation and will be consistent with other functions.

packages/react/README.md
```diff
@@ -148,6 +148,21 @@ const useClasses = makeStyles({
 });
 ```
 
+### `shorthands.flex`
```
layershifter, line 151: Can you please also add the same to `/apps/website/docs/react/api/shorthands.md`?

packages/core/src/shorthands/flex.ts
AkimaLunar: I've been pondering about this today… Letting the user pass arrays of values would make the expanding implementation actually more complex 🤔 With `flex` the type of the value and not the position defines what it is going to be when expanded. So if say someone passes `flex(1, ['10px', 2])` those should expand into:

```js
{
      flexGrow: 1,
      flexShrink: [1, 2],
      flexBasis: ['10px', 'auto'],
}
```
…because the second value could be both `flexShrink` and `flexBasis`. Same for say `['auto', 'none']`.

I can either add this logic _or_ restrict the usage of flex with three value syntax only. Although in the second case values like `flex('auto')`, flex('initial')`, etc. will fall through the cracks.

Personally, I would probably go for more complex logic for the sake of flexibility for the user, but will let you decide!

AkimaLunar: I would just need to rewrite the `mergeValue` and `mergeFallbacks` helpers differenty and replace `flex` with `generateFlex`.

layershifter: I have an opposite idea then: let's remove fallbacks for now due their complexity from `flex()` implementation.
It will be a bit inconsistent, but solves complexity issues. The general advice will be then: _You know about fallbacks, it's a rare case ➡️ you can use CSS longhands for them._

ling1726: > you can use CSS longhands for them.

@layershifter we'd need to explicitly enable CSS longhand for `flex` (they are stripped currently AFAIR)
https://github.com/microsoft/griffel/blob/0e24a70c7229e71f5828161e5616f3fe72e63e03/packages/core/src/constants.ts#L58

layershifter: @ling1726 why? What cannot be done with CSS longhands?

ling1726: ah sorry you meant explicitly use properties like `flexBasis`, my bad

AkimaLunar: @layershifter, yeah that makes sense 👌👌

packages/react/README.md
AkimaLunar: Done

packages/core/src/shorthands/flex.ts
AkimaLunar: Done

packages/core/src/shorthands/flex.ts
```diff
@@ -0,0 +1,121 @@
...
+export function flex(...values: FlexInput): FlexStyle {
+  const isOneValueSyntax = values.length === 1;
+  const isTwoValueSyntax = values.length === 2;
+  const isThreeValueSyntax = values.length === 3;
+
+  // One Value Syntax
+  if (isOneValueSyntax) {
+    const [firstValue] = values;
+
+    if (!Array.isArray(firstValue)) {
```
layershifter, line 41:
```suggestion
```

It cannot be an array, right?

AkimaLunar: That's the fallbacks leftovers, my bad! Thanks for catching 👌

packages/core/src/shorthands/flex.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    });
+  });
+
+  it('for a given values 2, 2', () => {
+    expect(flex(2, 2)).toEqual({
+      flexGrow: 2,
+      flexShrink: 2,
+      flexBasis: 'auto',
+    });
+  });
```
layershifter, line 74:
```suggestion
  it('for a given values 2, 2', () => {
    expect(flex(2, 2)).toEqual({
      flexGrow: 2,
      flexShrink: 2,
      flexBasis: 0,
    });
  });
```

[The spec](https://drafts.csswg.org/css-flexbox/#flex-common) is not really helpful there, but it seems that it should be `flexBasis: 0` in this case (the same behavior as in `flex: 2`):

image

image

Please correct me if I am wrong.

layershifter (commented):
One comment, other than that LGTM ❤️

packages/core/src/shorthands/flex.test.ts
AkimaLunar: Oh interesting… I was applying `auto`, because it is the initial value for `flex-basis`. But you are absolutely right! Will fix!

> ''flex: [<number [1,∞]>](https://drafts.csswg.org/css-values-4/#number-value)''
Equivalent to ''flex: [<number [1,∞]>](https://drafts.csswg.org/css-values-4/#number-value) 1 0''. Makes the flex item flexible and **sets the [flex basis](https://drafts.csswg.org/css-flexbox/#flex-flex-basis) to zero**, resulting in an item that receives the specified proportion of the free space in the flex container. If all items in the flex container use this pattern, their sizes will be proportional to the specified flex factor.

— [Flex property](https://drafts.csswg.org/css-flexbox/#flex-property)

AkimaLunar:
Question, @layershifter : Do you have some sort of logger in Griffel? I thought it would be nice DevX to have a warning in the console when they pass an un-parcerable value with the default fallback. When they _do_ pass something un-parcerable, should it return a initial (`{ flexGrow: 0,flexShrink: 1, flexBasis: 'auto' }`) or nothing (`{}`) you think?

layershifter:
> Question, @layershifter : Do you have some sort of logger in Griffel? I thought it would be nice DevX to have a warning in the console when they pass an un-parcerable value with the default fallback. When they _do_ pass something un-parcerable, should it return a initial (`{ flexGrow: 0,flexShrink: 1, flexBasis: 'auto' }`) or nothing (`{}`) you think?

Let's return nothing i.e. `{}` to clearly indicate that something went wrong. We don't have a logger, but you can add own messages like in this example (under `process.env.NODE_ENV` check as we don't want to see them in production 🐱 ):

https://github.com/microsoft/griffel/blob/01e52783012cb38d91b474858419d417b6c46410/packages/core/src/mergeClasses.ts#L68-L78

AkimaLunar:
> > Question, @layershifter : Do you have some sort of logger in Griffel? I thought it would be nice DevX to have a warning in the console when they pass an un-parcerable value with the default fallback. When they _do_ pass something un-parcerable, should it return a initial (`{ flexGrow: 0,flexShrink: 1, flexBasis: 'auto' }`) or nothing (`{}`) you think?
> 
> Let's return nothing i.e. `{}` to clearly indicate that something went wrong. We don't have a logger, but you can add own messages like in this example (under `process.env.NODE_ENV` check as we don't want to see them in production 🐱 ):
> 
> https://github.com/microsoft/griffel/blob/01e52783012cb38d91b474858419d417b6c46410/packages/core/src/mergeClasses.ts#L68-L78

Done 👌

layershifter (approved):

PR #130, opened 2022-06-04
Added shorthand for grid-area

GeoffCoxMSFT:
## Changes
- Added gridArea function with overloads for 1-4 values (row start, column start, row end, column end)
- Created logic based on grid-area MDN for missing values
- Added local helper function to determine if value is a custom-ident (could be pulled into a utils if other shorthands need it)
- Added unit tests for each overload

## Open issues
- The ValueOrArray generic did not see appropriate for the gridArea function, so I defined GridAreaInput as CSS.Property.GridArea.  I'm unsure if both value and arrays should be supported for gridArea.


miroslavstastny (commented):
The expansion depends on the value passed in, whether it is a custom ident or not. That is described in MDN and correctly implemented in this PR.

(fwiw, codesandbox demonstrating the difference: https://codesandbox.io/s/grid-area-expand-d1mfq9?file=/index.html)

The question is what should happen if a CSS variable is passed to the shorthand function - that can resolve to anything, only in runtime.

These are the possible options I see:

1. do not add the shorthand function as we cannot make it 100% correct
2. ignore the problem (CSS variable is currently considered a custom ident)
3. check there is a CSS variable passed to the shorthand function and throw in that case (is it a valid usecase to pass a theme token to `grid-area`? If so, isn't it just an edge case?)
4. blacklist the longhands and allow the shorthand (I would not expect you want to override/set only some of the longhands?)
5. have two separate functions, one for custom ident and the other one for non-custom ident (but developers would have hard time understanding)

miroslavstastny:
> 3. check there is a CSS variable passed to the shorthand function and throw in that case (is it a valid usecase to pass a theme token to `grid-area`? If so, isn't it just an edge case?)

`flex` shorthand has been merged recently in #127 which silently fails for `flex('var(--my-var)')`.
@layershifter, @GeoffCoxMSFT  I vote for `grid-area` to fail similarly, what do you think?

miroslavstastny:
Discussed offline with @layershifter and @ling1726 - let's check for the CSS variable in the value and fail silently.
Either we can check for `--` in the string or `var(`?

ling1726:
> Discussed offline with @layershifter and @ling1726 - let's check for the CSS variable in the value and fail silently. Either we can check for `--` in the string or `var(`?

https://github.com/kentcdodds/rtl-css-js/pull/62/files#diff-be37810fd676724e7edb9627d74b7583e7c38153b3804c2213e84ae32cf4a65fR14

That should work, it's in rtl-css-js and we've been using it internally for some time

GeoffCoxMSFT:
> > 3. check there is a CSS variable passed to the shorthand function and throw in that case (is it a valid usecase to pass a theme token to `grid-area`? If so, isn't it just an edge case?)
> 
> `flex` shorthand has been merged recently in #127 which silently fails for `flex('var(--my-var)')`. @layershifter, @GeoffCoxMSFT I vote for `grid-area` to fail similarly, what do you think?

I update the PR for the fail silently approach.  
I also added a console.error suppression for the flex shorthand tests to avoid failing tests when running `yarn test` locally.

I created this codepen to see how css vars are supported. https://codepen.io/GeoffCox/pen/xxYQpdG
We could have a 6th option: Check for css var and apply it to grid-row-start etc.

miroslavstastny:
> I created this codepen to see how css vars are supported. https://codepen.io/GeoffCox/pen/xxYQpdG
We could have a 6th option: Check for css var and apply it to grid-row-start etc.

I am sorry, I don't follow. The codepen shows that the browser expands based on the CSS variable value:

image

image

How are you suggesting to expand the css variable?

miroslavstastny (approved):

GeoffCoxMSFT:
> I am sorry, I don't follow. The codepen shows that the browser expands based on the CSS variable value:
> How are you suggesting to expand the css variable?

I'm saying that when we see `var(` in the parameter passed to the shorthand, we can output it on the individual css properties.  It would assume that the var is a value compatible with grid-row-start.  Basically pass through var references rather than erroring.

packages/core/src/shorthands/gridArea.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    });
+  });
+
+  it('for span custom-ident', () => {
+    expect(gridArea('header', 'span footer')).toEqual({
+      gridRowStart: 'header',
+      gridColumnStart: 'span footer',
+      gridRowEnd: 'header',
+      gridColumnEnd: 'auto',
+    });
```
layershifter, line 140:
```suggestion
    expect(gridArea('header', 'span body', 'span footer')).toEqual({
      gridRowStart: 'header',
      gridColumnStart: 'span body',
      gridRowEnd: 'span footer',
      gridColumnEnd: 'auto',
    });
```

Oops, copy-paste issue 😀 This test should have three arguments.

layershifter (approved):
Great work, approved with small suggestions 👍 

P.S. I also will push doc updates to website/README.

PR #133, opened 2022-06-13
experiment: Change DOM renderer API to accommodate extra metadata

ling1726:
Updates Griffel to work with a flat instead of nested structure for CSS
    rule to accommodate extra metadata for css rules during insertion.

The new datastructure leaves it open for each rule to store metadata which can be helpful to the renderer during CSS insertion


packages/core/src/renderer/createDOMRenderer.ts
```diff
@@ -40,41 +40,33 @@ export function createDOMRenderer(
...
-
-        // This is a hot path in rendering styles: ".length" is cached in "l" var to avoid accesses the property
-        for (let i = 0, l = cssRulesForBucket.length; i < l; i++) {
-          const ruleCSS = cssRulesForBucket[i];
-
-          if (renderer.insertionCache[ruleCSS]) {
-            continue;
-          }
+      for (let i = 0, l = cssRules.length; i < l; i++) {
+        const { cssRule: ruleCSS, bucket: styleBucketName } = cssRules[i];
```
ling1726, line 44: could be an array instead with numerical constants for faster access

layershifter:
Closing as we went with other option.

PR #135, opened 2022-06-14
feat: add Babel plugin for CSS extraction

layershifter:
This PR implements the Babel plugin that will be used for CSS extraction.

This plugin replaces `__styles` with `__css` function and removes CSS rules from output:

```tsx

export const stylesA = __styles(
  {
    root: {
      sj55zd: 'fe3e8s9',
    },
  },
  {
    d: ['.fe3e8s9{color:red;}'],
  },
);

export const stylesB = __styles(
  {
    root: {
      De3pzq: 'fcnqdeg',
    },
  },
  {
    d: ['.fcnqdeg{background-color:green;}'],
  },
);

//
// ⬇️⬇️⬇️
//

import { __styles, __css } from '@griffel/react';

export const stylesA = __css({
  root: {
    sj55zd: 'fe3e8s9',
  },
});

export const stylesB = __css({
  root: {
    De3pzq: 'fcnqdeg',
  },
});
```

CSS rules are returned as part of `meta` informantion.

### DIFF
--- a/packages/webpack-extraction-plugin/__fixtures__/babel/basic/output.ts
+++ b/packages/webpack-extraction-plugin/__fixtures__/babel/basic/output.ts
@@ -1,0 +1,11 @@
+import { __styles, __css } from '@griffel/react';
+export const styles = __css({
+  root: {
+    sj55zd: 'fe3e8s9',
+    uwmqm3: ['fycuoez', 'f8wuabp'],
+  },
+  icon: {
+    De3pzq: 'fcnqdeg',
+    Bi91k9c: 'faf35ka',
+  },
+});
--- a/packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
+++ b/packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
@@ -79,6 +79,7 @@ export const babelPluginStripGriffelRuntime = declare<
                 );
               }
 
+              // Returns the styles as a JavaScript object
               const evaluationResult = argumentPaths[1].evaluate();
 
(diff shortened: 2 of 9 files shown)


packages/webpack-extraction-plugin/__fixtures__/babel/basic/output.ts
```diff
@@ -0,0 +1,11 @@
+import { __styles, __css } from '@griffel/react';
```
ling1726, line 1:
```suggestion
import { __css } from '@griffel/react';
```

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
```diff
@@ -0,0 +1,106 @@
...
+              if (argumentPaths.length !== 2) {
+                throw calleePath.buildCodeFrameError(
+                  [
+                    '"__styles" function call should have exactly two arguments.',
+                    'Please report a bug (https://github.com/microsoft/griffel/issues) if this error happens',
+                  ].join(' '),
+                );
+              }
+
+              const evaluationResult = argumentPaths[1].evaluate();
```
ling1726, line 82:
```suggestion
              // returns the styles as a javascript object
              const evaluationResult = argumentPaths[1].evaluate();
```

packages/webpack-extraction-plugin/__fixtures__/babel/basic/output.ts
layershifter: It will be removed by tree shaking, I don't want to remove imports as something could go wrong :)

PR #136, opened 2022-06-14
Feat/sort media queries

ling1726:


packages/core/src/types.ts
```diff
@@ -137,7 +127,33 @@ export type CSSClasses = /* ltrClassName */ string | [/* ltrClassName */ string,
 export type CSSClassesMap = Record<PropertyHash, CSSClasses>;
 export type CSSClassesMapBySlot<Slots extends string | number> = Record<Slots, CSSClassesMap>;
 
-export type CSSRulesByBucket = Partial<Record<StyleBucketName, string[]>>;
+export type StyleBucketName = keyof CSSRulesByBucket;
+export type NestedCSSBucket = Record<string, string[]>;
+export type CSSRulesByBucket2 = Record<string, string[] | NestedCSSBucket>;
```
layershifter, line 132:
```suggestion
```

Dead code?

change/@griffel-babel-preset-25c907f6-ed33-4765-bb0b-2dd99c71200b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "feat: Add compareMediaQuery option to createDOMRenderer",
```
layershifter, line 3: Not sure that it describes what is happening.

packages/babel-preset/src/transformPlugin.ts
```diff
@@ -96,9 +96,17 @@ function isRequireDeclarator(
  */
 function dedupeCSSRules(cssRules: CSSRulesByBucket): CSSRulesByBucket {
```
layershifter, line 97: We can do it better. Let's move it to `packages/babel-preset/src/utils/dedupeRules.ts`
```ts
  export function dedupeCSSRules(cssRules: CSSRulesByBucket): CSSRulesByBucket {
    return Object.fromEntries(
      Object.entries(cssRules).map(([bucketName, cssRules]) => {
        if (Array.isArray(cssRules)) {
          return [bucketName, [...new Set(cssRules)]];
        }

        return [bucketName, dedupeCSSRules(cssRules)];
      }),
    );
  }
```

```ts
import { dedupeCSSRules } from './dedupeRules';

describe('dedupeCSSRules', () => {
  it('de-duplicates rules', () => {
    expect(dedupeCSSRules({ d: ['foo', 'bar'], k: ['foo', 'baz'] })).toEqual({ d: ['foo', 'bar'], k: ['foo', 'baz'] });
  });

  it('de-duplicates rules for media queries', () => {
    expect(dedupeCSSRules({ m: { foo: ['foo', 'bar'], baz: ['baz'] } })).toEqual({
      m: { foo: ['foo', 'bar'], baz: ['baz'] },
    });
  });
});

```

ling1726: I would prefer to keep the check on the bucket name to be very clear that this only happens for media queries

packages/core/src/common/snapshotSerializers.ts
```diff
@@ -53,7 +53,15 @@ export const griffelRulesSerializer: jest.SnapshotSerializerPlugin = {
...
+        rules = Object.keys(cssRulesByBucket[styleBucketName]!).reduce((mediaRules, mediaQuery) => {
+          mediaRules.push(...cssRulesByBucket[styleBucketName]![mediaQuery]);
+          return mediaRules;
+        }, [] as string[]);
+      } else {
+        rules = cssRulesByBucket[styleBucketName]!;
+      }
 
       return prettier.format(acc + rules.join(''), { parser: 'css' }).trim();
     }, '');
```
layershifter, line 67:
```ts
    const _value = value as StyleRulesTuple;
    const cssRulesByBucket = _value[1];

    function flattenRules(cssRulesByBucket: CSSRulesByBucket) {
      return Object.values(cssRulesByBucket)
        .map(cssRules => (Array.isArray(cssRules) ? cssRules : flattenRules(cssRules)))
        .flatMap(rules => rules);
    }

    return prettier.format(flattenRules(cssRulesByBucket).join(''), { parser: 'css' }).trim();
```

Let's also simplify this.

packages/babel-preset/src/transformPlugin.ts
layershifter: That could also work, not sure if TS will like it.

packages/core/src/renderer/createDOMRenderer.ts
```diff
@@ -32,51 +35,69 @@ export function createDOMRenderer(
   target: Document | undefined = typeof document === 'undefined' ? undefined : document,
   options: CreateDOMRendererOptions = {},
 ): GriffelRenderer {
-  const { unstable_filterCSSRule } = options;
+  const { unstable_filterCSSRule, compareMediaQuery = () => 1 } = options;
```
layershifter, line 38: We need to put something that works in a predictable way. For example:

```ts
(a, b) => a.name < b.name ? -1 : a.name > b.name ? 1 : 0;
```

PR #137, opened 2022-06-16
feat: Create Isomorphic stylesheet for style renderer

ling1726:
The isomorphic stylesheet is intended mainly for SSR to simplify
element rendering and make current and future (i.e. media queries) insertion more straightforward.

The SSR `renderToStyleElement` function no longer needs to know about insertion cache or specific buckets


packages/react/src/renderToStyleElements.ts
```diff
@@ -2,45 +2,37 @@ import { styleBucketOrdering } from '@griffel/core';
...
-      const cssRules = styles[bucketName].join('');
-
-      // We don't want to create empty style elements
-      if (cssRules.length === 0) {
+  const styleElements = Object.values(renderer.styleElements).sort((a, b) => {
+    return (
+      styleBucketOrdering.indexOf(a.bucketName as StyleBucketName) -
+      styleBucketOrdering.indexOf(b.bucketName as StyleBucketName)
+    );
+  });
```
layershifter, line 16:
```suggestion
  const styleElements = Object.values(renderer.styleElements).sort((a, b) => {
    return styleBucketOrdering.indexOf(a.bucketName) - styleBucketOrdering.indexOf(b.bucketName);
  });
```

packages/core/src/types.ts
```diff
@@ -85,6 +85,14 @@ export interface MakeStaticStylesOptions {
   renderer: GriffelRenderer;
 }
 
+export interface IsomorphicStyleSheet {
+  elementAttributes: Record<string, string>;
+  insertRule(rule: string): number | undefined;
+  element: HTMLStyleElement | undefined;
+  bucketName: string;
```
layershifter, line 92:
```suggestion
  bucketName: StyleBucketName;
```

packages/react/src/renderToStyleElements.ts
```diff
@@ -2,45 +2,37 @@ import { styleBucketOrdering } from '@griffel/core';
 import * as React from 'react';
 import type { GriffelRenderer, StyleBucketName } from '@griffel/core';
```
layershifter, line 3:
```suggestion
import type { GriffelRenderer } from '@griffel/core';
```

packages/core/src/types.ts
```diff
@@ -85,6 +85,14 @@ export interface MakeStaticStylesOptions {
   renderer: GriffelRenderer;
 }
 
+export interface IsomorphicStyleSheet {
+  elementAttributes: Record<string, string>;
```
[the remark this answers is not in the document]
ling1726: I would prefer to be explicit here

packages/core/src/renderer/getStyleSheetForBucket.ts
```diff
@@ -33,11 +34,19 @@ export const styleBucketOrdering: StyleBucketName[] = [
...
   renderer: GriffelRenderer,
   elementAttributes: Record<string, string> = {},
-): CSSStyleSheet {
+): IsomorphicStyleSheet {
   if (!renderer.styleElements[bucketName]) {
+    const tag: HTMLStyleElement | undefined = target && target.createElement('style');
+    const stylesheet = createIsomorphicStyleSheet(tag, bucketName, elementAttributes);
+    renderer.styleElements[bucketName] = stylesheet;
+
+    if (!target || !tag) {
```
layershifter, line 46:
```suggestion
    if (!tag) {
```

We are checking for `target` on line 42, right?

packages/core/src/renderer/rehydrateRendererCache.test.ts
```diff
@@ -0,0 +1,37 @@
...
+      }
+    `);
+  });
+
+  it('should create isomorphic stylesheet', () => {
+    const renderer = createDOMRenderer(document);
+    const styleElement = document.createElement('style');
+    document.head.appendChild(styleElement);
+    styleElement.setAttribute(DATA_BUCKET_ATTR, 'd');
+    styleElement.textContent = ".foo { color: 'red' }";
```
layershifter, line 26:
```suggestion

    document.head.appendChild(styleElement);
    styleElement.setAttribute(DATA_BUCKET_ATTR, 'd');

```

layershifter: We don't need `styleElement.textContent` in this test

layershifter (approved):

Issue #129, opened 2021-10-06
makeStyles: media queries should be sorted


layershifter:

ling1726:
Extra code repro to show that the same problem can occur even if the `makeStyles` call itself sorts media queries: https://codesandbox.io/s/media-order-forked-42wf1v?file=/src/App.js

PR #140, opened 2022-06-17
feat: Insert media queries into separate buckets

ling1726:
This PR introduces the concept of 'nested buckets' Media queries will be inserted into separate buckets based on the media query value. However, to keep existing categorization, media query buckets will all have a bucket value of `m`. Media buckets are always inserted after any other bucket. **Sorting the media buckets is out of scope of this PR**

```html
<! -- Before -->
<style data-make-styles-bucket="t">
  @media screen and (max-width: 1px) {
    .foo { color: red; }
  }

  @media screen and (max-width: 2px) {
    .foo { color: blue; }
  }

  @supports (display: grid) {
    div {
      display: grid;
    }
  }
</style>

<! -- After-->
<style data-make-styles-bucket="t">
  @supports (display: grid) {
    div {
      display: grid;
    }
  }
</style>

<style data-make-styles-bucket="m" media="screen and (max-width: 1px)">
  @media screen and (max-width: 1px) {
    .foo { color: red; }
  }
</style>

<style data-make-styles-bucket="m" media="screen and (max-width: 2px)">
  @media screen and (max-width: 2px) {
    .foo { color: blue; }
  }
</style>
```

Adds the concept of 'metadata' for entries in the CSS buckets. This allows the possibility of the bucket entry to be an array with an optional metadata object as the second array element. Currently only utilized for the media query buckets.

```js
// Before
const cssRulesByBucket = {
  d:[ '.foo { color: red; }' ],
  t: [ '@media screen and (max-width: 1px) { .foo { color: red; } }' ]
}

// After
const cssRulesByBucket = {
  d: [ '.foo { color: red; }' ],
  m: [ '@media screen and (max-width: 1px) { .foo { color: red; } }', metadata: { m: '@media screen and (max-width: 1px)' }]
}
```

Addresses: #129


packages/core/src/renderer/createDOMRenderer.ts
```diff
@@ -43,16 +44,17 @@ export function createDOMRenderer(
...
-          target,
-          renderer,
-          options.styleElementAttributes,
-        );
 
         // This is a hot path in rendering styles: ".length" is cached in "l" var to avoid accesses the property
         for (let i = 0, l = cssRulesForBucket.length; i < l; i++) {
-          const ruleCSS = cssRulesForBucket[i];
+          const [ruleCSS, metadata] = normalizeCSSBucketEntry(cssRulesForBucket[i]);
+          const sheet = getStyleSheetForBucket(
```
ling1726, line 51: should benchmark this with master to see if more function calls will affect hot path perf

ling1726: I ran 10 benchmark runs on the benchmark app for this branch and main and the numbers seem similar

main: 

image

PR:

griffel-before

packages/core/src/renderer/rehydrateRendererCache.test.ts
```diff
@@ -24,6 +23,7 @@ describe('rehydrateRendererCache', () => {
     const styleElement = document.createElement('style');
     document.head.appendChild(styleElement);
     styleElement.setAttribute(DATA_BUCKET_ATTR, 'd');
+    styleElement.textContent = ".foo { color: 'red' }";
```
layershifter, line 26:
```suggestion
```

Merge conflict?

packages/core/src/renderer/rehydrateRendererCache.test.ts
```diff
@@ -9,7 +9,6 @@ describe('rehydrateRendererCache', () => {
     styleElement.setAttribute(DATA_BUCKET_ATTR, 'd');
     document.head.appendChild(styleElement);
     styleElement.textContent = '.foo { color: red; }';
-
     rehydrateRendererCache(renderer, document);
```
layershifter, line 12: Merge conflict?

packages/core/src/renderer/getStyleSheetForBucket.ts
```diff
@@ -37,11 +39,21 @@ export function getStyleSheetForBucket(
   target: Document | undefined,
   renderer: GriffelRenderer,
   elementAttributes: Record<string, string> = {},
+  metadata?: Record<string, unknown>,
 ): IsomorphicStyleSheet {
-  if (!renderer.stylesheets[bucketName]) {
+  let stylesheetKey: StyleBucketName | string = bucketName;
+  if (bucketName === 'm' && metadata) {
+    stylesheetKey = metadata['m'] as string;
```
layershifter, line 46:
```suggestion
    stylesheetKey = bucketName + metadata['m'] as string;
```

Can we use `bucketName` as prefix rather than replace it completely?

layershifter (approved):

packages/core/src/renderer/rehydrateRendererCache.test.ts
ling1726: yeah, I tried to squash commits manually and it was an awful experience

PR #142, opened 2022-06-20
feat(devtools): jump to sourcemap

YuanboXue-Amber:
Add ability to jump to source for griffel devtools:
20YtOgbDZL

Source URL came from error stack.

TODO:
parse URL from error stack to get to original source
  - I've tried the parser in react-render-tracker. It has some limitations when used in this case:
     - it parses only base64 sourcemap (which is enough for TMP), we need to add ability to read extermal sourcemap
     - when using webpack, the parsed url from base64 sourcemap does not contain [namespace](https://webpack.js.org/configuration/output/#outputdevtoolmodulefilenametemplate) for some reason :/
     - it uses global `webpackChunk_<xxx>` object to resolve modules map. This object is in the host page's context, so some extra work is required to get this object from host using content-script.


packages/devtools/src/SlotCSSRules.tsx
```diff
@@ -69,3 +92,12 @@ export const SlotCSSRules: React.FC<{ slot: string; atomicRules: AtomicRules[] }
     </div>
   );
 };
+
+function openOriginalCode(sourceURL: string) {
+  chrome.devtools.inspectedWindow.eval<string>('window.location.origin', {}, async () => {
+    const results = sourceURL.split(':');
```
ling1726, line 98: nit: would be nice to add an example of the sourceURL in comment

ling1726 (approved):

packages/core/src/makeStyles.ts
```diff
@@ -12,6 +12,11 @@ export function makeStyles<Slots extends string | number>(stylesBySlots: StylesB
   let ltrClassNamesForSlots: Record<Slots, string> | null = null;
   let rtlClassNamesForSlots: Record<Slots, string> | null = null;
 
+  let sourceURL: string | undefined;
+  if (process.env.NODE_ENV !== 'production' && isDevToolsEnabled) {
+    sourceURL = getSourceURLfromError();
+  }
```
layershifter, line 18: Should not be the same done for `__styles`?

change/@griffel-core-1e5e67a5-96c2-4f20-9d02-feb94abc63ae.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "Support sourcemap for griffel devtools",
```
layershifter, line 3:
```suggestion
  "type": "patch",
  "comment": "feat: Add support for source maps in devtools",
```

nit: if there will be no bump - we will not get a version with needed features 🥹

layershifter (approved):

bsunderhus (approved):

packages/core/src/makeStyles.ts
YuanboXue-Amber: it is in __styles :)

PR #145, opened 2022-06-21
chore: add __css() functions for upcoming CSS extraction

layershifter:
This PR adds new `__css` functions to `@griffel/core` & `@griffel/react`.

The difference to existing `__styles` is that the part related to CSS insertion is completely stripped.

### DIFF
--- a/packages/react/bundle-size/__styles.fixture.js
+++ b/packages/react/bundle-size/__styles.fixture.js
@@ -3,5 +3,5 @@ import { __styles, mergeClasses } from '@griffel/react';
 console.log(__styles, mergeClasses);
 
 export default {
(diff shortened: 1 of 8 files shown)


packages/react/bundle-size/__styles.fixture.js
```diff
@@ -3,5 +3,5 @@ import { __styles, mergeClasses } from '@griffel/react';
 console.log(__styles, mergeClasses);
 
 export default {
-  name: 'makeStyles + mergeClasses (build time)',
+  name: '__styles + mergeClasses (build time)',
```
layershifter, line 6: Renamed fixture to be more meaningful.

PR #147, opened 2022-06-21
feat: add Webpack plugin for CSS extraction

layershifter:
This PR adds Webpack plugin to `@griffel/webpack-extraction-plugin`.

The plugin aggregates all `griffel.css` imports created by Webpack loader to a single asset via `optimization.splitChunks.cacheGroups`. Code splitting is not supported i.e. all CSS rules emitted by Griffel will go to a single chunk (that asset).

The sorting and deduplication of rules will be implemented in a follow up PR.


packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -0,0 +1,60 @@
+import { Compilation } from 'webpack';
+import type { Compiler, sources } from 'webpack';
+
+/**
+ * Forces all files with `griffel.css` be concatenated into a single asset.
+ */
+function forceCSSIntoOneStyleSheet(compiler: Compiler) {
+  compiler.options.optimization = compiler.options.optimization || {};
```
ling1726, line 8:
```suggestion
  compiler.options.optimization ??= {};
```

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -0,0 +1,60 @@
...
+  compiler.options.optimization.splitChunks = compiler.options.optimization.splitChunks || {};
```
ling1726, line 9:
```suggestion
  compiler.options.optimization.splitChunks ??= {};
```

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export class GriffelCSSExtractionPlugin {
+  static loader = require.resolve('./webpackLoader');
+
+  apply(compiler: Compiler): void {
+    forceCSSIntoOneStyleSheet(compiler);
+
+    compiler.hooks.compilation.tap('GriffelExtractPlugin', compilation => {
+      compilation.hooks.processAssets.tap(
```
ling1726, line 38: you only tap into `processAssets` after `compilation` to make sure that the build time transforms have all finished running ?

layershifter: I forgot to drop this comment... The whole part with `compiler.hooks.compilation.tap` is currently useless as the sorting/deduplication process is not implemented yet. Technically it can be deleted from the current PR, but I will need this part in the next PR anyway.

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.test.ts
```diff
@@ -0,0 +1,199 @@
...
+    let resultCSS = '';
+    let resultFsSnapshot: string[] = [];
+
+    let resultError: Error | webpack.StatsError = new Error();
+
+    try {
+      const result = await compileSourceWithWebpack(inputPath, options);
+
+      resultCSS = fixLineEndings(
+        prettier.format(result.cssOutput, {
```
ling1726, line 167: is `result.cssOuput` set by the mini-css-extract plugin ?

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
ling1726: why would you need to tap into the compilation hook ? the sorting/deduplication will still happen on asset processing level

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.test.ts
layershifter: Kinda, `mini-css-extract` plugin extracts CSS to `griffel.css`, `result.cssOutput` is a content of that file.

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
layershifter: Because I need to have an access to `compilation` to replace assets?

```
   compiler.hooks.compilation.tap('GriffelExtractPlugin', compilation => {
      compilation.hooks.processAssets.tap(
        {
          name: 'GriffelExtractPlugin',
        },
        assets => {
```

- [compilation](https://webpack.js.org/api/compiler-hooks/#compilation) is a _compiler_ hook
- [processAssets](https://webpack.js.org/api/compilation-hooks/#processassets) is a _compilation_ hook

ling1726: aaaah ok, I missed that the compilation object is provided by the compilation hook

ling1726 (approved):

PR #149, opened 2022-06-27
chore: prepare @griffel/webpack-extraction-plugin for publish

layershifter:
This PR:
- adds proper `README` file to `@griffel/webpack-extraction-plugin`
- removes `private: true` from its `package.json` to publish the package

### DIFF
--- a/packages/webpack-extraction-plugin/README.md
+++ b/packages/webpack-extraction-plugin/README.md
@@ -58,7 +58,8 @@ module.exports = {
           loader: '@griffel/webpack-loader',
         },
       },
-      // ".css" configured there for example, but "css-loader" is required to handle produces CSS assets
+      // "css-loader" is required to handle produced CSS assets by Griffel
+      // you can use "style-loader" or "MiniCssExtractPlugin.loader" to handle CSS insertion
       {
         test: /\.css$/,
         use: [MiniCssExtractPlugin.loader, 'css-loader'],
--- a/packages/webpack-extraction-plugin/README.md
+++ b/packages/webpack-extraction-plugin/README.md
@@ -27,6 +27,7 @@ _Currently, all CSS rules will be extracted to a single CSS file i.e. [code spli
 ## Usage
 
 Webpack documentation:
+
 - [Loaders](https://webpack.js.org/loaders/)
 - [Plugins](https://webpack.js.org/concepts/plugins/)
 
--- a/packages/webpack-extraction-plugin/README.md
+++ b/packages/webpack-extraction-plugin/README.md
@@ -70,7 +70,7 @@ module.exports = {
 };
 ```
 
-- Usage `mini-css-extract-plugin` is not required
+- `mini-css-extract-plugin` is not mandatory and configured there for example, you can use `style-loader` or other tooling to inject CSS on a page
 
 For better performance (to process less files) consider to use `include` for `GriffelCSSExtractionPlugin.loader`:
 
--- a/packages/webpack-extraction-plugin/README.md
+++ b/packages/webpack-extraction-plugin/README.md
@@ -1,3 +1,71 @@
-# TBD
+# Webpack plugin to perform CSS extraction in Griffel
 
-Not ready for consuming.
+> ⚠️ **Note** This package is experimental, API may change.
+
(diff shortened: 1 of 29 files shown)


packages/webpack-extraction-plugin/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@griffel/webpack-extraction-plugin",
   "version": "0.0.1",
-  "private": true,
```
layershifter, line 4: The plan is to publish it as `0.0.x` as I don't have enough confidence in it.

packages/webpack-extraction-plugin/README.md
```diff
@@ -1,3 +1,71 @@
...
+> ⚠️ **Note** This package is experimental, API may change.
+
+A plugin for Webpack 5 that performs CSS extraction for [`@griffel/react`](../react).
+
+## Install
+
+```bash
+yarn add --dev @griffel/webpack-extraction-plugin
+# or
+npm install --dev @griffel/webpack-extraction-plugin
```
Dwlad90, line 12: Hi, I guess that is the right way to install the devDependencies: 

```suggestion
npm install --save-dev @griffel/webpack-extraction-plugin
```

packages/webpack-extraction-plugin/README.md
```diff
@@ -1,3 +1,71 @@
...
+```js
+const { GriffelCSSExtractionPlugin } = require('@griffel/webpack-extraction-plugin');
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+
+module.exports = {
+  module: {
+    rules: [
+      {
+        test: /\.(js|ts|tsx)$/,
+        // Apply "exclude" only if your dependencies **do not use** Griffel
```
ling1726, line 46: should `include` be added to the docs too?

packages/webpack-extraction-plugin/README.md
```diff
@@ -1,3 +1,71 @@
...
+      },
+      // Add "@griffel/webpack-loader" if you use Griffel directly in your project
+      {
+        test: /\.(ts|tsx)$/,
+        exclude: /node_modules/,
+        use: {
+          loader: '@griffel/webpack-loader',
+        },
+      },
+      // ".css" configured there for example, but "css-loader" is required to handle produces CSS assets
```
ling1726, line 60: >.css configured there for example

what does this mean ?

packages/webpack-extraction-plugin/README.md
```diff
@@ -1,3 +1,71 @@
...
+        test: /\.css$/,
+        use: [MiniCssExtractPlugin.loader, 'css-loader'],
+      },
+    ],
+  },
+  plugins: [new MiniCssExtractPlugin(), new GriffelCSSExtractionPlugin()],
+};
+```
+
+- Usage `mini-css-extract-plugin` is not required
```
ling1726, line 71:
```suggestion
- Usage `mini-css-extract-plugin` is not required
```

Can you clarify this point? Do you mean that the css will simply be in the main js bundle if mini-css-extract-plugin is not used ?

packages/webpack-extraction-plugin/README.md
layershifter: Added a section below 64a52cd

packages/webpack-extraction-plugin/README.md
layershifter:
```suggestion
      // "css-loader" is required to handle produced CSS assets by Griffel
      // you can use "style-loader" or "MiniCssExtractPlugin.loader" to handle CSS insertion
```

Is it better?

packages/webpack-extraction-plugin/README.md
layershifter:
```suggestion
- `mini-css-extract-plugin` is not mandatory and configured there for example, you can use `style-loader` or other tooling to inject CSS on a page
```

Is it better?

ling1726: yup that's much better

packages/webpack-extraction-plugin/README.md
ling1726: yup looks good

ling1726 (approved):

PR #168, opened 2022-07-19
Make @griffel/devtools public

YuanboXue-Amber:
Make @griffel/devtools a public package


change/@griffel-devtools-167de6fc-ff1a-4cec-a1c9-12ad30b92981.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
```
layershifter, line 2:
```suggestion
  "type": "minor",
```

Let's bump it to have a proper version

YuanboXue-Amber: Updated to patch after chat :)

layershifter (approved):

Issue #162, opened 2022-07-12
babel-preset/webpack-loader: Build time transforms do not handle assets properly


layershifter:
### Repro steps

- Open sandbox, https://stackblitz.com/edit/sveltejs-kit-template-default-urdlga?file=src%2Findex.js (download locally if it works better)
- `yarn`
- `webpack`

### The problem

- Check output in `dist/main.*.js`
  ![image](https://user-images.githubusercontent.com/14183168/178512945-2c6fd249-0398-4257-91f9-964a58666788.png)
  - Runtime version (i.e. without transforms) uses proper file
  - Build time version (i.e. after transforms) references source file

The problem is in module evaluation and missing special handling around it to keep imports in place.

PR #170, opened 2022-07-21
fix: handle assets in Babel plugin

layershifter:
This PR fixes partially #162. It update Babel plugin to postprocess assets:

```mermaid
graph TD
    A[Evaluate styles to get an object with styles] --> B
    B[Preprocess styles to normalize paths *] --> C
    C[Resolve styles to classes mapping and CSS rules] --> D[Parse CSS rules to extract assets to imports *]
```

_New steps are marked with \*_.

## Preprocess styles to normalize paths

```ts
makeStyles({
  root: { backgroundImage: "url(../../assets/image.jpg)" }
});
// ⬇️⬇️⬇️
makeStyles({
    backgroundImage: 'url(assets/image.jpg)',
});
```

Linaria v3 outputs relative paths, but v4 will resolve to absolute:

- v3: `../image.png`
- v4: `/home/user/project/assets/image.png`

On this step paths will be normalized to use relative paths from `projectRoot` (new option for Babel preset) so we will get the same classes on if builds will run on different machines or the same asset will be imported from different files.

We cannot emit imports on this step as rules should be processed with `resolveStyleRulesForSlots()` first to generate classes.

## Parse CSS rules to extract assets to imports

On this stage we have updated AST and just extract assets from CSS rules (see `replaceAssetsWithImports()`).

---

### Before

```ts
import { __styles } from "@griffel/react";
import image from "./blank.jpg";
  {
    root: {
      Bcmaq0h: "fnwsaxv"
    }
  },
  {
    // ❌ An asset is a string so a bundler will ignore it
    d: [".fnwsaxv{background-image:url(/blank.jpg"]
  }
);
```

### After

```ts
import _asset from "./blank.jpg";
import { __styles } from "@griffel/react";
import image from "./blank.jpg";
  {
    root: {
      Bcmaq0h: "fnwsaxv"
    }
  },
  {
    // ✅ An asset is an import so a bundler will handle it
    d: [`.fnwsaxv{background-image:url(${_asset});}`]
  }
);
```


packages/babel-preset/tsconfig.spec.json
```diff
@@ -4,7 +4,7 @@
     "jsx": "react",
     "outDir": "../../dist/out-tsc",
     "module": "commonjs",
-    "types": ["jest", "node", "environment"]
+    "types": ["jest", "node", "environment", "static-assets"]
```
layershifter, line 7: To handle imports of `.jpg` files, needed only for fixtures.

packages/babel-preset/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "allowJs": true,
     "forceConsistentCasingInFileNames": true,
+    "lib": ["ES2019", "dom", "DOM.Iterable"],
```
layershifter, line 6: `Object.fromEntries()` are needed...

packages/babel-preset/__fixtures__/assets/output.ts
```diff
@@ -0,0 +1,13 @@
...
+import { __styles } from '@griffel/react';
+import image from './blank.jpg';
+export const useStyles = __styles(
+  {
+    root: {
+      Bcmaq0h: 'fnwsaxv',
+    },
+  },
+  {
+    d: [`.fnwsaxv{background-image:url(${_asset});}`],
```
ling1726, line 11: I would also test the case where there are multiple assets, either a separate test or add more assets to this fixture

packages/babel-preset/src/transformPlugin.ts
```diff
@@ -120,6 +122,7 @@ export const transformPlugin = declare<Partial<BabelPluginOptions>, PluginObj<Ba
         action: 'ignore',
       },
     ],
+    projectRoot: process.cwd(),
```
ling1726, line 125: is it safe to use cwd ? why not find the closest package.json ?

packages/babel-preset/__fixtures__/assets/output.ts
```diff
@@ -0,0 +1,13 @@
+import _asset from './blank.jpg';
+import { __styles } from '@griffel/react';
+import image from './blank.jpg';
```
ling1726, line 3: I guess the bundle will just strip the unused asset import ?

layershifter: Yup, correct.

packages/babel-preset/src/transformPlugin.ts
layershifter: Closest `package.json` is not really good option for monorepos as the same asset can be used in different packages:

```
/packages
/packages/assets
/packages/src-a # both "src-a" & "src-b" use a file from "assets", but relative path
/packages/src-b # will be different for "src-a" & "src-b" as both contain "package.json" files
```

packages/babel-preset/src/transformPlugin.ts
```diff
@@ -166,10 +180,23 @@ export const transformPlugin = declare<Partial<BabelPluginOptions>, PluginObj<Ba
...
+                normalizeStyleRules(
+                  pluginOptions.projectRoot,
+                  // Presence of "state.filename" is validated on `Program.enter()`
+                  state.filename as string,
+                  stylesBySlots,
+                ),
+              );
+              const uniqueCSSRules = dedupeCSSRules(cssRulesByBucket);
+
+              (callExpressionPath.get('arguments.0') as NodePath).remove();
```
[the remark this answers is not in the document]
layershifter: Assignment to a node does not create new references (paths) which will cause problems if a tree will be traversed. Best practices are described in [plugin-handbook.md#toc-manipulation](https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#toc-manipulation). However, unfortunately there is no something like`replaceInContainer()` that would be more accurate in this case.

packages/babel-preset/__fixtures__/assets/output.ts
layershifter: There is a different test that is broken with Linaria v3, it includes multiple nested assets 🙃

I will modify this fixture to include more assets.

layershifter: Updated 5b6db17

ling1726 (approved):

PR #171, opened 2022-07-21
fix: handle assets in CSS extraction

### LINKED ISSUE #162: babel-preset/webpack-loader: Build time transforms do not handle assets properly

layershifter:
Fixes #162.

This PR fixes assets handling for CSS extraction. Compared to the previous PR it does reverse transforms:

```ts
import _asset from './blank.jpg';

  {
    root: {
      Bcmaq0h: 'fnwsaxv',
    },
  },
  {
    d: [`.fnwsaxv{background-image:url(${_asset});}`],
  },
);
// ⬇️⬇️⬇️
import { __styles, __css } from '@griffel/react';

export const useStyles = __css({
  root: {
    Bcmaq0h: 'fnwsaxv',
  },
  // Rules are remove by the plugin, but it's a result of added code: it inlines imports
  // {
  //   d: [`.fnwsaxv{background-image:url(${'../__fixtures__/webpack/assets/blank.jpg'});}`],
  // },
});
```

All work is done in `babelPluginStripGriffelRuntime`:
- removes imports for assets used in `__styles` calls
- inlines imports used in `__styles` calls

Imports are inlined with relative path to `griffel.css` file so `css-loader` could handle them properly and extract assets.


packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.test.ts
```diff
@@ -34,6 +34,7 @@ async function compileSourceWithWebpack(
       path: path.resolve(__dirname),
       filename: 'bundle.js',
       pathinfo: false,
+      assetModuleFilename: '[name][ext]',
```
layershifter, line 37: Configure explicitly to have original filenames.

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.test.ts
```diff
@@ -16,9 +16,7 @@ pluginTester({
     },
   },
   pluginOptions: {
-    babelOptions: {
```
layershifter, line 19: Old options were copy-paste issue.

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.test.ts
```diff
@@ -16,9 +16,7 @@ pluginTester({
...
-      presets: ['@babel/typescript'],
-    },
+    resourceDirectory: process.cwd(),
```
layershifter, line 19: This value is just for testing, Webpack loader will pass the real one.

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
```diff
@@ -79,6 +115,59 @@ export const babelPluginStripGriffelRuntime = declare<
...
+                      // When imports are inlined, we need to adjust the relative paths inside url(..) expressions
+                      // to allow css-loader resolve an imported asset properly
+                      transformUrl(
+                        state.filename!,
+                        options.resourceDirectory!,
+                        importDeclarationPath.get('source').node.value,
+                      ),
+                    ),
+                  );
+                  importDeclarationPath.remove();
```
ling1726, line 167: why remove the import decalartion ? wouldn't it also get stripped by bundlers ?

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.test.ts
ling1726: will end users need to have this in their webpack configs ?

ling1726 (approved):
A few small questions but otherwise it's quite clear that you're using babel to find the import path of the inline template literal asset and just doing a replacement

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.test.ts
layershifter: Nope, it's just to have the same names for assets in `fs.json` files uses in tests.

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
layershifter: Technically these is no reason, but as these imports are used only for a specific declaration I decided to remove it to be a bit nicer 🐱

PR #172, opened 2022-07-21
fix: add css rules to devtools extension debugger when pass renderer on SSR

Dwlad90:
This PR fixes https://github.com/microsoft/griffel/issues/173.

### DIFF
--- a/packages/core/src/devtools/store.ts
+++ b/packages/core/src/devtools/store.ts
@@ -41,4 +41,7 @@ export const debugData = {
   getSequenceDetails: (sequenceHash: SequenceHash): typeof sequenceDetails[string] | undefined => {
     return sequenceDetails[sequenceHash];
   },
+  hasCSSRule: (rule: string): boolean => {
+    return cssRules.includes(rule);
+  },
 };
--- a/packages/core/src/devtools/store.ts
+++ b/packages/core/src/devtools/store.ts
@@ -4,7 +4,7 @@ import { mergeClassesCachedResults } from '../mergeClasses';
 
 const sequenceDetails: Record<SequenceHash, { slotName: string; sourceURL?: string }> = {};
 
-const cssRules: string[] = [];
+const cssRules = new Set<string>();
 
 export const debugData = {
   getChildrenSequences: (debugSequenceHash: SequenceHash): SequenceHash[] => {
@@ -24,7 +24,7 @@ export const debugData = {
   },
 
   addCSSRule: (rule: string) => {
-    cssRules.push(rule);
+    cssRules.add(rule);
   },
   addSequenceDetails: <Slots extends string | number>(
     classNamesForSlots: Record<Slots, string>,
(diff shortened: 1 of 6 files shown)


packages/core/src/renderer/createDOMRenderer.ts
```diff
@@ -67,14 +67,17 @@ export function createDOMRenderer(
             metadata,
           );
 
+          if (process.env.NODE_ENV !== 'production' && isDevToolsEnabled) {
+            if (!debugData.hasCSSRule(ruleCSS)) {
+              debugData.addCSSRule(ruleCSS);
+            }
+          }
+
```
layershifter, line 75: As it's SSR specific I was wondering if we could fill data in `rehydrateRendererCache()` instead:

https://github.com/microsoft/griffel/blob/79c97b9ec6b4b028e31e17b613e6d89a37bfd6a7/packages/core/src/renderer/rehydrateRendererCache.ts#L41-L45

Will it work?

packages/core/src/devtools/store.ts
```diff
@@ -41,4 +41,7 @@ export const debugData = {
   getSequenceDetails: (sequenceHash: SequenceHash): typeof sequenceDetails[string] | undefined => {
     return sequenceDetails[sequenceHash];
   },
+  hasCSSRule: (rule: string): boolean => {
+    return cssRules.includes(rule);
+  },
```
layershifter, line 46: I am thinking to turn `cssRules` into `Set` if it's really needed:

```diff
-const cssRules: string[] = [];
+const cssRules = new Set<string>();
```

```diff
  addCSSRule: (rule: string) => {
-    cssRules.push(rule);
+    cssRules.add(rule);
  },
```

```diff
  getCSSRules: (): string[] => {
-   return cssRules;
+   return Array.from(cssRules);
  },
```

layershifter:
FYI @YuanboXue-Amber

layershifter:
@Dwlad90 thanks for contributing, I left few comments 🙂

packages/devtools/src/utils.ts
```diff
@@ -10,12 +10,14 @@ export function getRulesBySlots(node: DebugResult, result: SlotInfo[] = []): Slo
...
         sourceURL,
-        rules: debugClassNames.map(({ className, overriddenBy }) => {
-          return {
-            cssRule: rules![className],
-            overriddenBy,
-          };
-        }),
+        rules: debugClassNames
+          .filter(({ className }) => !!rules?.[className])
+          .map(({ className, overriddenBy }) => {
```
Dwlad90, line 15: Use reduce instead of filter and map

packages/core/src/devtools/store.ts
Dwlad90: Yes, this will solve the duplication problems and eliminate the need to check if the string exists.

Dwlad90:
> 

@layershifter Thanks for a code review, I've update the PR according to your suggestions

YuanboXue-Amber (approved):

layershifter (approved):

PR #174, opened 2022-07-27
Fix(RTL): Flip CSS properties even if they contain CSS variables

miroslavstastny:
**What**
Allow RTL flipping of CSS properties even if they reference CSS variables.

```js
boxShadow: '6px 3px 8px 5px var(--css-variable)';

/* ⬇️ now flips offset-x in RTL ⬇️ */

boxShadow: '-6px 3px 8px 5px var(--css-variable)';
```

**Why**

Previously, CSS properties which contained `var()` were not flipped.

There are cases where, when a CSS variable is used, there is no way how the value can be flipped:

```js
padding: 'var(--foo)'; /* 🔴 No way how to flip - foo can be a single value as well as four? */
```

On the other hand, there are other use cases, where the flipping is possible and meaningful:
```js
padding: margin: '1px var(--token) 3px 4px'; /* ✅ should be flipped to '1px 4px 3px var(--token)' */
boxShadow: '6px 3px 8px 5px var(--css-variable);' /* ✅ CSS variable used only for color, offset can be safely flipped */
```

**Caveats**
If CSS variable is used for the first value in `shadow`, the flipped result might be wrong:
```js
boxShadow: 'var(--token) 3px 4px red' /* 💥 3px is considered the first number and is flipped to -3px */
```
This is the expected result if the `--token` resolves to `inset` but incorrect if it resolves to a number. In that case the `3px` is `offset-y` which should not have been flipped.

Takeaway: flipping values which contain CSS variables is problematic, styles authors must know what they are doing. As there are still valid and safe use cases, this behavior is better than just strictly disallowing anything what contains a CSS variable to be flipped.

### DIFF
--- a/change/@griffel-core-c60d0c4f-e496-4ec4-bb67-fae7384bfbe1.json
+++ b/change/@griffel-core-c60d0c4f-e496-4ec4-bb67-fae7384bfbe1.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix(RTL): Flip CSS properties even if they contain CSS variables",
+  "packageName": "@griffel/core",
+  "email": "miroslav.stastny@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@griffel-core-c60d0c4f-e496-4ec4-bb67-fae7384bfbe1.json
+++ b/change/@griffel-core-c60d0c4f-e496-4ec4-bb67-fae7384bfbe1.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "fix(RTL): Flip CSS properties even if they contain CSS variables",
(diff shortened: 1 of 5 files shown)


bsunderhus (approved):

change/@griffel-core-c60d0c4f-e496-4ec4-bb67-fae7384bfbe1.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix(RTL): Flip CSS properties even if they contain CSS variables",
```
layershifter, line 3:
```suggestion
  "comment": "fix: flip CSS properties even if they contain CSS variables",
```

_nit_

layershifter (approved):

PR #175, opened 2022-07-27
feat(devtools): open original source in source tab

YuanboXue-Amber:
PR #142 introduced the ability to open makeStyles call in source tab when clicking on '</>' button in section title. However it opens the compiled source. This PR enhance it to open real source. It should work with any build configuration (hopefully 🤞).

Before (can only open compiled source): 
before

Now (open real source): 
after


.yarn/patches/source-map-js-npm-1.0.2-ee4f9f9b30.patch
```diff
@@ -0,0 +1,63 @@
+diff --git a/lib/quick-sort.js b/lib/quick-sort.js
```
YuanboXue-Amber, line 1: This file is patched because the original file does evaluation (L41) and it is not allowed in extension.

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
```diff
@@ -0,0 +1,178 @@
+import type { MappedPosition, RawSourceMap } from 'source-map-js';
+import { getFilePath, getOriginalPosition, resources } from './sourceMapConsumer';
+
+// when only one source map appears in a source file: source url -> source map JSON
+// when more than one source maps appear in a source file: `${source}:${line}:${column}` -> source map JSON
+const sourceMapJSONs: Map<string, RawSourceMap> = new Map();
+
+const getComputedSouceMapJSON = ({ source, line, column }: MappedPosition) => {
```
ling1726, line 8:
```suggestion
const getComputedSourceMapJSON = ({ source, line, column }: MappedPosition) => {
```

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
```diff
@@ -0,0 +1,178 @@
...
+};
+
+export async function loadOriginalSourceLoc(sourceUrlWithLoc: string): Promise<MappedPosition> {
+  const paths = sourceUrlWithLoc.split(':');
+  const column = Number(paths.pop());
+  const line = Number(paths.pop());
+  const source = paths.join(':');
+  if (Number.isNaN(line) || Number.isNaN(column)) {
+    return { source, line: 1, column: 0 };
+  }
```
ling1726, line 19:
```suggestion
  const paths = sourceUrlWithLoc.split(':');
  const column = Number(paths.pop());
  const line = Number(paths.pop());
  const source = paths.join(':');
  if (Number.isNaN(line) || Number.isNaN(column)) {
    return { source, line: 1, column: 0 };
  }
```

nit: you could extract this to a `parseSoureUrl` function

```ts
function parseSourceUrl(sourceUrlWithLoc: string): MappedPosition {
  const paths = sourceUrlWithLoc.split(':');
  const column = Number(paths.pop());
  const line = Number(paths.pop());
  const source = paths.join(':');
  if (Number.isNaN(line) || Number.isNaN(column)) {
    return { source, line: 1, column: 0 };
  }

  return { source, line, column };
}
```

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
```diff
@@ -0,0 +1,178 @@
...
+        const sourceMapJSON = JSON.parse(decoded);
+
+        // TODO optionally turn on debug message
+        // if (__DEBUG__) {
+        //   console.groupCollapsed('[Griffel devtools] extractAndLoadSourceMapJSON() Inline source map');
+        //   console.log(sourceMapJSON);
+        //   console.groupEnd();
+        // }
+
+        if (sourceMapIncludesSource(sourceMapJSON, getFilePath(source))) {
```
ling1726, line 71: This code to insert into the map seems super weird with the usage of `sourceMapNum`, couldn't you just collect all entries in the loop and insert at the end ? it would simplify this code a lot

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
```diff
@@ -0,0 +1,178 @@
...
+};
+function sourceMapIncludesSource(sourcemap: RawSourceMap | IndexSourceMap, sourcePath: string): boolean {
+  if (sourcemap.mappings === undefined) {
+    return (sourcemap as IndexSourceMap).sections.some(section => {
+      return sourceMapIncludesSource(section.map, sourcePath);
+    });
+  }
+
+  return (sourcemap as RawSourceMap).sources.some(s => s === 'Inline Babel script' || s.includes(sourcePath));
+}
```
ling1726, line 178:
```suggestion
function sourceMapIncludesSource(sourcemap: RawSourceMap | IndexSourceMap, sourcePath: string): boolean {
  const sourcemaps: (IndexSourceMap | RawSourceMap)[] = [sourcemap];
  while(sourcemaps.length) {
    const curSourcemap = sourcemaps.pop()!;
    if (isRawSourceMap(curSourcemap)) {
      return curSourcemap.sources.some(s => s === 'Inline Babel script' || s.includes(sourcePath));
    }

    sourcemaps.push(...curSourcemap.sections.map(section => section.map));
  }

  return false;
}

function isRawSourceMap(sourcemap: RawSourceMap | IndexSourceMap): sourcemap is RawSourceMap {
  return !!sourcemap.mappings;
}
```
is perf an issue ? I see you've patched quick-sort, so if it is the iterative dive should be cheaper in theory. Also you can avoid the as casing with a simple type predicate

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
```diff
@@ -0,0 +1,178 @@
...
+      return response.text();
+    }
+    throw new Error(`[Griffel devtools] fetchRuntimeSource() bad response fetching ${url}: ${response.status}`);
+  } catch (error) {
+    // try to get url content from chrome resources api
+    const resource = (await resources).find(resource => resource.url === url);
+    if (resource) {
+      return new Promise(resolve => resource.getContent(content => resolve(content)));
+    }
+    throw new Error(`[Griffel devtools] fetchFiles() unable to fetch ${url}: ${error}`);
```
ling1726, line 142:
```suggestion
```

You never need this error right ? you just want to rethrow the earlier error. If the chrome resource api fetch fails, it will throw a compeltely different error

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
```diff
@@ -0,0 +1,178 @@
...
+        // if (__DEBUG__) {
+        //   console.groupCollapsed('[Griffel devtools] extractAndLoadSourceMapJSON() Inline source map');
+        //   console.log(sourceMapJSON);
+        //   console.groupEnd();
+        // }
+
+        if (sourceMapIncludesSource(sourceMapJSON, getFilePath(source))) {
+          // store computed sourceMapJSON
+          if (sourceMapNum > 1) {
+            sourceMapJSONs.set(`${source}:${line}:${column}`, sourceMapJSON);
```
ling1726, line 74: I'm interested to know why can't all sourcemaps be stored with line and column number ? what's the difference vs just source ?

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
```diff
@@ -0,0 +1,178 @@
...
+      }
+    } else {
+      externalSourceMapURLs.push(sourceMappingURL);
+    }
+
+    sourceMappingURLMatch = sourceMapRegex.exec(runtimeSourceCode);
+  }
+
+  // No inline sourcemap found. Check if there's external source map
+  let sourceMappingURL = externalSourceMapURLs.pop();
```
ling1726, line 98: why only use the latest external source map and skip all others ?

.yarn/patches/source-map-js-npm-1.0.2-ee4f9f9b30.patch
```diff
@@ -0,0 +1,63 @@
...
+index 23f9eda57f7383b5291361e4593cf2fdef5409ba..016785171b4d69d756ac795763abc6cf2c9d1b31 100644
+--- a/lib/quick-sort.js
```
ling1726, line 3: why is this patch necessary ? is it for performance reasons that you don't to clone and cache sort instances ?

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
```diff
@@ -0,0 +1,178 @@
...
+    const resource = (await resources).find(resource => resource.url === url);
+    if (resource) {
+      return new Promise(resolve => resource.getContent(content => resolve(content)));
+    }
+    throw new Error(`[Griffel devtools] fetchFiles() unable to fetch ${url}: ${error}`);
+  }
+}
+
+function decodeBase64String(encoded: string): string {
+  if (typeof atob === 'function') {
```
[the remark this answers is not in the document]
YuanboXue-Amber: I think this is not a problem either. I copied this part directly from react devtools https://github.com/facebook/react/blob/b66936ece7d9ad41a33e077933c9af0bda8bff87/packages/react-devtools-shared/src/hooks/parseHookNames/loadSourceAndMetadata.js#L121

Now to think about it, maybe it's because code in react devtools is not only running on the client side. I'm going to simplify this and keep only `atob`

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
YuanboXue-Amber: Perf is not an issue. I patched quick-sort because it uses `eval` and it is banned from devtools. I would prefer the recursive version because it's more readable IMO.

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
YuanboXue-Amber: Hmm I didn't get it? I think I need it: if fetch fails, it will be catched, and then I will re-throw it here

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
YuanboXue-Amber: When there's one source map per source, only using `source` url as key is good enough.
When there's multiple source map per source, one `source` url will be related with multiple source maps. In this case I use `${source}:${line}:${column}` as key

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
ling1726: I made a type in the original suggestion, you should just rethrow the caught error:

If original fetch throws an error and you try to use `resource.getContent`

If `resource.getContent` throws then a different error is thrown.

So there is no reason to rethrow a new error from the `catch` block

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
YuanboXue-Amber: This is also stoled from react-devtools https://github.com/facebook/react/blob/b66936ece7d9ad41a33e077933c9af0bda8bff87/packages/react-devtools-shared/src/hooks/parseHookNames/loadSourceAndMetadata.js#L239

They have a comment there
```tsx
// Files with external source maps should only have a single source map.
// More than one result might indicate an edge case,
// like a string in the source code that matched our "sourceMappingURL" regex.
// We should just skip over cases like this.
```
I will added the same comment too

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
ling1726: In that case I would suggest you still use the type predicate to avoid `as` casting

.yarn/patches/source-map-js-npm-1.0.2-ee4f9f9b30.patch
YuanboXue-Amber: The original code uses evaluation (in L41 ```let templateFn = new Function(`return ${template}`)();```) and it is not allowed in devtools :( So I found the related [commit](https://github.com/7rulnik/source-map-js/pull/2/files) in source-map-js and revert this commit via a patch.

Actually the commit is about perf improvement, so according to that the patched version is less performant. But it's better than having nothing.... I'm really surprised how few libraries are doing source map parsing.

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
YuanboXue-Amber: Added in L113

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
YuanboXue-Amber: Oh gotcha!

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
YuanboXue-Amber: Updated with `isRawSourceMap` instead of using `as`, it looks much better IMO :)

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
ling1726: OK but is there any difference in usage ? this code can be simplified a lot if you stored all sourcemaps with the `source:line:column` format

.yarn/patches/source-map-js-npm-1.0.2-ee4f9f9b30.patch
ling1726: does react-devtools use this library ?

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
YuanboXue-Amber: Yeah there is a difference:
Consider a file `http://myfile` which contains only one source map.

When the input is `http://myfile:1:2`, we do: read file -> find source map -> parse source map and get JSON-> locate line1 col2 in source map JSON -> done.

And the next time input could be `http://myfile:10:20` (another makeStyles hook in same file). If we store using `source:line:column`, we cannot reuse the source map JSON we get from the previous step. But if we store using just `source` as key, we can get source map JSON immediately.

.yarn/patches/source-map-js-npm-1.0.2-ee4f9f9b30.patch
YuanboXue-Amber: It does not, it parses sourcemap on its own 🙈

YuanboXue-Amber: here it is: https://github.com/facebook/react/blob/b66936ece7d9ad41a33e077933c9af0bda8bff87/packages/react-devtools-shared/src/hooks/SourceMapConsumer.js 
It's not very long but I would prefer the library

ling1726 (approved):

.yarn/patches/source-map-js-npm-1.0.2-ee4f9f9b30.patch
ling1726: I would trigger an issue to the library, see if they can release a version where your alternative could exist

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
ling1726: Ah got it thx

.yarn/patches/source-map-js-npm-1.0.2-ee4f9f9b30.patch
layershifter, line 1: To clarify, this patch will work as we bundling the extension, righj?

layershifter (approved):

.yarn/patches/source-map-js-npm-1.0.2-ee4f9f9b30.patch
YuanboXue-Amber: Yes :)

.yarn/patches/source-map-js-npm-1.0.2-ee4f9f9b30.patch
YuanboXue-Amber: https://github.com/7rulnik/source-map-js/issues/15

PR #180, opened 2022-08-02
chore: bump docusaurus to 2.0.1

miroslavstastny:
Final version of Docusaurus has been released -> bump the dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,9 +27,9 @@
   "devDependencies": {
     "@babel/preset-typescript": "7.12.13",
     "@codesandbox/sandpack-react": "0.13.15",
-    "@docusaurus/core": "2.0.0-rc.1",
-    "@docusaurus/module-type-aliases": "^2.0.0-rc.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -27,9 +27,9 @@
   "devDependencies": {
     "@babel/preset-typescript": "7.12.13",
     "@codesandbox/sandpack-react": "0.13.15",
-    "@docusaurus/core": "2.0.0-rc.1",
-    "@docusaurus/module-type-aliases": "^2.0.0-rc.1",
-    "@docusaurus/preset-classic": "2.0.0-rc.1",
+    "@docusaurus/core": "2.0.1",
+    "@docusaurus/module-type-aliases": "2.0.1",
```
miroslavstastny, line 31: What was the reason for this package to use caret dep? Should that stay that way?

bsunderhus (approved):

package.json
layershifter: No reason, use exact versions

PR #186, opened 2022-08-03
fix: assets handling in Windows env

layershifter:


packages/webpack-extraction-plugin/src/webpackLoader.test.ts
```diff
@@ -25,6 +25,9 @@ async function compileSourceWithWebpack(entryPath: string, options: CompileOptio
       filename: 'bundle.js',
       pathinfo: false,
     },
+    externals: {
+      '@griffel/react': 'Griffel',
+    },
```
layershifter, line 30: Using `externals` is better:
- we don't need Griffel itself for these tests
- with `externals` we don't depend on build artifacts from other packages

packages/babel-preset/src/assets/replaceAssetsWithImports.ts
```diff
@@ -1,21 +1,9 @@
...
-  const absoluteAssetPath = path.resolve(projectRoot, assetPath);
-  const assetDirectory = path.dirname(absoluteAssetPath);
-
-  if (fileDirectory === assetDirectory) {
-    return './' + path.basename(assetPath);
-  }
-
-  return path.relative(fileDirectory, absoluteAssetPath);
-}
-
```
layershifter, line 18: Went to a separate file.

christiango (approved):

miroslavstastny (approved):

PR #187, opened 2022-08-03
feat: re-export GriffelRenderer from @griffel/react

layershifter:
`GriffelRenderer` type is needed to defined proper custom types to be used in Next.js (see https://github.com/microsoft/fluentui/issues/24158, https://github.com/microsoft/fluentui/pull/24202).


packages/react/src/index.ts
```diff
@@ -1,4 +1,4 @@
-export { shorthands, mergeClasses, createDOMRenderer } from '@griffel/core';
+export { shorthands, mergeClasses, createDOMRenderer, GriffelRenderer } from '@griffel/core';
```
miroslavstastny, line 1: As there is a separate `export type` on the next line, shouldn't this go there?

layershifter: Oops, good catch 😨

miroslavstastny (approved):

PR #194, opened 2022-08-12
docs: update READMEs for next-extraction-plugin & webpack-loader

layershifter:
This PR updates README files based on feedback from #190.

### DIFF
--- a/packages/next-extraction-plugin/README.md
+++ b/packages/next-extraction-plugin/README.md
@@ -28,9 +28,9 @@ In `next.config.js` file you'll need to add the next-plugin from `@griffel/webpa
 // next.config.js
 const { withGriffelCSSExtraction } = require('@griffel/next-extraction-plugin');
 
-module.exports = withGriffelCssExtraction()({
+module.exports = withGriffelCSSExtraction()({
   webpack(config) {
-    config.module.rules.push({
+    config.module.rules.unshift({
       test: /\.(js|jsx)$/,
       exclude: /node_modules/,
       use: [
@@ -41,12 +41,17 @@ module.exports = withGriffelCssExtraction()({
     });
 
     // If your project uses TypeScript
-    config.module.rules.push({
+    config.module.rules.unshift({
       test: /\.(ts|tsx)$/,
       exclude: /node_modules/,
       use: [
         {
           loader: '@griffel/webpack-loader',
+          options: {
+            babelOptions: {
+              presets: ['next/babel'],
+            },
+          },
         },
       ],
     });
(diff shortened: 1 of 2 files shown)


packages/next-extraction-plugin/README.md
```diff
@@ -11,24 +10,47 @@ yarn add --dev @griffel/next-extraction-plugin
...
+      ],
+    });
+
+    // If your project uses TypeScript
+    config.module.rules.push({
+      test: /\.(ts|tsx)$/,
       exclude: /node_modules/,
       use: [
         {
           loader: '@griffel/webpack-loader',
```
yume-chan, line 49: Also need `babelOptions` here.

Note that in a typical Next app, `@babel/preset-typescript` needs to be installed manually.

Maybe better to use `next/babel` preset which comes with Next, see https://nextjs.org/docs/advanced-features/customizing-babel-config

yume-chan (changes requested):
I have tested my proposed changes with #195 patched locally. Both manually install `@babel/preset-typescript` and use `next/babel` directly works.

packages/next-extraction-plugin/README.md
layershifter: Good call, added 👍

packages/next-extraction-plugin/README.md
```diff
@@ -11,24 +10,47 @@ yarn add --dev @griffel/next-extraction-plugin
...
 ## Usage
 
 In `next.config.js` file you'll need to add the next-plugin from `@griffel/webpack-extraction-plugin` like so:
+
 ```js
 // next.config.js
-const { withGriffelCSSExtraction } = require('@griffel/webpack-extraction-plugin');
+const { withGriffelCSSExtraction } = require('@griffel/next-extraction-plugin');
 
 module.exports = withGriffelCssExtraction()({
```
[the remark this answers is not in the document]
layershifter: Oops, thanks 👍

yume-chan (approved):
LGTM.

packages/next-extraction-plugin/README.md
```diff
@@ -11,24 +10,52 @@ yarn add --dev @griffel/next-extraction-plugin
 npm install --save-dev @griffel/next-extraction-plugin
 ```
 
+If you haven't installed yet `@griffel/webpack-loader`, please also install it:
```
ling1726, line 13:
```suggestion
Please install `@griffel/webpack-loader` if you haven't done so already:
```

ling1726 (approved):

PR #199, opened 2022-08-30
fix(devtools): normalize relative path in source url

YuanboXue-Amber:
Devtools read source url from error stack. Sometimes these urls contain relative path, depending on how the app is bundled. However chrome always store resources using absolute path. This PR normalizes the source url so we can find it and read it using chrome resources API.

Screenshot 2022-08-30 at 9 51 17

### DIFF
--- a/packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
+++ b/packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
@@ -33,7 +33,19 @@ function parseSourceUrl(sourceUrlWithLoc: string): MappedPosition | null {
   const paths = sourceUrlWithLoc.split(':');
   const column = Number(paths.pop());
   const line = Number(paths.pop());
-  const source = paths.join(':');
+
+  let source = paths.join(':');
+  try {
+    // url can contain relative path when webpack is used; normalize it
+    const { protocol, pathname } = new URL(source);
+    if (!protocol.startsWith('http')) {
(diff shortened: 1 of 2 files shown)


packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
```diff
@@ -33,7 +33,19 @@ function parseSourceUrl(sourceUrlWithLoc: string): MappedPosition | null {
   const paths = sourceUrlWithLoc.split(':');
   const column = Number(paths.pop());
   const line = Number(paths.pop());
-  const source = paths.join(':');
+
+  let source = paths.join(':');
+  try {
```
ling1726, line 38: nit: you could pull it to a separate function like `normalizeWebpackRelativeUrl` and then document + test

ling1726 (approved):

packages/devtools/src/sourceMap/loadOriginalSourceLoc.ts
YuanboXue-Amber: The test won't work expected unfortunately as chrome has a [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=869291) when parsing url from unknown origin.
the `pathname` is different between chrome and node:
- chrome 
  Screenshot 2022-09-06 at 11 19 55
- node 
  Screenshot 2022-09-06 at 11 20 27

ling1726: Ouch, yeah thanks for the info

PR #200, opened 2022-08-30
fix(devtools): show override when overriding rule is the same

YuanboXue-Amber:
## This PR fix 2 problems in devtools.

Here's an example showing both problems: (using devtools on fluent v9 button component):

Screenshot 2022-08-30 at 19 07 47

- Problem 1: when overriding rule is the same rule, devtools is not able to determine the overridng rule.
In the above example, 'medium' and the 'primary' slot contains the same style for focus. The one in 'medium' overrides the one in 'primary'. But devtools failed to detect it.

- Problem 2: when selector contains className like `.fxxxx.yyyy` where `.fxxxx` is the className used by griffel, devtools failed to parse this className

After this PR, both problems are fixed. 
Additionally, when clicking on a style, devtools highlights the overriding style. But in the case of problem 1 where both styles are the same, this highlighting feature highlights both styles (because they have the same griffel className). I like it. So I updated the highlighting to always highlight both the selected style, and the overriding style. I think it makes more sense than how it works before.

After PR: 
Screenshot 2022-08-30 at 19 20 44

### DIFF
--- a/change/@griffel-core-7824828a-f980-4c5d-ba33-06d73ffb351d.json
+++ b/change/@griffel-core-7824828a-f980-4c5d-ba33-06d73ffb351d.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "fix devtools to show override when the overriding rule is the same",
+  "comment": "fix: update debug data collection to handle duplicate overrides",
   "packageName": "@griffel/core",
   "email": "yuanboxue@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@griffel-core-7824828a-f980-4c5d-ba33-06d73ffb351d.json
+++ b/change/@griffel-core-7824828a-f980-4c5d-ba33-06d73ffb351d.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix devtools to show override when the overriding rule is the same",
+  "packageName": "@griffel/core",
+  "email": "yuanboxue@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/core/src/devtools/getDebugTree.test.ts
+++ b/packages/core/src/devtools/getDebugTree.test.ts
@@ -17,7 +17,7 @@ const options: MakeStylesOptions = {
 const findSequenceHash = (classNames: string) =>
   classNames.split(' ').find(className => className.startsWith(SEQUENCE_PREFIX));
 
-const souceURLregex = /.*\/.*:[0-9]+:[0-9]+/; // url with line and column number
+const sourceURLregex = /.*\/.*:[0-9]+:[0-9]+/; // url with line and column number
 
 describe('getDebugTree', () => {
   it.each<'ltr' | 'rtl'>(['ltr', 'rtl'])('returns styles merge tree when dir=%p', dir => {
@@ -51,7 +51,7 @@ describe('getDebugTree', () => {
           },
           sequenceHash: sequenceGrid,
           slot: 'grid',
-          sourceURL: expect.stringMatching(souceURLregex),
+          sourceURL: expect.stringMatching(sourceURLregex),
         },
         {
           children: [
@@ -69,7 +69,7 @@ describe('getDebugTree', () => {
               },
               sequenceHash: sequenceBlock,
               slot: 'block',
-              sourceURL: expect.stringMatching(souceURLregex),
+              sourceURL: expect.stringMatching(sourceURLregex),
             },
           ],
           debugClassNames: [
@@ -129,7 +129,7 @@ describe('getDebugTree', () => {
           },
           sequenceHash: sequenceBlueBlock,
           slot: 'blueBlock',
-          sourceURL: expect.stringMatching(souceURLregex),
+          sourceURL: expect.stringMatching(sourceURLregex),
         },
         {
           children: [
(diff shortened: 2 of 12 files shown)


change/@griffel-core-7824828a-f980-4c5d-ba33-06d73ffb351d.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix devtools to show override when the overriding rule is the same",
```
layershifter, line 3:
```suggestion
  "comment": "fix: update debug data collection to handle duplicate overrides",
```

?

packages/core/src/devtools/getDebugTree.test.ts
```diff
@@ -37,6 +37,22 @@ describe('getDebugTree', () => {
...
+              overriddenBy: undefined,
+            },
+          ],
+          direction: dir,
+          rules: {
+            f13qh94s: '.f13qh94s{display:grid;}',
+          },
+          sequenceHash: sequenceGrid,
+          slot: 'grid',
+          sourceURL: expect.stringMatching(souceURLregex),
```
layershifter, line 54:
```suggestion
          sourceURL: expect.stringMatching(sourceURLregex),
```

A typo...

packages/core/src/devtools/utils.test.ts
```diff
@@ -72,4 +73,34 @@ describe('getDebugClassNames', () => {
...
   });
+
+  it('handles parent debug node contains overriding style, but the overriding rule is the same', () => {
+    const lookupItem: LookupItem = [
+      {
+        De3pzq: 'fdmssx0',
+      },
+      'ltr',
+    ];
+    const parentlLookupItem: LookupItem = [
```
layershifter, line 84:
```suggestion
    const parentLookupItem: LookupItem = [
```

A typo

packages/core/src/devtools/utils.ts
```diff
@@ -21,11 +22,30 @@ export function getDebugClassNames(
...
+      } else if (matching && parentLookupItem[0][propertyHash]) {
+        // parent node has current className (style), and has current selector:
+        // case 1. style is not overriden during current merging; it may be overriden in higher level of merging
+        // case 2. style is overriden in current merging by exactly the same rule in sibling nodes
+        const siblingHasSameRule = overridingSiblings?.filter(
+          overridingSibling =>
+            overridingSibling.debugClassNames.filter(
+              ({ className: siblingClassName }) => siblingClassName === className,
+            ).length,
+        ).length;
```
layershifter, line 37:
```suggestion
            ).length > 0,
        ).length > 0;
```

_nit_ Let's makes "booleans" booleans 🐱

packages/devtools/src/MonolithicRulesView.tsx
```diff
@@ -30,20 +30,29 @@ const useSingleRuleStyles = makeStyles({
 const SingleRuleView: React.FC<{ rule: RuleDetail; indent?: boolean }> = ({ rule, indent }) => {
   const { highlightedClass, setHighlightedClass } = useViewContext();
 
+  const [clicked, setClicked] = React.useState(false);
```
layershifter, line 33: `selected`/`setSelected`?

packages/devtools/src/getMonolithicCSSRules.test.ts
```diff
@@ -158,11 +158,12 @@ describe('getMonolithicCSSRules', () => {
     });
   });
 
-  it('group atomic css when class name selector has selector in front', () => {
+  it.only('group atomic css when class name selector has selector in front', () => {
```
layershifter, line 161:
```suggestion
  it('group atomic css when class name selector has selector in front', () => {
```

Can you please install [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest) and set `"jest/no-focused-tests": "error"`? 🐱 (in a separate PR as I am doing the same mistake often)

packages/devtools/src/getMonolithicCSSRules.ts
```diff
@@ -1,7 +1,8 @@
 import * as stylis from 'stylis';
 import type { AtomicRules, MonolithicAtRules, MonolithicRules, RuleDetail } from './types';
 
-const griffelClassNameRegex = /^\.f[0-9a-z]+$/;
+// match className like '.fvnxzrg' or '.fvnxzrg.fui-focus-visible'
+const griffelClassNameRegex = /^\.(f[0-9a-z]+)(\..*|$)/;
```
layershifter, line 5: Will it handle `body .fvnxzrg`?

layershifter (approved):

packages/devtools/src/getMonolithicCSSRules.test.ts
YuanboXue-Amber: yep let me do that in a separate PR :)

packages/devtools/src/getMonolithicCSSRules.ts
YuanboXue-Amber: Yeah it should, I have a test with rules like `'[data-keyboard-nav] .fabcde1{color:red;}'`

YuanboXue-Amber: The parser will try to find the griffel className, taken it out and leave the rest: `body .fvnxzrg` -> `body `

packages/devtools/src/getMonolithicCSSRules.test.ts
YuanboXue-Amber: Here we go: #208

PR #207, opened 2022-09-06
fix: handle nesting in comma separated selectors

### LINKED ISSUE #204: core: comma-separated nested selectors are not handled properly
Version: `@griffel/core@1.6.0`.
Repro: https://stackblitz.com/edit/node-sovmxb?file=index.test.js

---

```js
makeStyles({
  ":hover,:focus-within": {
    "::before": {
      color: "orange"
    }
  }
});
```

It's expected that following code will generate this rule:

```css
.fij4gri:hover::before,
.fij4gri:focus-within::before {
  color: orange;
}
```

But it generates:

```css
.fij4gri:hover,
/*           👆 "::before" is lost */
.fij4gri:focus-within::before {
  color: orange;
}
```

It's a real usage in Fluent UI: https://github.com/microsoft/fluentui/blob/94ab82dd00db0bf177908e7ccc60125efaa4e22d/packages/react-components/react-spinbutton/src/components/SpinButton/useSpinButtonStyles.ts#L113-L121

### LINKED ISSUE #206: core: :global() selector does not handle spaces
**Version**: `@griffel/core@1.6.0`.
**Repro**: https://stackblitz.com/edit/node-ab1ctm?file=index.test.js

---

```js
makeStyles({
  ":global(body) :focus": { color: "green" }
});
```

It's expected that following code will generate this rule:

```css
body .frou13r :focus {
  color: green;
}
```

But it generates:

```css
body .frou13r:focus {
/*           👆 " " is lost */
  color: green;
}
```

layershifter:
Fixes #204.
Fixes #206.

---

## Changes in selectors handling

In `resolveStyleRules()` & `compileCSS()` I renamed `pseudo` to `selectors` and changed their type to be `string[]`:
- To reflect what it contains as `selectors` can contain not only pseudo selectors.
- To handle selectors properly as it was a problem without a solution:
  ```ts
  makeStyles({
    ":hover,:focus-within": {
      "::before": {
        color: "orange"
      }
    }
  });
  ```

  Before we were concatenating inputs with selectors and getting `":hover,:focus-within::before"` that will be compiled by Stylis to:
  ```css
  .class:hover,.class:focus-within::before {}
  ```
  And without proper wrapping with curly braces we cannot handle it. With selectors as an array with can send as input `:hover,:focus-within { ::before { } }` that will be handled properly.

## Changes in `:global()` handling

Because before we didn't know scopes of selectors we were doing `.trim()` and were losing `" "` in rules (#206):

https://github.com/microsoft/griffel/blob/0e358cbda0f663246ed7c1cc1ad7db0bdc86c22c/packages/core/src/runtime/compileCSS.ts#L92-L93

Currently I kept existing hack with `:global()` in place to avoid more changes in this PR, but in a follow up - I will replace it with a Stylis plugin.

### DIFF
--- a/packages/core/src/runtime/compileCSS.test.ts
+++ b/packages/core/src/runtime/compileCSS.test.ts
@@ -1,21 +1,23 @@
 import { compileCSS, CompileCSSOptions, normalizePseudoSelector } from './compileCSS';
 
-const defaultOptions: Pick<CompileCSSOptions, 'rtlClassName' | 'className' | 'media' | 'pseudo' | 'support' | 'layer'> =
-  {
-    className: 'foo',
-    rtlClassName: 'rtl-foo',
-    media: '',
-    pseudo: '',
-    support: '',
-    layer: '',
-  };
+const defaultOptions: Pick<
+  CompileCSSOptions,
+  'rtlClassName' | 'className' | 'media' | 'selectors' | 'support' | 'layer'
+> = {
+  className: 'foo',
+  rtlClassName: 'rtl-foo',
+  media: '',
+  selectors: [],
+  support: '',
+  layer: '',
+};
 
 describe('compileCSS', () => {
   it('handles pseudo', () => {
     expect(
       compileCSS({
         ...defaultOptions,
-        pseudo: ':hover',
+        selectors: [':hover'],
         property: 'color',
         value: 'red',
       }),
@@ -27,7 +29,7 @@ describe('compileCSS', () => {
     expect(
       compileCSS({
         ...defaultOptions,
-        pseudo: ':focus:hover',
+        selectors: [':focus:hover'],
         property: 'color',
         value: 'red',
       }),
@@ -98,11 +100,11 @@ describe('compileCSS', () => {
     `);
   });
 
-  it('handles rtl properties with preudo selectors', () => {
+  it('handles rtl properties with pseudo selectors', () => {
     expect(
       compileCSS({
         ...defaultOptions,
-        pseudo: ':before',
+        selectors: [':before'],
 
         property: 'paddingLeft',
         value: '10px',
@@ -140,7 +142,7 @@ describe('compileCSS', () => {
       expect(
         compileCSS({
           ...defaultOptions,
-          pseudo: ':global(body)',
+          selectors: [':global(body)'],
           property: 'color',
           value: 'red',
         }),
@@ -152,13 +154,13 @@ describe('compileCSS', () => {
       expect(
         compileCSS({
           ...defaultOptions,
-          pseudo: ':global(body) &',
+          selectors: [':global(.fui-FluentProvider)', '& .focus:hover'],
           property: 'color',
           value: 'red',
         }),
       ).toMatchInlineSnapshot(`
       Array [
-        "body .foo{color:red;}",
+        ".fui-FluentProvider .foo .focus:hover{color:red;}",
       ]
     `);
     });
@@ -167,7 +169,7 @@ describe('compileCSS', () => {
       expect(
         compileCSS({
           ...defaultOptions,
-          pseudo: ':global(body)',
+          selectors: [':global(body)'],
           property: 'paddingLeft',
           value: '10px',
 
--- a/packages/core/src/runtime/compileCSS.ts
+++ b/packages/core/src/runtime/compileCSS.ts
@@ -6,7 +6,7 @@ import { normalizeNestedProperty } from './utils/normalizeNestedProperty';
 export interface CompileCSSOptions {
   className: string;
 
-  pseudo: string;
+  selectors: string[];
   media: string;
   layer: string;
   support: string;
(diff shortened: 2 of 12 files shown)


packages/core/src/runtime/compileCSS.test.ts
```diff
@@ -152,13 +154,13 @@ describe('compileCSS', () => {
       expect(
         compileCSS({
           ...defaultOptions,
-          pseudo: ':global(body) &',
```
layershifter, line 155: The original selector is invalid, replaced it with a proper test case.

packages/core/src/runtime/compileCSS.ts
```diff
@@ -60,61 +60,69 @@ export function compileCSSRules(cssRules: string): string[] {
...
-  // Should be handled by namespace plugin of Stylis, is buggy now
-  // Issues are reported:
-  // https://github.com/thysultan/stylis.js/issues/253
-  // https://github.com/thysultan/stylis.js/issues/252
-  if (pseudo.indexOf(':global(') === 0) {
-    // 👇 :global(GROUP_1)GROUP_2
-    const GLOBAL_PSEUDO_REGEX = /global\((.+)\)(.+)?/;
-    const [, globalSelector, restPseudo = ''] = GLOBAL_PSEUDO_REGEX.exec(pseudo)!;
+        return `${normalizedPseudoSelector} { ${acc} }`;
+      }
```
layershifter, line 89: I moved the hack with small changes, will replace it with a Stylis plugin in a follow up PR.

packages/core/src/runtime/getStyleBucketName.ts
```diff
@@ -52,20 +57,22 @@ export function getStyleBucketName(pseudo: string, layer: string, media: string,
     return 't';
   }
 
-  const normalizedPseudo = pseudo.trim();
+  if (selectors.length > 0) {
+    const normalizedPseudo = selectors[0].trim();
```
layershifter, line 61: We take into account only the first pseudo selector, so it's find to use the first element.

packages/core/src/runtime/resolveStyleRules.test.ts
```diff
@@ -412,6 +412,21 @@ describe('resolveStyleRules', () => {
...
+              color: 'orange',
+            },
+          },
+        }),
+      ).toMatchInlineSnapshot(`
+        .fij4gri:hover::before,
+        .fij4gri:focus-within::before {
+          color: orange;
+        }
+      `);
```
layershifter, line 429: Test case from #204.

packages/core/src/runtime/resolveStyleRules.test.ts
```diff
@@ -573,18 +588,6 @@ describe('resolveStyleRules', () => {
...
-      expect(
-        resolveStyleRules({
-          ':global(body) &': { color: 'green' },
-        }),
-      ).toMatchInlineSnapshot(`
-        body .fm1e7ra {
-          color: green;
-        }
-      `);
-    });
```
layershifter, line 586: Again a test that does not have any sense.

packages/core/src/runtime/resolveStyleRules.test.ts
```diff
@@ -609,34 +612,38 @@ describe('resolveStyleRules', () => {
       `);
       expect(
         resolveStyleRules({
+          ':global(body):focus': { color: 'pink' },
```
layershifter, line 615: Now spaces are handled (#206).

packages/core/src/runtime/resolveStyleRules.test.ts
```diff
@@ -609,34 +612,38 @@ describe('resolveStyleRules', () => {
...
+      expect(
+        resolveStyleRules({
+          ':focus': { ':global(body)': { color: 'green' } },
+        }),
+      ).toMatchInlineSnapshot(`
+        body .fz7er5p:focus {
+          color: green;
+        }
+      `);
+    });
```
layershifter, line 646: As now we have scopes this test surprisingly works 🐱

packages/core/src/runtime/compileCSS.ts
```diff
@@ -60,50 +60,58 @@ export function compileCSSRules(cssRules: string): string[] {
...
-    : `{ ${hyphenateProperty(property)}: ${value}; }`;
+  if (pseudos.length > 0) {
+    cssRule = pseudos.reduceRight((acc, selector, index) => {
+      // Should be handled by namespace plugin of Stylis, is buggy now
+      // Issues are reported:
+      // https://github.com/thysultan/stylis.js/issues/253
+      // https://github.com/thysultan/stylis.js/issues/252
+      if (selector.indexOf(':global(') === 0) {
+        // 👇 :global(GROUP_1)GROUP_2
+        const GLOBAL_PSEUDO_REGEX = /global\((.+)\)(.+)?/;
```
ling1726, line 75: maybe hoist the regex so it doesn't get created on every iteration ?

packages/core/src/runtime/compileCSS.ts
```diff
@@ -6,7 +6,7 @@ import { normalizeNestedProperty } from './utils/normalizeNestedProperty';
 export interface CompileCSSOptions {
   className: string;
 
-  pseudo: string;
+  selectors: string[];
```
ling1726, line 9: would the length of selectors ever exceed 2 ? It seems that the only case we are handling here is pseudo selector + element. It shouldn't be possible to go further since it would be invalid css ?

layershifter: Can you please provide an example?

ling1726 (approved):

miroslavstastny (approved):

PR #209, opened 2022-09-14
Add TSDoc to unsupported CSS properties

NotWoods:
When teammates on Loop start working with Griffel, the shorthand type errors confuse them. There's no clear reason why the error shows up unless you know to look at Griffel's docs ahead of time. As most developers are just working with the Fluent UI library, they don't know to check here.

Adding back an interface for the unsupported shorthand properties makes this documentation easier to discover, as using "Go to definition" will link developers to the interface. I've added TSDoc comments to the overall interface as well as to each property.

Ideally the comments on the properties would show up when using makeStyles too, but I couldn't get that to work in VSCode. Nevertheless, I think this change helps developers understand why Griffel shows type errors for shorthands.

### DIFF
--- a/change/@griffel-core-7c2fb1fa-c42f-4195-8948-dbac68156b32.json
+++ b/change/@griffel-core-7c2fb1fa-c42f-4195-8948-dbac68156b32.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add TSDoc to unsupported CSS properties",
+  "packageName": "@griffel/core",
+  "email": "tigeroakes@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@griffel-core-7c2fb1fa-c42f-4195-8948-dbac68156b32.json
+++ b/change/@griffel-core-7c2fb1fa-c42f-4195-8948-dbac68156b32.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Add TSDoc to unsupported CSS properties",
+  "comment": "fix: add TSDoc to unsupported CSS properties",
   "packageName": "@griffel/core",
   "email": "tigeroakes@microsoft.com",
   "dependentChangeType": "patch"
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1,9 +1,108 @@
 import * as CSS from 'csstype';
-import { UNSUPPORTED_CSS_PROPERTIES } from './constants';
 
 export type GriffelStylesCSSValue = string | 0;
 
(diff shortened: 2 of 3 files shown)


packages/core/src/types.ts
```diff
@@ -1,9 +1,108 @@
...
 
 export type GriffelStylesCSSValue = string | 0;
 
-export type GriffelStylesUnsupportedCSSProperties = Record<keyof typeof UNSUPPORTED_CSS_PROPERTIES, never>;
+/**
+ * Griffel doesn't support expansion of CSS shorthands.
+ * They can be replaced either with the corresponding longhand properties or with the `shorthands` helper functions.
+ * @see https://griffel.js.org/react/guides/limitations#css-shorthands-are-not-supported
+ */
+export interface GriffelStylesUnsupportedCSSProperties extends Record<keyof CSS.StandardShorthandProperties, never> {
```
ling1726, line 10:
```suggestion
export interface GriffelStylesUnsupportedCSSProperties extends Record<keyof typeof UNSUPPORTED_CSS_PROPERTIES, never> {
```

Is it still possible to use the internal unsupported properties lookup ? We do explicit handling there to strip those properties from the styles. It might seem stupid but at least we know that there is a link between what is happening in the library and the types here

packages/core/src/types.ts
```diff
@@ -1,9 +1,108 @@
...
+  /** @deprecated */
+  all: never;
+  /** @deprecated Use corresponding longhand properties such as `animationName` and `animationDuration` instead. */
+  animation: never;
+  /** @deprecated Use corresponding longhand properties such as `backgroundImage` and `backgroundSize` instead. */
```
ling1726, line 15: Awesome work with the documentation!

ling1726 (approved):
LGTM

ling1726:
@NotWoods  can you run `yarn change` and generate a changefile for the changes  https://github.com/microsoft/griffel/blob/main/CONTRIBUTING.md#submitting-changes

change/@griffel-core-7c2fb1fa-c42f-4195-8948-dbac68156b32.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add TSDoc to unsupported CSS properties",
```
ling1726, line 3:
```suggestion
  "comment": "fix: Add TSDoc to unsupported CSS properties",
```

bsunderhus (approved):
LGTM! this definitely helps with documentation!

layershifter (approved):

layershifter:
@NotWoods great job ❤️ FYI: there is [a ESLint plugin](https://github.com/microsoft/griffel/tree/main/packages/eslint-plugin) that aims to provide similar experience.

PR #215, opened 2022-09-20
docs: bump docusaurus & force dark theme

layershifter:


apps/website/src/components/Playground/code/app.js
```diff
@@ -1,11 +1,18 @@
 import hljs from 'highlight.js/lib/core';
-import 'highlight.js/styles/vs.css';
+import 'highlight.js/styles/vs2015.css';
```
layershifter, line 2: Also changing a theme for preview in try out.

bsunderhus: nit: Can you add this to the description?

bsunderhus (approved):

PR #216, opened 2022-09-20
docs: fix breadcrumbs on try out pages

layershifter:
Depends on #215. Fixes breabcrumbs on Try out page.

### Details

The PR adds a patch to `@docusaurus/theme-common` to handle hashes as a part of URL (I am not aware about any other way to handle it except ejecting the whole breadcrumb component).

It also changes items in a custom sidebar to group them into "Try out" category that will be used in breadcrumbs.

### Before

image

### After

image

PR #217, opened 2022-09-20
docs: add pages about Atomic CSS, build time optimisations

layershifter:
~~Requires #215, #216.~~

- Updates a page about Atomic CSS with real content
- Adds pages about ahead-of-time compilation
- Adds pages about CSS extraction

### DIFF
--- a/.yarn/patches/mdx-mermaid-npm-1.3.2-e8a6432fce.patch
+++ b/.yarn/patches/mdx-mermaid-npm-1.3.2-e8a6432fce.patch
@@ -1,0 +1,13 @@
+diff --git a/lib/Mermaid.mjs b/lib/Mermaid.mjs
+index 5140cd5e556ede3debb33ae2202f04480b288d2f..204a3f97d590c9a384ed1bdcfcf7b0dc1375c616 100644
+--- a/lib/Mermaid.mjs
++++ b/lib/Mermaid.mjs
+@@ -45,7 +45,7 @@ const Mermaid = ({ chart, config: configSrc }) => {
+     if (typeof window === 'undefined') {
+         return React.createElement("div", { className: "mermaid", "data-mermaid-src": chart }, chart);
+     }
+-    const config = useMemo(() => typeof configSrc === 'string' ? JSON.parse(configSrc) : configSrc, [configSrc]);
++    const config = useMemo(() => typeof configSrc === 'string' ? JSON.parse(configSrc) : configSrc || {}, [configSrc]);
+     const html = document.querySelector('html');
+     const [rerender, setRerender] = useState(false);
+     const theme = useMemo(() => getTheme(html, config), [config, rerender]);
--- a/apps/website/docs/react/ahead-of-time-compilation/introduction.md
+++ b/apps/website/docs/react/ahead-of-time-compilation/introduction.md
@@ -1,0 +1,14 @@
+---
+sidebar_position: 1
+---
+
+# Introduction
+
+While there is nothing wrong with the associated runtime costs of a CSS-in-JS engine, larger and more complex applications might want to optimize for performance.
+
+Griffel does the expensive runtime work only happens on the first render of the component. This one-time work can be further optimized at build time by pre-computing and transforming styles.
+
+## What to use?
+
+- For library developers, please use [Babel preset](/react/ahead-of-time-compilation/with-babel)
+- For application developers, please use [Webpack loader](/react/ahead-of-time-compilation/with-webpack) (supports Next.js)
--- a/apps/website/docs/react/ahead-of-time-compilation/introduction.md
+++ b/apps/website/docs/react/ahead-of-time-compilation/introduction.md
@@ -6,7 +6,7 @@ sidebar_position: 1
 
 While there is nothing wrong with the associated runtime costs of a CSS-in-JS engine, larger and more complex applications might want to optimize for performance.
 
-Griffel does the expensive runtime work only happens on the first render of the component. This one-time work can be further optimized at build time by pre-computing and transforming styles.
+Griffel only does the expensive runtime on the first render of the component. This work can be further optimized at build time by pre-computing and transforming styles.
 
 ## What to use?
 
--- a/apps/website/docs/react/ahead-of-time-compilation/technical-details.md
+++ b/apps/website/docs/react/ahead-of-time-compilation/technical-details.md
@@ -1,0 +1,146 @@
+---
+sidebar_position: 4
+---
+
+# Technical details
+
+## What is being optimized with AOT?
+
+:::info
+
+Style resolution only needs to happen on the initial render of a component. Therefore, without build time optimization the performance is comparable with the 2nd and consecutive renders.
+
+It is reasonable to introduce build time optimization if/when it is required.
+
+:::
+
+```jsx
+import { makeStyles } from '@griffel/react';
+
+// 1. Invocation of makeStyles creates a styling hook that will be used inside a component.
+const useStyles = makeStyles({
+  root: { paddingLeft: '1px', display: 'flex' },
+});
+
+function Component() {
+  // 2. The hook call resolves styles which are injected into the document.
+  const classes = useStyles();
+
+  return <div className={classes.root} />;
+}
+```
+
+You can look at the graph below which describes what work is done during style resolution:
+
+```mermaid
+stateDiagram-v2
+    direction LR
+
+    INVOKE_USE_STYLES: useStyles() invocation
+    COMPUTE_RTL_STYLES: Compute RTL styles
+    COMPUTE_CSS_CLASSES: Compute CSS classes
+    COMPUTE_CSS_RULES: Compute CSS rules
+    INSERT_TO_DOM: Insert to DOM
+
+    INVOKE_USE_STYLES --> COMPUTE_RTL_STYLES
+    COMPUTE_RTL_STYLES --> COMPUTE_CSS_CLASSES
+    COMPUTE_CSS_CLASSES --> COMPUTE_CSS_RULES
+    COMPUTE_CSS_RULES --> INSERT_TO_DOM
+
+    note right of INVOKE_USE_STYLES
+     {
+        &nbsp;&nbsp;paddingLeft: '1px',
+        &nbsp;&nbsp;display: 'flex'
+     }
+    end note
+    note right of COMPUTE_RTL_STYLES
+     {
+        &nbsp;&nbsp;paddingLeft: '1px',
+        &nbsp;&nbsp;paddingRight: '1px',
+        &nbsp;&nbsp;display: 'flex'
+      }
+    end note
+    note right of COMPUTE_CSS_CLASSES
+     {
+       &nbsp;&nbsp;paddingLeft: '1px', // .f10xn8zz
+       &nbsp;&nbsp;paddingRight: '1px', // .f136y8j8
+       &nbsp;&nbsp;display: 'flex' // .f22iagw
+     }
+    end note
+    note right of COMPUTE_CSS_RULES
+     .f10xn8zz { padding-left: 1px }
+     .f136y8j8 { padding-right: 1px }
+     .f22iagw { display: flex }
+    end note
+```
+
+:::note
+
+This work only happens once, during first render.
+
+:::
+
+The final result before the CSS rules are inserted into DOM can be compiled ahead of time during build time through the methods described above.
+Once the styles of our simple example are transformed at build time the resulting bundle contains a result similar to what is in our diagram.
+
+The actual runtime code of `makeStyles` is completely stripped from the bundle and replaced with a lightweight function (`__styles`) that simply concatenates the CSS classes and inserts them to DOM.
+
+```jsx
+const useStyles = __styles(
+  {
+    root: {
+      mc9l5x: 'f22iagw',
+      uwmqm3: ['f10xn8zz', 'f136y8j8'],
+    },
+  },
+  {
+    d: ['.f22iagw{display:flex;}', '.f10xn8zz{padding-left:1px;}', '.f136y8j8{padding-right:1px;}'],
+  },
+);
+
+function Component() {
+  const classes = useStyles();
+
+  return <div className={classes.root} />;
+}
+```
+
+## Module evaluation process
+
+Let's consider the following scenario:
+
+```js
+// constants.js
+export const PADDING_TOKEN = '1px';
+```
+
+```js
+// common.js
+export const commonStyles = () => ({
+  display: 'flex',
+  justifyContent: 'center',
+  alignItems: 'center',
+});
+```
+
+```js
+// styles.js
+import { makeStyles } from '@griffel/react';
+import { PADDING_TOKEN } from './constants';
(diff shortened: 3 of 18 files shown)


.yarn/patches/mdx-mermaid-npm-1.3.2-e8a6432fce.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/lib/Mermaid.mjs b/lib/Mermaid.mjs
```
layershifter, line 1: https://github.com/sjwall/mdx-mermaid/issues/67

apps/website/docs/react/guides/child-window-rendering.md
```diff
@@ -2,10 +2,9 @@
 sidebar_position: 5
 ---
 
-# Child Window/Shadow DOM Rendering
-
-When rendering in the main browser window, many components will need access to window or document to apply styles, listening for events, or measuring things. However it is possible to render to child windows and elements hosted in iframes.
+# Child Window
```
layershifter, line 5: I removed "Shadow DOM" there to fit a single line in the sidebar

packages/webpack-extraction-plugin/README.md
```diff
@@ -56,6 +56,11 @@ module.exports = {
         exclude: /node_modules/,
         use: {
           loader: '@griffel/webpack-loader',
+          options: {
+            babelOptions: {
+              presets: ['@babel/preset-typescript'],
+            },
+          },
```
layershifter, line 63: It's mandatory in 99% percent of cases for TS, so it's better to include it.

apps/website/docs/react/ahead-of-time-compilation/introduction.md
```diff
@@ -0,0 +1,14 @@
+---
+sidebar_position: 1
+---
+
+# Introduction
+
+While there is nothing wrong with the associated runtime costs of a CSS-in-JS engine, larger and more complex applications might want to optimize for performance.
+
+Griffel does the expensive runtime work only happens on the first render of the component. This one-time work can be further optimized at build time by pre-computing and transforming styles.
```
ling1726, line 9:
```suggestion
Griffel only does the expensive runtime on the first render of the component. This work can be further optimized at build time by pre-computing and transforming styles.
```

apps/website/docs/react/ahead-of-time-compilation/technical-details.md
```diff
@@ -0,0 +1,146 @@
+---
+sidebar_position: 4
+---
+
+# Technical details
+
+## What is being optimized with AOT?
```
ling1726, line 7:
```suggestion
## What is being optimized with AOT (**A**head **O**f **T**ime) complilation ?
```

apps/website/docs/react/ahead-of-time-compilation/with-webpack.md
```diff
@@ -0,0 +1,130 @@
...
+        },
+      },
+    ],
+  },
+};
+```
+
+## Usage with Next.js
+
+Next.js allows to [tweak Webpack's config](https://nextjs.org/docs/api-reference/next.config.js/custom-webpack-config) so the same options are applicanle to its config.
```
ling1726, line 90:
```suggestion
Next.js lets users [tweak Webpack's config](https://nextjs.org/docs/api-reference/next.config.js/custom-webpack-config) so the same options are applicable to its config.
```

apps/website/docs/react/css-extraction/introduction.md
```diff
@@ -0,0 +1,23 @@
+---
+sidebar_position: 1
+---
+
+# Introduction
+
+:::caution
+
+This technology is experimental, please [report issues](https://github.com/microsoft/griffel/issues) if you will face them.
```
ling1726, line 9:
```suggestion
This technology is experimental, please [report any issues](https://github.com/microsoft/griffel/issues) if found.
```

apps/website/docs/react/css-extraction/introduction.md
```diff
@@ -0,0 +1,23 @@
...
+
+:::
+
+While [ahead-of-time compilation allows](/react/ahead-of-time-compilation/introduction) performs optimization to reduce runtime parts, the goal of CSS extraction is remove runtime insertion to DOM and produce CSS stylesheets.
```
ling1726, line 13:
```suggestion
While [ahead-of-time compilation allows](/react/ahead-of-time-compilation/introduction) performs optimization to reduce runtime work, the goal of CSS extraction is to remove runtime insertion to DOM and produce CSS stylesheets.
```

apps/website/docs/react/css-extraction/introduction.md
```diff
@@ -0,0 +1,23 @@
...
+
+## When to use it?
+
+It's designed to be used **only** in applications.
+
+## How it works?
```
ling1726, line 19:
```suggestion
## How does it work?
```

ling1726:
```suggestion
## How it works
```

Both are fine

apps/website/docs/react/css-extraction/with-webpack.md
```diff
@@ -0,0 +1,104 @@
...
+        test: /\.css$/,
+        use: [MiniCssExtractPlugin.loader, 'css-loader'],
+      },
+    ],
+  },
+  plugins: [new MiniCssExtractPlugin(), new GriffelCSSExtractionPlugin()],
+};
+```
+
+- `mini-css-extract-plugin` is not mandatory and configured there for example, you can use `style-loader` or other tooling to inject CSS on a page
```
ling1726, line 79:
```suggestion
- `mini-css-extract-plugin` is not mandatory and is used as am example, you can use `style-loader` or other tooling to inject CSS on a page
```

apps/website/docs/react/css-extraction/with-webpack.md
```diff
@@ -0,0 +1,104 @@
...
+
+For better performance (to process less files) consider to use `include` for `GriffelCSSExtractionPlugin.loader`:
```
ling1726, line 81:
```suggestion
For better performance (to process less files) consider using `include` for `GriffelCSSExtractionPlugin.loader`:
```

apps/website/docs/react/guides/atomic-css.md
```diff
@@ -4,4 +4,104 @@ sidebar_position: 1
 
 # Atomic CSS
 
-This page is not ready yet.
+Atomic CSS is an opposite approach to monolithic classes. In Atomic CSS every property-value is written as a single CSS rule.
```
ling1726, line 7:
```suggestion
Atomic CSS is a constrasting approach to monolithic classes. In Atomic CSS every property-value is written as a single CSS rule.
```

apps/website/docs/react/guides/atomic-css.md
```diff
@@ -4,4 +4,104 @@ sidebar_position: 1
...
+/* Atomic CSS */
+.display-flex {
+  display: flex;
+}
+.align-items-center {
+  align-items: center;
+}
+```
+
+Atomic CSS enables CSS rules re-use and reduces total amount of defined rules. Other components can re-use the same CSS rules and create fewer rules as styles grow.
```
ling1726, line 32:
```suggestion
Atomic CSS enables CSS rules re-use and reduces the total amount of defined rules. Other components can re-use the same CSS rules and create fewer rules as styles grow.
```

apps/website/docs/react/guides/atomic-css.md
```diff
@@ -4,4 +4,104 @@ sidebar_position: 1
...
+
+## Style overrides
+
+While Griffel works by default in runtime it has [ahead-of-time compilation](/react/ahead-of-time-compilation/introduction). AOT compilation process [cannot create new rules at runtime](/react/guides/limitations) and because of that we rely on [merging CSS rules](/react/api/merge-classes).
```
ling1726, line 36:
```suggestion
While Griffel works by default in runtime it has [ahead-of-time compilation](/react/ahead-of-time-compilation/introduction). The AOT compilation process [cannot create new rules at runtime](/react/guides/limitations) and because of that, we rely on [merging CSS rules](/react/api/merge-classes) to apply conditional styles at runtime.
```

apps/website/docs/react/guides/atomic-css.md
```diff
@@ -4,4 +4,104 @@ sidebar_position: 1
...
+  align-items: center;
+}
+
+/* Group B */
+.align-items-end {
+  align-items: end;
+}
+```
+
+There are two pieces that are used to create classes: property name (`display`, `align-items`) & value (`flex`, `center`, `end`). Using [mergeClasses](/react/api/merge-classes) we ensure that only one set of properties is applied. The last defined property-value pair wins.
```
ling1726, line 55:
```suggestion
Classes are created using two parts: property name (`display`, `align-items`) & value (`flex`, `center`, `end`). Using [mergeClasses](/react/api/merge-classes) we ensure that only one set of properties is applied. The last defined property-value pair wins.
```

apps/website/docs/react/guides/atomic-css.md
```diff
@@ -4,4 +4,104 @@ sidebar_position: 1
...
+
+```html
+<!-- Example: a result of merging classes -->
+<button class="display-flex align-items-end"></button>
+```
+
+## LVFHA order of pseudo classes
+
+Pseudo classes in CSS (like `:hover`, `:link`, `:focus`, `:active`) have equal specificity and the result is determined by [order of appearance](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Cascade_and_inheritance#understanding_the_cascade).
+When atomic classes inserted to DOM, they are inserted based on definition, basically based on usage. This means the resulting styles would be based on which components’ styles were rendered first. A [CodeSandbox](https://codesandbox.io/s/lvfha-puzzle-ihbict) that shows this behavior.
```
ling1726, line 65:
```suggestion
When atomic classes are inserted to DOM, they are inserted based on definition, basically based on usage. This means the resulting styles would be based on which components’ styles were rendered first. A [CodeSandbox](https://codesandbox.io/s/lvfha-puzzle-ihbict) that shows this behavior.
```

apps/website/docs/react/guides/atomic-css.md
```diff
@@ -4,4 +4,104 @@ sidebar_position: 1
...
+- `:hover`
+- `:active`
+
+The last defined pseudo wins.
+
+## Trade-offs
+
+### Larger classes
+
+Atomic CSS increases size of HTML markup as more classes added to elements: every CSS rules adds a class to the element.
```
ling1726, line 83:
```suggestion
Atomic CSS increases size of HTML markup as more classes added to elements: every CSS rule adds a class to the element.
```

apps/website/docs/react/guides/atomic-css.md
```diff
@@ -4,4 +4,104 @@ sidebar_position: 1
...
+| Class names            |     1 |     10 |     50 |    100 |    200 |    500 | 1000   |
+| ---------------------- | ----: | -----: | -----: | -----: | -----: | -----: | ------ |
+| Elapsed (ms)           |    53 |     57 |     74 |     89 |    127 |    207 | 372    |
+| Over baseline (ms)     |     0 |      4 |     21 |     36 |     74 |    154 | 319    |
+| Match attempts         | 10000 |  10000 |  10000 |  10000 |  10000 |  10000 | 10000  |
+| Elapsed time per match |   N/A | 0.0004 | 0.0021 | 0.0036 | 0.0074 | 0.0154 | 0.0319 |
+
+_Measured in Edge 105 & Surface Go 2 on battery power._
+
+Cases when elements have 100 classes and more are rare, but you can easily get there with [nested selectors](/react/api/make-styles#nesting-selector) what we recommend to avoid.
```
ling1726, line 101:
```suggestion
Cases when elements have 100 classes and more are rare, but you can easily reach that threshold with [nested selectors](/react/api/make-styles#nesting-selector), which we recommend avoiding.
```

apps/website/docs/react/guides/limitations.md
```diff
@@ -4,6 +4,52 @@ sidebar_position: 2
...
+    [`.${props.area}`]: { display: 'block' },
+  });
+}
+```
+
+As Griffel performs ahead-of-time [compilation](/react/ahead-of-time-compilation/introduction) values used in CSS rules should be static to be compiled.
+
+### Workarounds
+
+- Use inline styles on elements. They don't have the best performance, but it will be faster than invoke any CSS-in-JS for frequently changing values.
```
ling1726, line 25:
```suggestion
- Use inline styles on elements. They don't have the best performance, but it will be faster than invoking any CSS-in-JS for frequently changing values.
```

apps/website/docs/react/guides/limitations.md
```diff
@@ -4,6 +4,52 @@ sidebar_position: 2
...
+    /* your styles */
+  });
+
+  function App(props) {
+    const classes = useClasses();
+    return <div style={{ color: props.color }} />;
+  }
+  ```
+
+- Use local CSS variables for nested/pseudo selectors. Prefer to use inline styles, but they can't be used for pseudo selector, for example.
```
ling1726, line 38: why recommend inline styles over local css variables ?

apps/website/docs/react/guides/limitations.md
```diff
@@ -4,6 +4,52 @@ sidebar_position: 2
 
 # Limitations
 
+## Runtime styles
+
+Styles can't be created at runtime which includes dynamic selectors as well.
```
ling1726, line 9: This section is missing a part on conditional styles

layershifter: Did not get it, can you please clarify?

layershifter: Got it, will push updates 👍

apps/website/docs/react/guides/limitations.md
layershifter: IMO it's redundant for one offs as it requires additional work without obvious benefits: you still need to pass a CSS variable via inline styles.

```ts
makeStyles({
  color: { color: 'var(--color)', }
})

<div style={{ '--color': props.color }} />

// vs.

<div style={{ color: props.color }} />
```

apps/website/docs/react/guides/limitations.md
layershifter: @ling1726 added in 1bbbeae.

apps/website/docs/react/guides/limitations.md
```diff
@@ -4,6 +4,73 @@ sidebar_position: 2
...
+    [`.${props.area}`]: { display: 'block' },
+  });
+}
+```
+
+As Griffel performs ahead-of-time [compilation](/react/ahead-of-time-compilation/introduction) values used in CSS rules should be static so that they can be compiled.
+
+### Workarounds
+
+- Use variants enumeration. If you know values in advance and a set is limited the best option is to enumerate them.
```
ling1726, line 25:
```suggestion
- Enumerate variants. If you know values in advance and a set is limited the best option is to enumerate them.
```

ling1726 (approved):

PR #224, opened 2022-09-27
chore: create stylis plugin to handle :global()

### LINKED ISSUE #205: core: use stylis plugin for global instead of a hack
Currently we are using a hack to handle `:global()` selector:

https://github.com/microsoft/griffel/blob/a03d53c8fc6d4aebbf7fd4234a6a75cd5ebc7ee7/packages/core/src/runtime/compileCSS.ts#L69-L73

The blocking issue (https://github.com/thysultan/stylis/issues/253) to implement it as a plugin was resolved. We should implement it as a plugin to unblock upcoming work related to `makeResetStyles`.

layershifter:
Fixes #205.


packages/core/src/runtime/stylis/globalPlugin.ts
```diff
@@ -0,0 +1,55 @@
...
+                  process.env.NODE_ENV === 'production'
+                    ? ''
+                    : `A token before "global()" selector should be ":": ${JSON.stringify(acc)}`,
+                );
+              }
+
+              delete acc[acc.length - 1];
+              delete children[index + 1];
+
+              return acc;
```
layershifter, line 45: I don't like logic there, so opened for better proposals

packages/core/src/runtime/stylis/globalPlugin.ts
```diff
@@ -0,0 +1,55 @@
+import { RULESET, tokenize } from 'stylis';
+import type { Middleware } from 'stylis';
+
+export const globalPlugin: Middleware = element => {
+  switch (element.type) {
+    case RULESET:
+      if (typeof element.props === 'string') {
+        throw new Error(
```
ling1726, line 8: Why would throwing an error in production with an empty message be a good thing ?

layershifter: Save bundle size a bit. I will change it

ling1726: I'm also wondering why throw the error in production at all ? it seems inconsistent with other errors we have which only throw at devtime

layershifter: > I'm also wondering why throw the error in production at all ? it seems inconsistent with other errors we have which only throw at devtime

It's consistent: we _throw_ always, but we _don't log_ in production...

layershifter: Okay, except this one:
https://github.com/microsoft/griffel/blob/ebe44fdd9c34a87a4106c95c5b47f5719c215e8f/packages/react/src/makeStyles.ts#L32-L41

But I don't see a reasonable way to not throw there.

For example there, we can do:

```diff
      if (typeof element.props === 'string') {
+        if (process.env.NODE_ENV !== 'production') {
        throw new Error(
          `"element.props" has type "string" (${JSON.stringify(
            element.props,
            null,
            2,
          )}), it's not expected. Please report a bug if it happens.`,
        );
+     }
+     return;
      }
```

But what to do for another one instead of throwing? Ideas?

ling1726 (approved):

PR #232, opened 2022-09-27
feat: add makeResetStyles to @griffel/core

### LINKED ISSUE #225: core: implement makeResetStyles
There is no goal to change Griffel to follow another approach, but there are edge cases where Atomic CSS does not show the best performance.

Once an element has many HTML class names each pointing to different CSS rules, Blink based browsers (Chrome, Edge) have linear performance degradation based on an amount of classes.

| Class names            |     1 |     10 |     50 |    100 |    200 |    500 | 1000   |
| ---------------------- | ----: | -----: | -----: | -----: | -----: | -----: | ------ |
| Elapsed (ms)           |    53 |     57 |     74 |     89 |    127 |    207 | 372    |
| Over baseline (ms)     |     0 |      4 |     21 |     36 |     74 |    154 | 319    |
| Match attempts         | 10000 |  10000 |  10000 |  10000 |  10000 |  10000 | 10000  |
| Elapsed time per match |   N/A | 0.0004 | 0.0021 | 0.0036 | 0.0074 | 0.0154 | 0.0319 |

_Measured in Edge 105 & Surface Go 2 on battery power._

---

Cases when elements have 100 classes and more are rare, but you can easily get there with nested selectors what we recommend to avoid:

```ts
makeStyles({
  button: {
    "& .abc": {
      ...shorthands.padding("4px"),
      ...shorthands.margin("4px"),
      ...shorthands.border("4px", "solid", "red")
    },
    "& .fde": {
      ...shorthands.padding("4px"),
      ...shorthands.margin("4px"),
      ...shorthands.border("4px", "solid", "red")
    }
  }
});
```

For example, the snippet above generates 40 classes. The similar problem is mentioned in [styleq: Supporting high-performance layouts](https://www.npmjs.com/package/styleq#supporting-high-performance-layouts)

> For example, a component library may optimize the "reset" styles for its core components by flattening those styles, and then inserting those rules into the CSS style sheet before all the atomic CSS. That way atomic CSS will always override the reset rules, and the layout performance of the core components will be significantly improved.
>
> https://www.npmjs.com/package/styleq#supporting-high-performance-layouts

Basically, the goal to implement a way to generate such rules. We don't want to bake this into `makeStyles`, so the decision is to create `makeResetStyles`.

layershifter:
Fixes #225.

This PR implements `makeResetStyles()` function in `@griffel/core`.

- It generates a monolithic classes with `rf` prefix
- Produced CSS rules are insert into new `r` bucket, so any Atomic rules will have higher precedence
- Shorthands are allowed in `makeResetStyles` calls are there are no reasons to forbid them

### DIFF
--- a/change/@griffel-core-40e0578a-53de-40b8-9ed2-958d4fe70437.json
+++ b/change/@griffel-core-40e0578a-53de-40b8-9ed2-958d4fe70437.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "feat: add makeResetStyles to @griffel/core",
+  "packageName": "@griffel/core",
+  "email": "olfedias@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@griffel-core-40e0578a-53de-40b8-9ed2-958d4fe70437.json
+++ b/change/@griffel-core-40e0578a-53de-40b8-9ed2-958d4fe70437.json
@@ -1,5 +1,5 @@
 {
-  "type": "patch",
+  "type": "minor",
   "comment": "feat: add makeResetStyles to @griffel/core",
   "packageName": "@griffel/core",
   "email": "olfedias@microsoft.com",
--- a/e2e/typescript/src/assets/fixture-reset.ts
+++ b/e2e/typescript/src/assets/fixture-reset.ts
@@ -143,6 +143,11 @@ assertType({
   boxSizing: 'outline-box',
   zIndex: 1,
 });
+assertType({
+  zIndex: 1,
+  // @ts-expect-error type check still fails on outline-box, not on any other line
+  boxSizing: 'outline-box',
+});
 
 assertType({
   // @ts-expect-error Object is not assignable to CSS property
--- a/e2e/typescript/src/assets/fixture-reset.ts
+++ b/e2e/typescript/src/assets/fixture-reset.ts
@@ -1,0 +1,321 @@
+import { GriffelResetStyle } from '@griffel/core';
+
+function assertType(style: GriffelResetStyle): GriffelResetStyle {
+  return style;
+}
+
+// Animation
+assertType({ animationName: 'foo' });
+
+// Basic styles
+//
+
+assertType({ flex: 0 });
+assertType({ flex: 1 });
+assertType({ zIndex: 0 });
+assertType({ zIndex: 1 });
+
+assertType({ padding: '5px' });
+assertType({ paddingLeft: '5px' });
+assertType({ color: 'beige' });
+
+// CSS variables
+//
+
+assertType({ fontWeight: 'var(--foo)' });
+assertType({ flexShrink: 'var(--bar)' });
+assertType({ opacity: 'var(--baz)' });
+assertType({ zIndex: 'var(--qux)' });
+
+assertType({ '--color': 'red' });
+
+// Mixins
+//
+
+assertType({
+  ...{ padding: '5px' },
+  ...{ paddingLeft: '5px' },
+  ...{ color: 'red' },
+});
+
+// Strict selector defined via "CSS.Pseudos"
+//
+
+assertType({
+  ':hover': { flexShrink: 0 },
+  ':focus': { flexShrink: 'initial' },
+  ':active': { zIndex: 0 },
+  ':visited': { zIndex: 1 },
+});
+
+assertType({
+  ':hover': { fontWeight: 'var(--foo)' },
+  ':focus': { flexShrink: 'var(--bar)' },
+  ':active': { opacity: 'var(--bar)' },
+  ':visited': { zIndex: 'var(--qux)' },
+});
+
+assertType({
+  ':hover': { color: 'beige' },
+  ':active': { padding: '5px' },
+  ':focus': { paddingLeft: '5px' },
+});
+
+assertType({
+  ':hover': { '--color': 'red' },
+});
+
+// Custom selectors
+//
+
+assertType({
+  ':hover:focus': { flexShrink: 0 },
+  ':hover:active': { flexShrink: 'initial' },
+  ':hover:visited': { zIndex: 0 },
+  ':hover:focus-visible': { zIndex: 1 },
+});
+
+assertType({
+  ':link:hover': { fontWeight: 'var(--foo)' },
+  ':link:focus': { flexShrink: 'var(--bar)' },
+  ':link:active': { opacity: 'var(--bar)' },
+  ':link:visited': { zIndex: 'var(--qux)' },
+});
+
+assertType({
+  '.bar': { color: 'beige' },
+  '.foo': { padding: '5px' },
+  '.qux': { paddingLeft: '5px' },
+});
+
+assertType({
+  '.bar': { '--color': 'red' },
+});
+
+// Nested custom selectors
+//
+
+assertType({
+  '.foo': {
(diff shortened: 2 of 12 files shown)


e2e/typescript/src/assets/fixture-reset.ts
```diff
@@ -0,0 +1,324 @@
+import { GriffelResetStyle } from '@griffel/core';
```
layershifter, line 1: It's a copy of `fixture.ts` with exclusions of shorthands that are not banned.

change/@griffel-core-40e0578a-53de-40b8-9ed2-958d4fe70437.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
bsunderhus, line 2:
```suggestion
  "type": "minor",
```
Shouldn't this be a minor? as it's a new feature....

packages/core/src/runtime/resolveResetStyleRules.ts
```diff
@@ -0,0 +1,122 @@
...
+import { hyphenateProperty } from './utils/hyphenateProperty';
+
+/**
+ * @internal
+ */
+export function createStringFromStyles(styles: GriffelResetStyle) {
+  let ltrCSS = '';
+  let rtlCSS = '';
+
+  // eslint-disable-next-line guard-for-in
```
bsunderhus, line 21: is this for performance reasons? the `styles` here is provided by the user, this might leak some weird things if someone is tweaking around a custom object as styles.

packages/core/src/runtime/resolveResetStyleRules.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      ltrCSS += `${hyphenateProperty(property)}:${value};`;
+      rtlCSS += `${hyphenateProperty(rtlProperty)}:${rtlValue};`;
+
+      continue;
+    }
+
+    if (property === 'animationName') {
+      // TODO: handle animations
+      throw new Error();
```
bsunderhus, line 41: Shouldn't this be encapsulated inside a `process.env.NODE_ENV !== 'production'`? we're trying to avoid throwing things in production...

Also, empty Error? 😅

packages/core/src/runtime/resolveResetStyleRules.test.ts
```diff
@@ -0,0 +1,188 @@
...
+        "rfvhnavh",
+        null,
+        Array [
+          "@layer utilities{color:orange;:focus{color:yellow;}}",
+        ],
+      ]
+    `);
+  });
+
+  it('handles support queries', () => {
```
ling1726, line 148: worth testing nested `@supports @media` queries ?

packages/core/src/makeResetStyles.ts
```diff
@@ -0,0 +1,35 @@
+import { resolveResetStyleRules } from './runtime/resolveResetStyleRules';
+import type { GriffelResetStyle, MakeStylesOptions } from './types';
+
+/**
+ * @internal
+ */
+export function makeResetStyles(styles: GriffelResetStyle) {
```
ling1726, line 7: worth adding a benchmark in the benchmark app ?

ling1726 (approved):

packages/core/src/runtime/resolveResetStyleRules.ts
ling1726: Let's track this in a follow up issue

packages/core/src/runtime/resolveResetStyleRules.ts
layershifter: > is this for performance reasons?

Yes, we use it consistently across in other places, too. It's not expected to see custom objects there, but we can change it if someone will report issues.

packages/core/src/runtime/resolveResetStyleRules.ts
layershifter: I was sure that I added support for it 🙁 But... #239.

packages/core/src/makeResetStyles.ts
layershifter: Yes (#240), but once I will have a next PR with changes to `@griffel/react` out.

packages/core/src/runtime/resolveResetStyleRules.test.ts
layershifter: Added in d12d11f. Good part is that Stylis handles it for us for it's more solid actually.

e2e/typescript/src/assets/fixture-reset.ts
```diff
@@ -0,0 +1,321 @@
...
+  // @ts-expect-error "paddingLeft" cannot be numeric value
+  paddingLeft: 5,
+});
+assertType({
+  // @ts-expect-error "0" is invalid value for "color"
+  color: 0,
+});
+
+assertType({
+  // @ts-expect-error type check still fails on outline-box, not on any other line
```
miroslavstastny, line 142: nit: I would like to see the test for the two lines swapped as well.
Is it expected to fail on the first property in the object (first line) or on the invalid `boxSizing`?

layershifter: Added in a487682.

> Is it expected to fail on the first property in the object (first line) or on the invalid boxSizing?

It fails on invalid `boxSizing`

packages/core/src/constants.ts
```diff
@@ -6,6 +6,9 @@ export const DATA_BUCKET_ATTR = 'data-make-styles-bucket';
 /** @internal */
 export const HASH_PREFIX = 'f';
 
+/** @internal */
+export const RESET_HASH_PREFIX = 'rf';
```
miroslavstastny, line 10: What is the reason for the prefix to be 2 chars instead of just one?

layershifter: No reason, 8676f41

packages/core/src/runtime/resolveResetStyleRules.ts
```diff
@@ -0,0 +1,122 @@
...
+import { isLayerSelector } from './utils/isLayerSelector';
+import { isNestedSelector } from './utils/isNestedSelector';
+import { isSupportQuerySelector } from './utils/isSupportQuerySelector';
+import { isObject } from './utils/isObject';
+import { hyphenateProperty } from './utils/hyphenateProperty';
+
+/**
+ * @internal
+ */
+export function createStringFromStyles(styles: GriffelResetStyle) {
```
miroslavstastny, line 17: This function is currently private to this file. Why is it exported?

packages/core/src/runtime/resolveResetStyleRules.ts
```diff
@@ -0,0 +1,122 @@
...
+  let ltrCSS = '';
+  let rtlCSS = '';
+
+  // eslint-disable-next-line guard-for-in
+  for (const property in styles) {
+    const value = styles[property as keyof GriffelStyle];
+
+    // eslint-disable-next-line eqeqeq
```
miroslavstastny, line 25: Chro!

miroslavstastny (approved):

packages/core/src/runtime/resolveResetStyleRules.ts
layershifter: No reason, I covered it with tests initially, but then switched to `resolveResetStyleRules`.

PR #234, opened 2022-09-29
fix: don't extract http:, https:, data: and hash only urls as assets

### LINKED ISSUE #233: babel-preset: produces invalid import for `filter: url(#a)`
Input:

```js
import { makeStyles } from "@griffel/core";

export const useClasses = makeStyles({
  root: {
    filter: `url(#a)`
  }
})
```

Output:

```js
import _asset from "./#a";
import { __styles } from "@griffel/core";
export const useClasses = __styles({
  "root": {
    "Bhu2qc9": "f1rurakv"
  }
}, {
  "d": [`.f1rurakv{-webkit-filter:url(src/#a);filter:url(${_asset});}`]
});
```

`filter: url(#a)` refers to a `filter` element with ID `a` in an SVG element, it's not an asset that needs to be imported, so it should output as is.

Also, even if `#a` is an asset, `-webkit-filter:url(src/#a)` should be `-webkit-filter:url(${_asset})`.

Complete repro: https://github.com/yume-chan/griffel-filter-url

yume-chan:
Fixes #233 

Points to note:

1. `data:` URLs are actually extracted as asset imports (not ignored) before this change.
2. I also ignored `http:` and `https:` URLs, they should be external resources, not part of module resolution.
3. I can't find where prefixes are added, so `-webkit-filter:url(packages/babel-preset/__fixtures__/assets/a.svg#a);` is still incorrect.
4. There is another rarer case: a `data:` URL contains an SVG image, which contains other relative `url()`s, this is still not handled.


packages/babel-preset/__fixtures__/assets/code.ts
```diff
@@ -6,6 +6,12 @@ import empty from './empty.jpg';
...
 export const useStyles = makeStyles({
   rootA: { backgroundImage: `url(${blank})` },
-  rootB: { backgroundImage: `url(${blankDuplicate})` },
+  rootB: { backgroundImage: `url("${blankDuplicate}")` },
   rootC: { backgroundImage: `url(${empty})` },
+  httpUrl: { backgroundImage: 'url(http://www.example.com)' },
+  httpsUrl: { backgroundImage: 'url(https://www.example.com)' },
+  httpsUrlWithQuotes: { backgroundImage: 'url("https://www.example.com")' },
+  dataUrl: { backgroundImage: 'url(data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7)' },
+  hashOnly: { filter: 'url(#a)' },
```
layershifter, line 15: Can you please move out assets that are not "extractable" to a separate fixture? I.e. create `packages/babel-preset/__fixtures__/assets-urls/code.ts`, `packages/babel-preset/__fixtures__/assets/output.ts` and add to `packages/babel-preset/src/transformPlugin.test.ts`?

packages/babel-preset/src/assets/replaceAssetsWithImports.ts
```diff
@@ -40,10 +40,23 @@ export function replaceAssetsWithImports(
...
+          return;
+        }
+
+        // Handle `filter: url(#id)` and `filter: url(./a.svg#id)`
+        const [pathname, hash] = result.url.split('#');
+        if (!pathname) {
+          return;
+        }
+
+        const assetIdentifier = getAssetIdentifier(pathname);
```
layershifter, line 53: Is it mandatory to do any changes there? We ran there on `StringLiteral` and it could be there only if an asset contains imports.

packages/babel-preset/src/assets/normalizeStyleRules.ts
```diff
@@ -23,15 +23,24 @@ export function normalizeStyleRule(
...
-  return `${result.prefix}${normalizeAssetPath(path, projectRoot, filename, url)}${result.suffix}`;
+  // Handle `filter: url(#id)` and `filter: url(./a.svg#id)`
+  const [pathname, hash] = url.split('#');
+  if (!pathname) {
+    return `${result.prefix}${url}${result.suffix}`;
+  }
+
+  return `${result.prefix}${normalizeAssetPath(path, projectRoot, filename, pathname)}${hash ? `#${hash}` : ''}${
+    result.suffix
+  }`;
```
layershifter, line 43: So, only for assets that are in this branch we are need post-processing, right? 🤔 

Can we add a special prefix `griffel||` or special symbol `\f` to identify them in a simpler way in `replaceAssetsWithImports`?

packages/babel-preset/src/assets/replaceAssetsWithImports.ts
yume-chan: Didn't `normalizeStyleRules.ts` output string literal in both case? As noted in OP, current version will transform `data:` URLs to assets:

Input:

```js
import { makeStyles } from "@griffel/core";

export const useClasses = makeStyles({
  root: {
    backgroundImage: `url(data:foo)`
  }
})
```

Output:

```js
import _asset from "../data:test";
import { __styles } from "@griffel/core";
export const useClasses = __styles({
  "root": {
    "Bcmaq0h": "f1tpm5in"
  }
}, {
  "d": [`.f1tpm5in{background-image:url(${_asset});}`]
});
```

layershifter:
> I can't find where prefixes are added, so -webkit-filter:url(packages/babel-preset/__fixtures__/assets/a.svg#a); is still incorrect.

Interesting that it's not processed in the same way, I will check later...

packages/babel-preset/src/assets/replaceAssetsWithImports.ts
layershifter: @yume-chan yeah, sorry for confusion. You're correct 👍

yume-chan:
> I can't find where prefixes are added, so `-webkit-filter:url(packages/babel-preset/__fixtures__/assets/a.svg#a);` is still incorrect.

I know, the whole CSS block is passed into `replaceAssetsWithImports` as one string, `".fv04sme{-webkit-filter:url(packages/babel-preset/__fixtures__/assets/a.svg#a);filter:url(packages/babel-preset/__fixtures__/assets/a.svg#a);}"` in this case, so it contains two `url()`, but the Regex only matched the later one. 

It may be difficult to correctly parse the block.

yume-chan:
Now it also has problem with multiple URLs like `backgroundImage: "url(./a.png), url(./b.png)"` (https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Backgrounds_and_Borders/Using_multiple_backgrounds), only the last one is extracted. `maskImage` and `filter` also supports multiple values.

The best solution might be to use a CSS parser for semantically parsing the input and handle each rule differently. Considering the complexity, I don't think I will have time to complete that.

layershifter:
> The best solution might be to use a CSS parser for semantically parsing the input and handle each rule differently. Considering the complexity, I don't think I will have time to complete that.

Good call, I will look into it 👀

layershifter:
> The best solution might be to use a CSS parser for semantically parsing the input and handle each rule differently. Considering the complexity, I don't think I will have time to complete that.

No worries, I created an issue #238 and you can check a fix in #237. Once the fix will be merged I will rebase your changes 😉

layershifter:
@yume-chan #237 was merged, do you need help with rebasing changes?

yume-chan:
@layershifter I have rebased main into my branch and it's ready to be reviewed.

packages/babel-preset/src/assets/normalizeStyleRules.ts
```diff
@@ -23,15 +23,24 @@ export function normalizeStyleRule(
...
   const result = parseStringWithUrl(ruleValue);
-  // Quotes in URL are optional, so we can also normalize them
+  // Quotes in URL are optional, we can also normalize them,
+  // so two rules are treated as same when they only differ in quotations.
   // https://www.w3.org/TR/CSS2/syndata.html#value-def-uri
-  const url = result.url.replace(/['|"](.+)['|"]/, '$1');
+  const url = result.url.replace(/^(['|"])(.+)\1$/, '$2');
 
-  if (url.startsWith('data:')) {
+  if (url.startsWith('data:') || url.startsWith('http:') || url.startsWith('https:')) {
```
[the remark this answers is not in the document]
yume-chan: Added in [isAssetUrl()](https://github.com/microsoft/griffel/pull/234/files#diff-0fa3456d4b7f31d80034d3d2371838879370f0d693a069767c29386e6cce7fe7)

packages/babel-preset/__fixtures__/assets/code.ts
yume-chan: Tests are separated as #237 does.

packages/babel-preset/src/assets/normalizeStyleRules.ts
yume-chan: I think the current approach (using `isAssertUrl` at both places) is clear enough.

layershifter (approved):
Really like how it was simplified, great job ❤️

layershifter:
Released in `@griffel/babel-preset@1.4.1` & `@griffel/webpack-loader@2.1.1` 🚢

PR #236, opened 2022-10-03
feat: add support for makeResetStyles transforms

### LINKED ISSUE #227: babel-preset/webpack-loader: implement makeResetStyles support
Our Babel preset and Webpack loader that perform AOT should handle calls of `makeResetStyles`, see #225.

layershifter:
Fixes #227.

This PR adds new tests and modifies `@griffel/babel-preset` or `@griffel/webpack-loader` to handle `makeResetStyles` transforms.

`resetImportName` added to `modules` options of Babel preset and will be documented in a separate PR.


packages/webpack-loader/project.json
```diff
@@ -16,8 +16,7 @@
       "options": {
         "jestConfig": "packages/webpack-loader/jest.config.js",
         "passWithNoTests": true
-      },
-      "dependsOn": [{ "target": "build", "projects": "dependencies" }]
```
layershifter, line 20: I changed `@griffel/react` to use externals, so we need to build less code during each test and we don't need to build packages before

packages/webpack-loader/src/webpackLoader.test.ts
```diff
@@ -175,46 +174,73 @@ function testFixture(fixtureName: string, options: CompileOptions = {}) {
...
+          { moduleSource: 'make-styles', importName: 'makeStyles', resetImportName: 'createResetStyles' },
+        ]),
+      ).toBe(true);
+
+      expect(
+        shouldTransformSourceCode(`import { Button } from "@fluentui/react"`, [
+          { moduleSource: '@griffel/react', importName: 'makeStyles', resetImportName: 'makeResetStyles' },
+        ]),
+      ).toBe(false);
+    });
```
layershifter, line 226: The same test, just moved into `describe` block

packages/webpack-loader/src/webpackLoader.test.ts
```diff
@@ -175,46 +174,73 @@ function testFixture(fixtureName: string, options: CompileOptions = {}) {
...
 
 describe('shouldTransformSourceCode', () => {
-  it('handles defaults', () => {
-    expect(shouldTransformSourceCode(`import { makeStyles } from "@griffel/react"`, undefined)).toBe(true);
-    expect(shouldTransformSourceCode(`import { Button } from "@fluentui/react"`, undefined)).toBe(false);
+  describe('handles defaults', () => {
+    it('makeStyles', () => {
+      expect(shouldTransformSourceCode(`import { makeStyles } from "@griffel/react"`, undefined)).toBe(true);
+      expect(shouldTransformSourceCode(`import { Button } from "@fluentui/react"`, undefined)).toBe(false);
+    });
```
layershifter, line 181: The same test, just moved into `describe` block

packages/webpack-loader/src/webpackLoader.test.ts
```diff
@@ -175,46 +174,73 @@ function testFixture(fixtureName: string, options: CompileOptions = {}) {
...
+          { moduleSource: 'make-styles', importName: 'createStyles' },
+        ]),
+      ).toBe(true);
+
+      expect(
+        shouldTransformSourceCode(`import { Button } from "@fluentui/react"`, [
+          { moduleSource: '@griffel/react', importName: 'makeStyles' },
+        ]),
+      ).toBe(false);
+    });
```
layershifter, line 207: The same test, just moved into `describe` block

packages/babel-preset/src/transformPlugin.ts
```diff
@@ -12,45 +18,63 @@ import { evaluatePaths } from './utils/evaluatePaths';
 import { BabelPluginOptions } from './types';
 import { validateOptions } from './validateOptions';
 
+enum FunctionKinds {
+  makeStyles = 'makeStyles',
+  makeResetStyles = 'makeResetStyles',
+}
```
layershifter, line 24: Leaves the space for transforms of `makeStaticStyles`

packages/babel-preset/src/transformPlugin.ts
```diff
@@ -12,45 +18,60 @@ import { evaluatePaths } from './utils/evaluatePaths';
...
+    throw callExpression.buildCodeFrameError(`${functionKind}() function accepts only a single param`);
   }
 
   const definitionsPath = argumentPaths[0];
 
-  if (!definitionsPath.isObjectExpression()) {
-    throw definitionsPath.buildCodeFrameError('makeStyles() function accepts only an object as a param');
+  if (functionKind === 'makeStyles') {
+    if (!definitionsPath.isObjectExpression()) {
+      throw definitionsPath.buildCodeFrameError('makeStyles() function accepts only an object as a param');
```
ling1726, line 46: both makeStyles and makeResetStyles accept an object only as param right ?

packages/babel-preset/src/types.ts
```diff
@@ -3,7 +3,11 @@ import type { EvalRule } from '@linaria/babel-preset';
 
 export type BabelPluginOptions = {
   /** Defines set of modules and imports handled by a transformPlugin. */
-  modules?: { moduleSource: string; importName: string }[];
+  modules?: {
+    moduleSource: string;
+    importName: string;
+    resetImportName?: string;
```
ling1726, line 9:
```suggestion
    makeStylesimportName: string;
    makeResetStylesImportName?: string;
```

it's a bit longer, but it might make things clearer in the plugin code

ling1726 (approved):
Changes look more or less OK to me, it's simply just extending the plugin state to differentiate between the two functions and then handling the replacement appropriately.

Evaluation hasn't changed at all from what I can see

layershifter:
> Evaluation hasn't changed at all from what I can see

Yup, that's correct.

packages/babel-preset/src/transformPlugin.ts
layershifter: Correct, will check it.

layershifter: Restored to the previous condition, don't remember why it looks like that

PR #237, opened 2022-10-03
fix: handle multiple url()

### LINKED ISSUE #238: babel-preset: does not handle multiple url()
## Versions 

`@griffel/babel-preset@1.3.7`

## Issue

_Raised by @yume-chan in https://github.com/microsoft/griffel/pull/234#issuecomment-1262265035._

Our `@griffel/babel-preset` doesn't handle multiple URLs in a string: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Backgrounds_and_Borders/Using_multiple_backgrounds

#### Input

```ts

import blank from './blank.jpg';
import empty from './empty.jpg';

  root: { backgroundImage: `url(${blank}), url(${empty})` },
});
```

#### Output 🚨

```ts
import _asset from './empty.jpg';
import blank from './blank.jpg';
import empty from './empty.jpg';
  {
    root: {
      Bcmaq0h: 'f1t88go',
    },
  },
  {
    d: [`.f1t88go{background-image:url(./blank.jpg),url(${_asset});}`],
    //                             💥 "./blank.jpg" should be an import
  },
);
```

#### Expected output

```ts
import _asset2 from './empty.jpg';
import _asset from './blank.jpg';
import blank from './blank.jpg';
import empty from './empty.jpg';
  {
    root: {
      Bcmaq0h: 'classname',
    },
  },
  {
    d: [`.classname{background-image:url(${_asset}),url(${_asset2});}`],
  },
);
```

layershifter:
Fixes #238. _See the bug for the issue description in detail._

The main change is that we use `tokenize` from `stylis` instead of regexps. This is more robust and allows to parse multiple `url()` expressions.

`@griffel/babel-preset` and `@griffel/webpack-extraction-plugin` were updated/adjusted to handle this scenario.


packages/babel-preset/src/assets/isAssetUrl.ts
```diff
@@ -0,0 +1,3 @@
+export function isAssetUrl(value: string): boolean {
```
layershifter, line 1: I intentionally didn't cover this by tests directly as it will be changed/removed in #234.

packages/babel-preset/src/assets/normalizeStyleRules.ts
```diff
@@ -22,16 +24,19 @@ export function normalizeStyleRule(
     return ruleValue;
   }
 
-  const result = parseStringWithUrl(ruleValue);
-  // Quotes in URL are optional, so we can also normalize them
-  // https://www.w3.org/TR/CSS2/syndata.html#value-def-uri
-  const url = result.url.replace(/['|"](.+)['|"]/, '$1');
+  return tokenize(ruleValue).reduce((result, token, index, array) => {
```
layershifter, line 27: Instead of a regex I am using `tokenize()` from stylis so we can parse all `url()` entries properly.

packages/babel-preset/src/transformPlugin.test.ts
```diff
@@ -35,11 +35,6 @@ pluginTester({
     // Fixtures
     //
     //
-    {
-      title: 'assets import handling',
-      fixture: path.resolve(fixturesDir, 'assets', 'code.ts'),
-      outputFixture: path.resolve(fixturesDir, 'assets', 'output.ts'),
-    },
```
layershifter, line 42: Moved into a separate group.

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
```diff
@@ -121,54 +121,56 @@ export const babelPluginStripGriffelRuntime = declare<
...
-                    t.stringLiteral(
-                      // When imports are inlined, we need to adjust the relative paths inside url(..) expressions
-                      // to allow css-loader resolve an imported asset properly
-                      transformUrl(
-                        state.filename!,
-                        options.resourceDirectory!,
-                        importDeclarationPath.get('source').node.value,
+                  const expressionPaths = literalPath.get('expressions');
+
+                  expressionPaths.map(expressionPath => {
```
layershifter, line 126: The inner part is 100% percent: now we parse all expressions instead of only first.

packages/babel-preset/src/assets/replaceAssetsWithImports.ts
```diff
@@ -29,6 +30,39 @@ export function replaceAssetsWithImports(
     return assetIdentifiers.get(assetPath) as t.Identifier;
   }
 
+  function buildTemplateLiteralFromValue(value: string): t.TemplateLiteral {
+    const tokens = tokenize(value);
+
+    const quasis: t.TemplateElement[] = [];
```
ling1726, line 36: what are quasis ?

layershifter: It's how Babel names string in template literals i.e. `foo` from `foo${bar}`:

image

https://babeljs.io/docs/en/babel-types#templateliteral

packages/babel-preset/src/assets/replaceAssetsWithImports.ts
```diff
@@ -29,6 +30,39 @@ export function replaceAssetsWithImports(
...
+    const expressions: t.Identifier[] = [];
+
+    let acc = '';
+
+    for (let i = 0, l = tokens.length; i < l; i++) {
+      for (; i < l; i++) {
+        acc += tokens[i];
+
+        if (tokens[i] === 'url') {
+          const url = tokens[i + 1].slice(1, -1);
```
ling1726, line 46: why ignore the first and last characters of a url ?

ling1726: I'm going by your example of tokens: `('.foo', '{', 'url', 'image.png', '}')`

yume-chan: The first and last characters are parens. `url(image.png)` tokenizes to `["url", "(image.png)"]`

packages/babel-preset/src/assets/replaceAssetsWithImports.ts
```diff
@@ -29,6 +30,39 @@ export function replaceAssetsWithImports(
...
+            acc = ')';
+            i++;
+
+            break;
+          }
+        }
+      }
+    }
+
+    quasis.push(t.templateElement({ raw: acc }, true));
```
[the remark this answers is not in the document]
ling1726: I can't think of much better....

but I don't quite understand why you need the second nested loop, once you find the `url` token you break out of the inner loop and the outer one continues where the inner one left off anyway ??

ling1726 (approved):
I have doubts about the ugly nested loop in the preset, but otherwise LGTM

packages/babel-preset/src/assets/replaceAssetsWithImports.ts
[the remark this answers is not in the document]
layershifter: > but I don't quite understand why you need the second nested loop, once you find the `url` token you break out of the inner loop and the outer one continues where the inner one left off anyway ??

Good catch, it's a leftover from the previous attempt. Removed.

PR #241, opened 2022-10-04
feat: add support for animationName in makeResetStyles

### LINKED ISSUE #239: core: support animations in makeResetStyles
#232 introduces `makeResetStyles`, but it lacks support for `animationName` property that handles `@keyframes`. This should be implemented.

layershifter:
Fixes #239.


packages/core/src/makeResetStyles.test.ts
```diff
@@ -33,4 +33,22 @@ describe('makeResetStyles', () => {
...
+
+    expect(renderer).toMatchInlineSnapshot(`
+      .rfgb6zd6 {
+        padding: 40px 20px 10px 5px;
+      }
+      .rfjhindo {
+        padding: 40px 5px 10px 20px;
+      }
+    `);
+  });
```
layershifter, line 53: Add a new test to check classes.

packages/core/src/runtime/resolveResetStyleRules.test.ts
```diff
@@ -1,5 +1,20 @@
...
+  },
+  print(value) {
+    /**
+     * test function makes sure that value is the guarded type
+     */
+    const _value = value as ReturnType<typeof resolveResetStyleRules>;
+
+    return prettier.format(_value[2].join(''), { parser: 'css' }).trim();
+  },
+});
```
layershifter, line 16: Added a serializer to make CSS output nicer.

ling1726 (approved):

PR #247, opened 2022-10-04
chore: add tests for hydration in makeResetStyles

### LINKED ISSUE #246: core/react: support for hydration in makeResetStyles
`makeStyles()` supports hydration process, but `makeResetStyles` does not (_while it should!_). Missing hydration process results in doubled styles definitions on a client.

Similar to #44.

layershifter:
Fixes #246.

Surprisingly it just works as rules generated by `makeResetStyles` are just CSS rules and are inserted into the same bucket🗿
This PR adds tests to ensure that it will _continue_ to work.

### DIFF
--- a/packages/react/src/createDOMRenderer.test.tsx
+++ b/packages/react/src/createDOMRenderer.test.tsx
@@ -1,9 +1,10 @@
-import { createDOMRenderer } from '@griffel/core';
+import { createDOMRenderer, mergeClasses } from '@griffel/core';
 import * as React from 'react';
 import { hydrate } from 'react-dom';
 import { renderToStaticMarkup } from 'react-dom/server';
 
 import { makeStyles } from './makeStyles';
+import { makeResetStyles } from './makeResetStyles';
 import { RendererProvider } from './RendererContext';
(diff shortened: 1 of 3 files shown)


packages/react/src/createDOMRenderer.test.tsx
```diff
@@ -13,11 +14,11 @@ describe('createDOMRenderer', () => {
 
     const clientRenderer = createDOMRenderer(document);
     const serverRenderer = createDOMRenderer(
-      // we should use "null" as "undefined" will fallback to "document" which is present in this environment
+      // we should use "null" as "undefined" will fall back to "document" which is present in this environment
       null as unknown as undefined,
```
bsunderhus, line 18: nit: perhaps we should update `createDOMRenderer` types to support this scenario also?

bsunderhus (approved):

ling1726 (approved):

packages/react/src/createDOMRenderer.test.tsx
layershifter: Not really, as it expected to have `document` on client side always.

PR #250, opened 2022-10-06
feat: add support for makeResetStyles transforms to CSS extraction

### LINKED ISSUE #231: css-extraction: handle makeResetStyles
Our `@griffel/webpack-extraction-plugin` should handle `makeResetStyles` calls properly. See #225 for more details.

layershifter:
Fixes #231.


packages/webpack-extraction-plugin/src/sortCSSRules.ts
```diff
@@ -1,6 +1,16 @@
...
+  RESET_HASH_PREFIX,
+  getStyleBucketName,
+  GriffelRenderer,
+  StyleBucketName,
+  styleBucketOrdering,
+} from '@griffel/core';
 import { COMMENT, compile, Element, KEYFRAMES, MEDIA, RULESET, serialize, stringify, SUPPORTS, tokenize } from 'stylis';
 
+function isResetClassName(className: string) {
+  return className[0] === '.' && className[1] === RESET_HASH_PREFIX;
```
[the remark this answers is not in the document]
ling1726: can stylis return something other than `.` as the first character ?

packages/webpack-extraction-plugin/src/sortCSSRules.ts
```diff
@@ -46,6 +56,10 @@ export function getStyleBucketNameFromElement(element: Element): StyleBucketName
     return 'k';
   }
 
+  if (isResetClassName(element.value)) {
+    return 'r';
+  }
```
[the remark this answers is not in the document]
ling1726: sorry I don't understand why this is not scalable

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
```diff
@@ -173,30 +205,39 @@ export const babelPluginStripGriffelRuntime = declare<
...
-              });
+                argumentPath.remove();
+              }
 
-              argumentPaths[1].remove();
+              if (functionKind === '__resetStyles') {
+                const cssRules = evaluationResult.value as string[];
+
+                state.cssRules!.push(...cssRules);
+                argumentPaths[2].remove();
```
ling1726, line 239:
```suggestion
                argumentPaths.remove();
```

you set the argumentPath correctly for both cases

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
```diff
@@ -101,16 +114,26 @@ export const babelPluginStripGriffelRuntime = declare<
...
+                argumentPath = argumentPaths[2];
+                functionKind = '__resetStyles';
+
+                calleePath.replaceWith(t.identifier('__resetCSS'));
+              } else {
+                return;
+              }
 
-              if (argumentPaths.length !== 2) {
+              if (functionKind === '__styles' && argumentPaths.length !== 2) {
```
ling1726, line 136: nit: you could just parametrize the two errors ?

ling1726: you could actually parametrize a lot:
* argument position
* argument length
* replacement function kind

It avoids some duplication, but it's not  a lot so it's your call

packages/webpack-extraction-plugin/src/sortCSSRules.ts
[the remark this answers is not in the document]
layershifter: It's not handling global selectors and few other cases:

```
body .baz { color: orange; }
/* ⬇️⬇️⬇️ */
/* element.value is "body .baz" */
```

I will work on integration of CSS extraction to one of products and will be looking into potential issues to see how to make sorting better (or reimplement it completely).

packages/webpack-extraction-plugin/src/sortCSSRules.ts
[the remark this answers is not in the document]
layershifter: Yes, `tr .foo` for example.

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
layershifter: Done in 1aeddeb

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
layershifter: Done in 99a99ac

ling1726 (approved):

PR #252, opened 2022-10-11
chore: Adds benchmark for v7 merge-styles

ling1726:
Installs `@uifabric/merge-styles` to the repo and creates a new benchmark implementation

### DIFF
--- a/apps/benchmark/src/implementations/merge-styles-v7/Dot.tsx
+++ b/apps/benchmark/src/implementations/merge-styles-v7/Dot.tsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import { DotProps } from '../types';
+import { mergeStyleSets } from '@uifabric/merge-styles';
+
+const getClassNames = (size: number, x: number, y: number, color: string) => {
+  return mergeStyleSets({
+    root: {
+      position: 'absolute',
+      cursor: 'pointer',
+      width: 0,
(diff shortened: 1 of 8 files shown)


apps/benchmark/src/implementations/merge-styles-v7/Dot.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { DotProps } from '../types';
+import { mergeStyleSets } from '@uifabric/merge-styles';
+
+const getClassNames = (size: number, x: number, y: number, color: string) => {
+  return mergeStyleSets({
+    root: {
```
layershifter, line 7:
```suggestion
```

nit: This should be just `mergeStyles` call, correct?

apps/benchmark/src/implementations/merge-styles-v7/View.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { mergeStyleSets } from '@uifabric/merge-styles';
+
+const getClassNames = () => {
+  return mergeStyleSets({
+    root: {
```
layershifter, line 6:
```suggestion
  return mergeStyles({
```

This should be just `mergeStyles` call, correct?

layershifter (approved):

PR #253, opened 2022-10-13
chore: remove mermaid patch

JPeer264:
I came across a little bug in `mdx-mermaid` and saw that in this repo is a patch in it. Thought it would be cleaner to have a default `{}` instead of actually patching it.

(ref: https://github.com/sjwall/mdx-mermaid/issues/67)


JPeer264:
@microsoft-github-policy-service agree [company="Stöcklmair Jan Peer"]

JPeer264:

layershifter:
@JPeer264 nice, thanks 👍 Don't know why it didn't work when I was trying to do the same 😨

layershifter (approved):

PR #255, opened 2022-10-18
fix: relax check in transform plugin

layershifter:
Fixes a problem from https://github.com/microsoft/fluentui/pull/25216.

https://github.com/microsoft/griffel/blob/b360b562bc7180fa13c6d0a1979673983bfe5ee0/packages/babel-preset/src/transformPlugin.ts#L44-L46

The check that we have currently is too strict and breaks legit cases:

```
ERR! ERR! SyntaxError: /mnt/vss/_work/1/s/packages/react-components/react-button/src/components/Button/useButtonStyles.ts: makeResetStyles() function accepts only an object as a param
ERR! ERR!   13 | var buttonSpacingSmall = '3px';
ERR! ERR!   14 | var buttonSpacingMedium = '5px';
ERR! ERR! > 15 | var useBaseClassName = makeResetStyles(Object.assign({
ERR! ERR!      |                                        ^
ERR! ERR!   16 |   alignItems: 'center',
ERR! ERR!   17 |   boxSizing: 'border-box',
ERR! ERR!   18 |   display: 'inline-flex',
```

This PR simplifies it to match returned types. The test was removed as I can't build a compilable code to throw there: [`t.callExpression`](https://babeljs.io/docs/en/babel-types#callexpression) accepts `Array<Expression | SpreadElement | JSXNamespacedName | ArgumentPlaceholder>`, but I can't imagine/create a case when `JSXNamespacedName | ArgumentPlaceholder` will be arguments.

### DIFF
--- a/packages/babel-preset/src/transformPlugin.ts
+++ b/packages/babel-preset/src/transformPlugin.ts
@@ -32,7 +32,7 @@ function getDefinitionPathFromCallExpression(
   functionKind: FunctionKinds,
   callExpression: NodePath<t.CallExpression>,
 ): NodePath<t.Expression | t.SpreadElement> {
-  const argumentPaths = callExpression.get('arguments') as NodePath<t.Node>[];
+  const argumentPaths = callExpression.get('arguments');
   const hasValidArguments = Array.isArray(argumentPaths) && argumentPaths.length === 1;
 
   if (!hasValidArguments) {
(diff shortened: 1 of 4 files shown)


packages/babel-preset/src/transformPlugin.ts
```diff
@@ -41,11 +41,11 @@ function getDefinitionPathFromCallExpression(
 
   const definitionsPath = argumentPaths[0];
 
-  if (!definitionsPath.isObjectExpression()) {
-    throw definitionsPath.buildCodeFrameError(`${functionKind}() function accepts only an object as a param`);
+  if (definitionsPath.isExpression() || definitionsPath.isSpreadElement()) {
```
bsunderhus, line 44: Would be nice to have a check for those cases you don't imagine happening... `JSXNamespacedName | ArgumentPlaceholder`

bsunderhus (approved):

PR #259, opened 2022-10-24
fix: add export for __resetCSS

layershifter:
Follow up for #250.

- Fixes function name
- Exports `__resetCSS` function


packages/react/src/__resetCSS.ts
```diff
@@ -7,7 +7,7 @@ import { useTextDirection } from './TextDirectionContext';
  * @internal
  */
 // eslint-disable-next-line @typescript-eslint/naming-convention
-export function __resetStyles(ltrClassName: string, rtlClassName: string | null) {
+export function __resetCSS(ltrClassName: string, rtlClassName: string | null) {
```
layershifter, line 10: 🤦

bsunderhus (approved):

PR #260, opened 2022-10-24
feat: add validation into mergeClasses for makeResetStyles calls

### LINKED ISSUE #230: core: ensure that there no multiple occupancies of classes produced by makeResetStyles in mergeClasses calls
See #225 for details.

We should add a runtime validation to `mergeClasses()` to ensure that the following scenario is impossible:

```tsx
const useClassNameA = makeResetStyles()
const useClassNameB = makeResetStyles()

function App() {
  const classNameA = useClassNameA()
  const classNameB = useClassNameB()

  return <div className={mergeClasses(classNameA, classNameB)} />
}
```

### LINKED ISSUE #261: jest-serializer: implement makeResetStyles support
Our Jest serializer does not handle `makeResetStyles` calls and keeps its classes (discovered in #25216).

See #225 for more details about `makeResetStyles`.

layershifter:
Fixes #261.
Fixes #230.

This PR modifies:
- `mergeClasses()` in Griffel core to emit errors if a call contains multiple occupancies of classes from `makeResetStyles`.
- Jest serializer to handle classes from `makeResetStyles`

Both changes uses new _global_ variable `DEBUG_RESET_CLASSES` to store classes. This works similarly to `DEFINITION_LOOKUP_TABLE`, but we need `DEBUG_RESET_CLASSES` only in development/tests so it has zero impact on production bundles.

### DIFF
--- a/change/@griffel-jest-serializer-d855a4d4-e4c6-4f70-a258-c22caa4992ea.json
+++ b/change/@griffel-jest-serializer-d855a4d4-e4c6-4f70-a258-c22caa4992ea.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "feat: add support for makeResetStyles",
+  "packageName": "@griffel/jest-serializer",
+  "email": "olfedias@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@griffel-jest-serializer-d855a4d4-e4c6-4f70-a258-c22caa4992ea.json
+++ b/change/@griffel-jest-serializer-d855a4d4-e4c6-4f70-a258-c22caa4992ea.json
@@ -1,5 +1,5 @@
 {
-  "type": "patch",
+  "type": "minor",
   "comment": "feat: add support for makeResetStyles",
   "packageName": "@griffel/jest-serializer",
   "email": "olfedias@microsoft.com",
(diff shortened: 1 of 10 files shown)


packages/core/src/makeResetStyles.ts
```diff
@@ -1,9 +1,7 @@
+import { DEBUG_RESET_CLASSES } from './constants';
 import { resolveResetStyleRules } from './runtime/resolveResetStyleRules';
 import type { GriffelResetStyle, MakeStylesOptions } from './types';
 
-/**
- * @internal
- */
```
layershifter, line 6: Oops, a leftover from the time when it was unstable.

packages/jest-serializer/src/index.ts
```diff
@@ -34,7 +40,7 @@ export function print(val: unknown) {
   /**
    * form parts of regular expression and removes collected classNames from string
    * @example
-   * regex = /r?(f16th3vw|frdkuqy0|fat0sn40|fjseox00)/
```
layershifter, line 37: We don't use `r` prefix for RTL classes for a long time.

change/@griffel-jest-serializer-d855a4d4-e4c6-4f70-a258-c22caa4992ea.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
ling1726, line 2: shouldn't this be a `minor` ?

layershifter: Yup

layershifter, line 2:
```suggestion
  "type": "minor",
```

ling1726 (approved):

bsunderhus (approved):

PR #263, opened 2022-10-26
chore: allow React 18 in peerDependencies

### LINKED ISSUE #262: Allow React 18 in peerDependencies
Fluent UI library has already merged, and I would think soonish to release support for react 18; any chance griffle will update the peer dependency sooner and make it compatible with react 18?

https://github.com/microsoft/fluentui/issues/22241

https://github.com/microsoft/griffel/blob/main/packages/react/package.json#L16

layershifter:
Fixes #262.

This PR:
- Allows React in `peerDependencies` of `@griffel/react`
- Cleanups dependencies in `@griffel/devtools` (check comments for more context)


change/@griffel-devtools-b3f344cf-70bc-43eb-af48-db8d33273092.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "fix: update dependencies in @griffel/devtools",
```
layershifter, line 3: See below ⬇️

packages/devtools/package.json
```diff
@@ -7,15 +7,7 @@
...
-    "@emotion/hash": "^0.8.0",
-    "@griffel/react": "^1.4.1",
-    "js-beautify": "^1.14.0",
-    "prism-react-renderer": "1.2.1",
-    "react-dom": ">=16.8.0 <18.0.0",
-    "react": ">=16.8.0 <18.0.0",
-    "source-map-js": "1.0.2",
-    "stylis": "^4.0.13"
+  "devDependencies": {
+    "@griffel/react": "^1.4.1"
```
layershifter, line 11: This package should not have dependencies at all as we bundle it into application. The problem is that `import/no-extraneous-dependencies` cannot be configured for that case 🗿

This config will read root `package.json` and will still complain about missing `@griffel/react`. That's why I left `@griffel/react` here in `devDependencies`.

P.S. It's also a reason why we don't need a version bump: artifacts are not affected.

miroslavstastny (approved):

PR #264, opened 2022-10-27
fix(jest-serializer): handles whitespaces better

layershifter:
#260 added support for classes generated by `makeResetStyles`, the problem is that whitespaces were still present in output:

image

This PR fixes it.

### DIFF
--- a/packages/jest-serializer/src/index.test.tsx
+++ b/packages/jest-serializer/src/index.test.tsx
@@ -1,11 +1,8 @@
 import * as React from 'react';
-import * as ReactDOM from 'react-dom';
-import * as ReactTestUtils from 'react-dom/test-utils';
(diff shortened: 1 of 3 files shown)


packages/jest-serializer/src/index.test.tsx
```diff
@@ -1,8 +1,11 @@
 import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+import * as ReactTestUtils from 'react-dom/test-utils';
 import { makeResetStyles, makeStyles, mergeClasses, TextDirectionProvider } from '@griffel/react';
 import { render } from '@testing-library/react';
 
 import { print, test } from './index';
+import { renderIntoDocument } from 'react-dom/test-utils';
```
bsunderhus, line 8: nit: unused import

bsunderhus (approved):

PR #268, opened 2022-11-01
fix: use style buckets in CSS extraction

layershifter:
This PR changes the approach for sorting CSS rules in a produced CSS asset by `@griffel/webpack-extraction-plugin`.

### Before

We used `sortCSSRules` to parse CSS rules and sort them into buckets in the similar order as rules in `resolveStyleRules()`. However, it was not **the same way**.

### After

The loader used for CSS extraction now emits comments to CSS to identify chunks and meta information (media queries case).

```css
/** @griffel:css-start [d] **/
.fe3e8s9 {
  color: red;
}
/** @griffel:css-end **/
/** @griffel:css-start [m] [{"m":"screen and (max-width: 100px)"}] **/
@media screen and (max-width: 100px) {
  .fr5o61b {
    color: red;
  }
}
/** @griffel:css-end **/
/** @griffel:css-start [m] [{"m":"screen and (max-width: 100px)"}] **/
@media screen and (max-width: 100px) {
  .f1j0ers2 {
    display: grid;
  }
}
/** @griffel:css-end **/
```

This solves the sorting problem: now we can use the same approach. It also allows us clearly identify our CSS rules in output and unblocks other approaches for the extraction plugin.

### DIFF
--- a/packages/webpack-extraction-plugin/src/parseCSSRules.ts
+++ b/packages/webpack-extraction-plugin/src/parseCSSRules.ts
@@ -15,7 +15,7 @@ export function parseCSSRules(css: string) {
   let cssBucketName: StyleBucketName | null = null;
   let cssMeta: Record<string, unknown> | null = null;
 
-  elements.forEach(element => {
+  for (const element of elements) {
     if (element.type === COMMENT) {
       if (element.value.indexOf('/** @griffel:css-start') === 0) {
         cssBucketName = element.value.charAt(24) as StyleBucketName;
@@ -24,14 +24,14 @@ export function parseCSSRules(css: string) {
           cssMeta = JSON.parse(element.value.slice(28, -5));
         }
 
-        return;
+        continue;
       }
 
       if (element.value.indexOf('/** @griffel:css-end') === 0) {
         cssBucketName = null;
         cssMeta = null;
 
-        return;
+        continue;
       }
     }
 
@@ -40,11 +40,11 @@ export function parseCSSRules(css: string) {
       const bucketEntry: CSSBucketEntry = element.type === MEDIA ? [cssRule, cssMeta!] : cssRule;
 
       cssRulesByBucket[cssBucketName].push(bucketEntry);
-      return;
+      continue;
     }
 
     unrelatedElements.push(element);
-  });
+  }
 
   return { cssRulesByBucket, remainingCSS: serialize(unrelatedElements, stringify) };
 }
--- a/packages/webpack-extraction-plugin/src/parseCSSRules.ts
+++ b/packages/webpack-extraction-plugin/src/parseCSSRules.ts
@@ -1,0 +1,50 @@
+import { styleBucketOrdering } from '@griffel/core';
+import { COMMENT, compile, MEDIA, serialize, stringify } from 'stylis';
+
+import type { CSSBucketEntry, CSSRulesByBucket, StyleBucketName } from '@griffel/core';
+
+export function parseCSSRules(css: string) {
+  const cssRulesByBucket = styleBucketOrdering.reduce<CSSRulesByBucket>((acc, styleBucketName) => {
+    acc[styleBucketName] = [];
+
+    return acc;
+  }, {}) as Required<CSSRulesByBucket>;
+  const elements = compile(css);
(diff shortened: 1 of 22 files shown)


packages/webpack-extraction-plugin/src/parseCSSRules.ts
```diff
@@ -0,0 +1,50 @@
...
+        cssBucketName = element.value.charAt(24) as StyleBucketName;
+
+        if (cssBucketName === 'm') {
+          cssMeta = JSON.parse(element.value.slice(28, -5));
+        }
+
+        return;
+      }
+
+      if (element.value.indexOf('/** @griffel:css-end') === 0) {
```
ling1726, line 30: would `for...of` be a better choice ? then you could simply `break` from the loop. It woudl be clearer than setting the bucket and meta values to `null`

packages/webpack-extraction-plugin/src/webpackLoader.ts
```diff
@@ -78,11 +75,41 @@ function webpackLoader(
...
+    if (result.cssRulesByBucket) {
+      const css = Object.entries(result.cssRulesByBucket).reduce((acc, [cssBucketName, cssBucketRules]) => {
+        if (cssBucketName === 'm') {
+          return (
+            acc +
+            cssBucketRules
+              .map(entry => {
+                return (
+                  `/** @griffel:css-start [${cssBucketName}] [${JSON.stringify(entry[1])}] **/` +
+                  '\n' +
```
ling1726, line 87: nit: would look g abit nicer if you used a string array and then joined with newline

packages/webpack-extraction-plugin/src/sortCSSRules.ts
```diff
@@ -1,99 +1,54 @@
...
+  return styleBucketOrdering
+    .map(styleBucketName => {
+      return {
+        styleBucketName,
+        cssBucketEntries: Object.values(
+          Object.fromEntries(
+            setOfCSSRules.flatMap(cssRulesByBucket => {
+              if (Array.isArray(cssRulesByBucket[styleBucketName])) {
+                return cssRulesByBucket[styleBucketName]!.map(bucketEntry => [
+                  getCSSRuleFromBucketEntry(bucketEntry),
```
ling1726, line 25: what's the point of the return of `getCSSRuleFromBucketEntry` here? You simply discard it after calling `Object.values`

packages/webpack-extraction-plugin/src/parseCSSRules.ts
layershifter: `cssMeta` & `cssBucketName` are needed for next iterations in a loop. I am not sure that `for or` will improve it anyhow as there is a need in additional iteration:

```ts
for (let i = 0, l = elements.length; i < l; i++) {
    if (elements[i].type === COMMENT) {
      if (elements[i].value.indexOf('/** @griffel:css-start') === 0) {
        const cssBucketName = elements[i].value.charAt(24) as StyleBucketName;
        const cssMeta: Record<string, unknown> | null =
          cssBucketName === 'm' ? JSON.parse(elements[i].value.slice(28, -5)) : null;

        for (i++; i < l; i++) {
          if (elements[i].value.indexOf('/** @griffel:css-end') === 0) {
            break;
          }

          const cssRule = serialize([elements[i]], stringify);
          const bucketEntry: CSSBucketEntry = cssMeta ? [cssRule, cssMeta] : cssRule;

          cssRulesByBucket[cssBucketName].push(bucketEntry);
        }
      }

      continue;
    }

    unrelatedElements.push(elements[i]);
  }
```

And IMO it's more ugly. Anyway I changed to `for of`.

packages/webpack-extraction-plugin/src/webpackLoader.ts
layershifter: Agree, pushed

packages/webpack-extraction-plugin/src/sortCSSRules.ts
ling1726: Makes sense, could you please add some documentation? It's really not obvious what the intention is there

layershifter: Add a comment

PR #277, opened 2022-11-23
chore: rework CSS extraction

layershifter:
This PR refactors Webpack plugin for CSS extraction.

## Before

We use the same approach as [CompiledCSS](https://github.com/atlassian-labs/compiled/blob/40211567504538272f726085e7b5f8c7c4eb2255/packages/webpack-loader/src/extract-plugin.ts#L38-L69): we create a new `cacheGroup` in `optimization.splitChunks` and force all generated CSS files by us to go there.

As a result we will get a single asset (CSS file) that we can process and sort rules. This works, but this approach requires `optimization.splitChunks` to be **enabled**. One of our key partners has this setting **disabled** in dev environment: all try outs to enable it were not successful.

This approach also prevents all attempts to implement chunks splitting support and avoid a single CSS file.

## After

TL;DR Does the same, works differently ¯\_(ツ)_/¯

The key difference is that we don't use `optimization.splitChunks` and moving CSS files to a target chunk.

### Initial attempt

Inspired by https://github.com/atlassian-labs/compiled/pull/724. All processing was done during [`.processAssets` stage](https://webpack.js.org/api/compilation-hooks/#processassets) and it worked: we got a CSS file per a chunk (if it contained `makeStyles` calls) and I was able to extract rules from it to a dedicated chunk.

**But:**
- If a chunk contained **only** Griffel's CSS after processing we got an empty file, for example:
   ```
   - griffel.css (contains all CSS)
   - chunk.css (became empty file)
   - async-chunk.css (became empty file)
   ```
- These empty files **could not be removed** (via [`compilation.deleteAsset()`](https://webpack.js.org/api/compilation-object/#deleteasset)) as `mini-css-extract-plugin` [will register](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/26334462e419026086856787d672b052cd916c62/src/index.js#L943-L945) async chunks in JS runtime and these assets will fail to load.
- Whatever modifications to assets we will do - we should do them before [`.runtimeRequirementInTree`](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/26334462e419026086856787d672b052cd916c62/src/index.js#L1071-L1076) (undocumented Webpack hook) otherwise chunks that have empty CSS could be registered in JS runtime 💣 

### Implementation in this PR

After multiple attempts to do something with assets I realized that [`.processAssets`](https://webpack.js.org/api/compilation-hooks/#processassets) is **too late** to transfer CSS between chunks. The problem that at any stage before `.runtimeRequirementInTree` we don't have assets - we operate with modules ([not really well documented](https://webpack.js.org/concepts/under-the-hood/)). _Simple explanation: an asset is a concatenated set of modules._

On `.afterChunks` (again undocumented, happens after a chunk graph gets **generated** and modules **were collected**):
- we attach a custom chunk to main entrypoint
- we move all modules produced by `mini-css-extract-plugin` to that chunk (if they contain CSS produced by Griffel)

On `.processAssets` we are in the same condition actually as currently: all CSS modules produced by Griffel were moved to a dedicated chunk -> this chunk has a single CSS file -> we sort rules in it ✅

### DIFF
--- a/e2e/nextjs/project.json
+++ b/e2e/nextjs/project.json
@@ -2,12 +2,13 @@
   "root": "e2e/nextjs",
   "sourceRoot": "e2e/nextjs/src",
   "projectType": "library",
+  "implicitDependencies": ["@griffel/webpack-loader"],
   "targets": {
     "test": {
       "executor": "@nrwl/workspace:run-commands",
+      "dependsOn": [{ "target": "build", "projects": "dependencies" }],
       "options": {
         "cwd": "e2e/nextjs",
-        "dependsOn": [{ "target": "build", "projects": "dependencies" }],
         "commands": [{ "command": "swc-node src/test.ts" }],
         "outputPath": []
       }
--- a/packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
+++ b/packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
@@ -63,7 +63,7 @@ function ensureModuleHasPostOrderIndex(griffelChunk: Chunk, cssModule: Module) {
       continue;
     }
 
-    // "mini-css-extract" requires to an index on modules, modules without indexes will be filtered out and throw
+    // "mini-css-extract" requires an index to exist on modules. A module with an index will be filtered out and the plugin will throw
     // https://github.com/webpack-contrib/mini-css-extract-plugin/blob/26334462e419026086856787d672b052cd916c62/src/index.js#L1133-L1140
     group.setModulePostOrderIndex(
       cssModule,
--- a/packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
+++ b/packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
@@ -1,29 +1,35 @@
 import { defaultCompareMediaQueries, GriffelRenderer } from '@griffel/core';
 import { Compilation } from 'webpack';
-import type { Compiler, sources } from 'webpack';
+import type { Chunk, Compiler, Module, sources } from 'webpack';
 
-import { sortCSSRules } from './sortCSSRules';
 import { parseCSSRules } from './parseCSSRules';
+import { sortCSSRules } from './sortCSSRules';
 
 export type GriffelCSSExtractionPluginOptions = {
   compareMediaQueries?: GriffelRenderer['compareMediaQueries'];
 };
 
-/**
- * Forces all files with `griffel.css` be concatenated into a single asset.
- */
-function forceCSSIntoOneStyleSheet(compiler: Compiler) {
-  compiler.options.optimization ??= {};
-  compiler.options.optimization.splitChunks ??= {};
-  compiler.options.optimization.splitChunks.cacheGroups ??= {};
+type IterableElement<TargetIterable> = TargetIterable extends Iterable<infer ElementType> ? ElementType : never;
+type ChunkGroup = IterableElement<Chunk['groupsIterable']>;
 
-  compiler.options.optimization.splitChunks.cacheGroups['griffel'] = {
-    name: `griffel`,
-    type: 'css/mini-extract',
-    chunks: 'all',
-    test: /griffel.css/,
-    enforce: true,
-  };
+const PLUGIN_NAME = 'GriffelExtractPlugin';
+
+function attachGriffelChunkToMainEntryPoint(compilation: Compilation, griffelChunk: Chunk) {
+  const entryPoints = Array.from(compilation.entrypoints.values());
+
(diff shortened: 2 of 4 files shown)


e2e/nextjs/project.json
```diff
@@ -2,12 +2,13 @@
   "root": "e2e/nextjs",
   "sourceRoot": "e2e/nextjs/src",
   "projectType": "library",
+  "implicitDependencies": ["@griffel/webpack-loader"],
```
layershifter, line 5: As we don't `import`/`require` `@griffel/webpack-loader` anywhere - Nx fails to detect it as a dependency.

e2e/nextjs/project.json
```diff
@@ -2,12 +2,13 @@
...
   "targets": {
     "test": {
       "executor": "@nrwl/workspace:run-commands",
+      "dependsOn": [{ "target": "build", "projects": "dependencies" }],
```
layershifter, line 9: Have been in a wrong spot 🤦

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -1,29 +1,35 @@
...
+const PLUGIN_NAME = 'GriffelExtractPlugin';
+
+function attachGriffelChunkToMainEntryPoint(compilation: Compilation, griffelChunk: Chunk) {
+  const entryPoints = Array.from(compilation.entrypoints.values());
+
+  if (entryPoints.length === 0) {
+    throw new Error('Failed to find and entry points in "compilation.entrypoints"');
+  }
+
+  const mainEntryPoint = entryPoints[0];
```
ling1726, line 24: is the main entrypoint guaranteed to be the first entrypoint in `compilation.entrypoints` ?

ling1726: actually, what if webpack is configured with multiple entrypoints?

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -36,6 +42,63 @@ function getAssetSourceContents(assetSource: sources.Source): string {
...
+  return false;
+}
+
+function ensureModuleHasPostOrderIndex(griffelChunk: Chunk, cssModule: Module) {
+  for (const group of griffelChunk.groupsIterable) {
+    if (group.getModulePostOrderIndex(cssModule)) {
+      continue;
+    }
+
+    // "mini-css-extract" requires to an index on modules, modules without indexes will be filtered out and throw
```
ling1726, line 66:
```suggestion
    // "mini-css-extract" requires an index to exist on modules. A module with an index will be filtered out and the plugin will throw
```

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -36,6 +42,63 @@ function getAssetSourceContents(assetSource: sources.Source): string {
...
+    // https://github.com/webpack-contrib/mini-css-extract-plugin/blob/26334462e419026086856787d672b052cd916c62/src/index.js#L1133-L1140
+    group.setModulePostOrderIndex(
+      cssModule,
+      // It's bad to use private APIs, but it's more reliable than just random indexes
```
ling1726, line 70:
```suggestion
      // It's bad to use private APIs, but it's more reliable than random indexes
```

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -46,27 +109,35 @@ export class GriffelCSSExtractionPlugin {
...
           stage: Compilation.PROCESS_ASSETS_STAGE_PRE_PROCESS,
         },
         assets => {
-          const griffelAsset = Object.entries(assets).find(([assetName]) => assetName.includes('griffel'));
+          const cssAssetDetails = Object.entries(assets).find(([assetName]) => griffelChunk.files.has(assetName));
 
-          if (!griffelAsset) {
-            return;
+          if (typeof cssAssetDetails === 'undefined') {
+            throw new Error('Failed to an asset that contains Griffel CSS output');
```
ling1726, line 130:
```suggestion
            throw new Error('Failed to find an asset that contains Griffel CSS output');
```

ling1726 (approved):

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
layershifter: It's a good question. In all scenarios that I currently tested attaching to main entrypoint was enough.

PR #279, opened 2022-11-23
test: create E2E test for Next.js and CSS extraction

layershifter:
This PR adds E2E test for Next.js plugin to ensure that CSS extraction works as expected.

- Some utils were moved out from TS into `@griffel/e2e-utils` to share them between tests
- Tests and utils were rewritten to TS for better expirence
- `swc-node` was added to run TS code
- `log-symbols` was downgraded as it requires ESM, but `.mts` is not supported by SWC https://github.com/swc-project/swc/issues/4679


e2e/typescript/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "allowJs": true,
     "checkJs": true,
     "module": "CommonJS",
+    "lib": ["ES2019", "DOM", "DOM.Iterable"],
```
layershifter, line 7: To use `Object.fromEntries`. It seems that we just need to enable `ES2019` everywhere. I will do it in a separate PR.

tsconfig.base.json
```diff
@@ -18,7 +18,7 @@
       "@griffel/babel-preset": ["packages/babel-preset/src/index.ts"],
       "@griffel/core": ["packages/core/src/index.ts"],
       "@griffel/devtools": ["packages/devtools/src/index.ts"],
-      "@griffel/e2e-typescript": ["e2e/typescript/src/index.ts"],
```
layershifter, line 21: These is no sense to have `@griffel/e2e-typescript` there as it exports nothing.

e2e/nextjs/src/assets/next.config.js
```diff
@@ -0,0 +1,33 @@
+const { withGriffelCSSExtraction } = require('@griffel/next-extraction-plugin');
```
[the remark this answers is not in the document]
ling1726: you could just add that as a comment to the file?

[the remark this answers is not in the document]
ling1726: or README

e2e/utils/src/installPackages.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  if (packages.length > 0) {
+    const yarnOutput = await sh(`yarn info ${packages.join(' ')} --json`, rootDir, true);
+    const parsedYarnOutput = yarnOutput
+      .split('\n')
+      .filter(Boolean)
+      .map(line => JSON.parse(line));
+
+    packagesWithVersions = Object.fromEntries(
+      parsedYarnOutput.map(info => [info.value.split('@patch')[0].split('@npm')[0], info.children.Version]),
```
ling1726, line 32: what kind of package would have `@patch`?

ling1726 (approved):

e2e/nextjs/src/assets/next.config.js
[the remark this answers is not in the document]
layershifter: Added

e2e/utils/src/installPackages.ts
layershifter: image

Yarn applies a patch to TS.

layershifter: I will probably cover some of these utils with tests in upcoming PRs.

PR #281, opened 2022-11-25
chore: bump monosize

layershifter:


.github/workflows/bundle-size.yml
```diff
@@ -18,7 +18,7 @@ jobs:
 
       - uses: actions/setup-node@v1
         with:
-          node-version: '14'
+          node-version: '18'
```
layershifter, line 21: It's intentional as Node 18 contains native fetch. fetch is required for that storage adapter.

ling1726 (approved):

PR #282, opened 2022-11-29
Hook naming lint rule

### LINKED ISSUE #213: eslint-plugin: Add rule to check that makeStyles' returned hook is named use*
To keep with rules of hooks, it would be nice to have a rule to make sure `makeStyles` is always used with a variable whose name starts with "use".

Valid:
```js
const useStyles = makeStyles(...);
```

Invalid:
```js
const styles = makeStyles(...);
```

We've had some devs who don't think `useStyles` is a hook, since they're not used to the factory pattern. A linter rule would help enforce and educate them.

NotWoods:
Fixes #213

Introduce a new linter rule to enforce the React hook naming pattern

### DIFF
--- a/packages/eslint-plugin/src/rules/hook-naming.ts
+++ b/packages/eslint-plugin/src/rules/hook-naming.ts
@@ -1,0 +1,42 @@
+import { ESLintUtils } from '@typescript-eslint/utils';
+
+import { createRule } from '../utils/createRule';
+import { isIdentifier, isMakeStylesIdentifier } from '../utils/helpers';
+
+export const RULE_NAME = 'hook-naming';
+
+export const hookNamingRule: ReturnType<ReturnType<typeof ESLintUtils.RuleCreator>> = createRule({
+  name: RULE_NAME,
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Enforce that hooks returned from makeStyles() calls are named with a "use" prefix',
+      recommended: 'error',
+    },
+    messages: {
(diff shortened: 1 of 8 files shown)


packages/eslint-plugin/src/rules/hook-naming.ts
```diff
@@ -0,0 +1,42 @@
...
+export const hookNamingRule: ReturnType<ReturnType<typeof ESLintUtils.RuleCreator>> = createRule({
+  name: RULE_NAME,
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'Enforce that hooks returned from makeStyles() calls are named with a "use" prefix',
+      recommended: 'error',
+    },
+    messages: {
+      invalidMakeStylesHookNameFound: 'Hooks must start with the prefix "use"',
```
ling1726, line 17:
```suggestion
      invalidMakeStylesHookNameFound: '`makeStyles` returns a hook. Hooks must start with the prefix `use`',
```

nit: let's be explicit this lint rule is supposed to target `makeStyles` calls

ling1726 (approved):
This is a good improvement. I think there's are two edge cases regarding the import that aren't handled

```ts
// makeStyles import is aliased
import { makeStyles as getStyles } from '@griffel/react';
// something else is aliased as makeStyles
import { something as makeStyles } from 'other-library';
```

I think the first example would be more realistic to handle than the second. However I see that the `no-shorthands` has a similar problem so I won't block the PR for it, but it might be nice to have an issue to address this for both rules in the future

layershifter:
Released in `@griffel/eslint-plugin@1.1.0` 🎉

PR #283, opened 2022-11-30
fix: don't throw in the plugin

layershifter:


e2e/nextjs/src/test.ts
```diff
@@ -22,6 +22,7 @@ async function performTest() {
     const resolutions = await Promise.all([
       packLocalPackage(rootDir, tempDir, '@griffel/core'),
       packLocalPackage(rootDir, tempDir, '@griffel/react'),
+      packLocalPackage(rootDir, tempDir, '@griffel/webpack-extraction-plugin'),
```
layershifter, line 25: This explains how CI passed before. Module was installed from NPM 🤦

ling1726 (approved):

PR #285, opened 2022-12-01
chore: bump loader-utils, schema-utils and beachball

layershifter:
- Removes `loader-utils` & `schema-utils` from dependencies
- Bumps loader-utils in `yarn.lock` resolve security alerts
- Bumps beachball to resolve security alerts

### DIFF
--- a/.yarn/patches/beachball-npm-2.31.5-0e84ec4233.patch
+++ b/.yarn/patches/beachball-npm-2.31.5-0e84ec4233.patch
@@ -1,5 +1,5 @@
 diff --git a/lib/packageManager/packagePublish.js b/lib/packageManager/packagePublish.js
-index 6848f4d9fa070cd5262de8dc19a0a305b0dc4b6d..a016d229f92f567f43b2cc3d403c38725ceb1d70 100644
+index 6a211b2b472c8d16f776f79a89d5914cabd390b7..f0006dd42e1f8aae581e61c6c5b6dfe495500ed1 100644
 --- a/lib/packageManager/packagePublish.js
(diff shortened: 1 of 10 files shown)


.yarn/patches/beachball-npm-2.31.5-0e84ec4233.patch
```diff
@@ -19,9 +19,8 @@ index 6848f4d9fa070cd5262de8dc19a0a305b0dc4b6d..a016d229f92f567f43b2cc3d403c3872
      }
 +    console.log(`publish cwd: ${artifactsPath}`);
      console.log(`publish command: ${args.join(' ')}`);
--    return npm_1.npm(args, { cwd: packagePath, timeout });
-+    return npm_1.npm(args, { cwd: artifactsPath, timeout });
+-    return npm_1.npmAsync(args, { cwd: packagePath, timeout, all: true });
++    return npm_1.npmAsync(args, { cwd: artifactsPath, timeout, all: true });
```
layershifter, line 23: Patch is the same as in `monosize`, it works (confirmed by successful releases)

packages/webpack-extraction-plugin/src/webpackLoader.ts
```diff
@@ -46,13 +43,6 @@ function webpackLoader(
   // https://github.com/webpack/webpack/issues/14946
   this.cacheable();
 
-  const options = getOptions(this) as WebpackLoaderOptions;
```
layershifter, line 49: We don't use options, so we don't need to validate them.

ling1726 (approved):

PR #289, opened 2022-12-06
fix: avoid empty CSS imports

### LINKED ISSUE #266: extraction-plugin: support for appDir in NextJS 13
Hi,
I was trying to use  NextJS v13 appDir mode with `@griffel` and I get the following errors:

```bash
Attempted import error: '__css' is not exported from '@griffel/react' (imported as '__css').
```

and

```bash
TypeError: core.createDOMRenderer is not a function
    at 4415 (/home/projects/nextjs-ltekvi/.next/server/chunks/792.js:3446:88)
    at __webpack_require__ (/home/projects/nextjs-ltekvi/.next/server/webpack-runtime.js:25:43)
    at 5933 (/home/projects/nextjs-ltekvi/.next/server/chunks/792.js:3829:23)
    at __webpack_require__ (/home/projects/nextjs-ltekvi/.next/server/webpack-runtime.js:25:43)
    at 9692 (/home/projects/nextjs-ltekvi/.next/server/chunks/792.js:3710:18)
    at __webpack_require__ (/home/projects/nextjs-ltekvi/.next/server/webpack-runtime.js:25:43)
    at 4312 (/home/projects/nextjs-ltekvi/.next/server/app/page.js:99:72)
    at __webpack_require__ (/home/projects/nextjs-ltekvi/.next/server/webpack-runtime.js:25:43)
    at eval (/home/projects/nextjs-ltekvi/node_modules/next/dist/compiled/react-server-dom-webpack/client.js:918:337)
```

There are errors only when using the '@griffel/next-extraction-plugin' package. 

To reproduce these errors g to the [Sandbox](https://stackblitz.com/edit/nextjs-ltekvi) and run the command:
```bash 
npm install && npx next build
```

layershifter:
Fixes #266. More details in https://github.com/microsoft/griffel/issues/266#issuecomment-1340962907.

Our Babel plugin that strips runtime also returns a set of CSS rules in its metadata. Before this PR it always returned at least an empty object. Because of this the check below didn't work:

https://github.com/microsoft/griffel/blob/1289343d950319f48888cfe70c1c84af066f08db/packages/webpack-extraction-plugin/src/webpackLoader.ts#L68

As the check was passing always we were creating invalid imports for files that contain `__styles` as a string. [Files like that](https://github.com/microsoft/fluentui/blob/f3e6e65b7f6435dc6b09bfd626e99921a10983d8/packages/merge-styles/src/Stylesheet.ts#L90) were processed => got empty set of CSS rules => created an invalid import => crashed build.

### DIFF
--- a/packages/webpack-extraction-plugin/__fixtures__/webpack/missing-calls/output.ts
+++ b/packages/webpack-extraction-plugin/__fixtures__/webpack/missing-calls/output.ts
@@ -1,0 +1,1 @@
+export const foo = '__styles';
--- a/change/@griffel-webpack-extraction-plugin-46602e0c-3b1b-4553-a919-32df4dda9fd9.json
+++ b/change/@griffel-webpack-extraction-plugin-46602e0c-3b1b-4553-a919-32df4dda9fd9.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix: avoid empty CSS imports",
+  "packageName": "@griffel/webpack-extraction-plugin",
+  "email": "olfedias@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/webpack-extraction-plugin/__fixtures__/webpack/missing-calls/code.ts
+++ b/packages/webpack-extraction-plugin/__fixtures__/webpack/missing-calls/code.ts
@@ -1,0 +1,1 @@
+export const foo = '__styles';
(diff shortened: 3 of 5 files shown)


packages/webpack-extraction-plugin/__fixtures__/webpack/missing-calls/output.ts
```diff
@@ -0,0 +1 @@
+export const foo = '__styles';
```
ling1726, line 1: wouldn't it make more sense to have better regex ? like `__styles()`? 🤔

layershifter: Yes, but this loader targets `node_modules` and needs to process thousands files in real projects, so if will put a regex in the hot path I believe that it will be a bottleneck.

ling1726: perhaps some kind of sliding window algorithm might be able to do the same thing while still being faster than Regex, especially you know how the call starts and terminates 🤔

but I guess handling empty buckets is a better gate

ling1726 (approved):

PR #291, opened 2022-12-07
chore: remove path magic in extraction process

layershifter:
With changes in this PR we don't use this file anymore. The recipe is described in [Webpack's documentation](https://webpack.js.org/api/loaders/#inline-matchresource). It was impossible to do it before #277, but now we don't rely on `cacheGroups` anymore.

## Before

CSS was was emitted to a fake file `griffel.css` in the virtual loader, this file has a path that might look like `node_modules/@griffel/webpack-extraction-plugin/virtual-loader/griffel.css`. Image files are relative to the userland code
file, so we had to construct a relative path from the fake CSS file to the image file in userland code.

## After

The fake `griffel.css` file is is under the same path as the original code file, so there is no need to do any changes to the image path since it will already be relative to the original code file.

### DIFF
--- a/packages/webpack-extraction-plugin/src/webpackLoader.ts
+++ b/packages/webpack-extraction-plugin/src/webpackLoader.ts
@@ -8,10 +8,7 @@ type WebpackLoaderOptions = never;
 
 type WebpackLoaderParams = Parameters<webpack.LoaderDefinitionFunction<WebpackLoaderOptions>>;
 
-const resourceDirectory = path.resolve(__dirname, '..', 'virtual-loader');
-
-const virtualLoaderPath = path.resolve(resourceDirectory, 'index.js');
-const resourcePath = path.resolve(resourceDirectory, 'griffel.css');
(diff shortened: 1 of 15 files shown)


packages/webpack-extraction-plugin/src/webpackLoader.ts
```diff
@@ -94,14 +90,19 @@ function webpackLoader(
...
+
+      const request = `${outputFileName}!=!${virtualLoaderPath}?style=${toURIComponent(css)}!${this.resourcePath}`;
+      const stringifiedRequest = JSON.stringify(this.utils.contextify(this.context || this.rootContext, request));
 
-      this.callback(null, `${result.code}\n\n${request};`, result.sourceMap);
+      this.callback(
+        null,
+        // Heads up!
+        // This is probably a bug, but "import" does not work properly there: files from node_modules are processed, but
+        // they will not be present in chunks 😳
```
layershifter, line 102: This costs me 5 hours of debugging in product, will create an issue in Webpack as it's clearly bug in Webpack or in css-loader/mini-css-extract.

ling1726 (approved):

PR #296, opened 2022-12-20
fix: avoid mutations in getStyleSheetForBucket

### LINKED ISSUE #295: Style element attributes are shared and mutated in renderer
https://codesandbox.io/s/nostalgic-jepsen-qui5et?file=/src/App.js

If you inspect the above example you will see that **all style elements** have the media selector. This is because the user provided `styleElementAttributes` are reused and mutated

https://github.com/microsoft/griffel/blob/6176db7b43ca197e1136657f65003240f120f0c8/packages/core/src/renderer/createDOMRenderer.ts#L66

https://github.com/microsoft/griffel/blob/6176db7b43ca197e1136657f65003240f120f0c8/packages/core/src/renderer/getStyleSheetForBucket.ts#L63

(only bucket `m` should have the `media` attribute)
image

layershifter:
Fixes #295.

### DIFF
--- a/change/@griffel-core-e59b2cbc-4a06-4803-ac9b-c6ed34521758.json
+++ b/change/@griffel-core-e59b2cbc-4a06-4803-ac9b-c6ed34521758.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix: avoid mutations in getStyleSheetForBucket",
+  "packageName": "@griffel/core",
+  "email": "olfedias@microsoft.com",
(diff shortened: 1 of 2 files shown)


ling1726:
Please do separate PRs or name this PR a `feat` since we're exposing smth new in public API

layershifter:
> Please do separate PRs or name this PR a `feat` since we're exposing smth new in public API

Done 👍

ling1726 (approved):

PR #308, opened 2023-01-06
chore: adjust error messaging for rules which don't include ampersands

dzearing:
With this rule:

```
rule: {
  '.my-class': {
    ...
  }
}
```

Before:

Would fail to register and emit a `console.error` message indicating something was wrong.

After

Same behavior, but the message indicates an ampersand is missing in '.my-class'.

### DIFF
--- a/change/@griffel-core-d096f264-8c73-426c-b236-bcbf30ad8f49.json
+++ b/change/@griffel-core-d096f264-8c73-426c-b236-bcbf30ad8f49.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Adjusting console logging to emit a more descriptive error message.",
+  "packageName": "@griffel/core",
+  "email": "dzearing@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@griffel-core-d096f264-8c73-426c-b236-bcbf30ad8f49.json
+++ b/change/@griffel-core-d096f264-8c73-426c-b236-bcbf30ad8f49.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Adjusting console logging to emit a more descriptive error message.",
+  "comment": "chore: adjust console logging to emit a more descriptive error message",
   "packageName": "@griffel/core",
   "email": "dzearing@microsoft.com",
   "dependentChangeType": "patch"
--- a/packages/core/src/runtime/resolveResetStyleRules.ts
+++ b/packages/core/src/runtime/resolveResetStyleRules.ts
@@ -119,8 +119,19 @@ function createStringFromStyles(styles: GriffelResetStyle) {
     }
 
     if (process.env.NODE_ENV !== 'production') {
-      // eslint-disable-next-line no-console
-      console.error(`Please fix the unresolved style rule: \n ${property} \n ${JSON.stringify(value, null, 2)}"`);
+      if (property.indexOf('&') === -1) {
+        // eslint-disable-next-line no-console
+        console.error(
+          `Please fix the unresolved style rule (it may be missing an ampersand placeholder where the generated class name should be injected): \n ${property} \n ${JSON.stringify(
+            value,
+            null,
+            2,
+          )}"`,
+        );
+      } else {
+        // eslint-disable-next-line no-console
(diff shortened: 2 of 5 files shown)


change/@griffel-core-d096f264-8c73-426c-b236-bcbf30ad8f49.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Adjusting console logging to emit a more descriptive error message.",
```
layershifter, line 3:
```suggestion
  "comment": "chore: adjust console logging to emit a more descriptive error message",
```

packages/core/src/runtime/resolveResetStyleRules.ts
```diff
@@ -119,8 +119,17 @@ function createStringFromStyles(styles: GriffelResetStyle) {
...
+      if (property.indexOf('&') === -1) {
+        // eslint-disable-next-line no-console
+        console.error(
+          `Please fix the unresolved style rule (it is missing an ampersand placeholder where the generated class name should be injected): \n ${property} \n ${ruleText}`,
+        );
+      } else {
+        // eslint-disable-next-line no-console
+        console.error(`Please fix the unresolved style rule: \n ${property} \n ${ruleText}`);
+      }
     }
```
layershifter, line 133:
```suggestion
    if (process.env.NODE_ENV !== 'production') {
      // Oops, a rule was not resolved
      warnAboutUnresolvedRule(property, value);
    }
```

layershifter (commented):
I tried to improve the message even more to make it more actionable:

image

As we use the same error in two places, another in `resolveStyleRules.ts`:

https://github.com/microsoft/griffel/blob/eb26eafe951623518da5a6fecf8142bab00d85b8/packages/core/src/runtime/resolveStyleRules.ts#L322-L325

I think that it's better to implement a single function to do that.

#### `packages/core/src/runtime/warnAboutUnresolvedRule.ts`

```js
import type { GriffelResetStyle, GriffelStyle } from '@griffel/core';

export function warnAboutUnresolvedRule(property: string, value: GriffelStyle | GriffelResetStyle) {
  const ruleText = JSON.stringify(value, null, 2);
  const message: string[] = [
    '@griffel/react: A rule was not resolved to CSS properly. ' +
      'Please check your `makeStyles` or `makeResetStyles` calls for following:',
    ' '.repeat(2) + 'makeStyles({',
    ' '.repeat(4) + `[slot]: {`,
    ' '.repeat(6) +
      `"${property}": ${ruleText
        .split('\n')
        .map((l, n) => ' '.repeat(n === 0 ? 0 : 6) + l)
        .join('\n')}`,
    ' '.repeat(4) + '}',
    ' '.repeat(2) + `})`,
    '',
  ];

  if (property.indexOf('&') === -1) {
    message.push(
      `It looks that you're are using a nested selector, but it is missing an ampersand placeholder where the generated class name should be injected.`,
    );
    message.push(`Try to update a property to include it i.e "${property}" => "&${property}".`);
  } else {
    message.push('');
    message.push(
      "If it's not obvious what triggers a problem, please report an issue at https://github.com/microsoft/griffel/issues",
    );
  }

  // eslint-disable-next-line no-console
  console.error(message.join('\n'));
}
```

#### `packages/core/src/runtime/warnAboutUnresolvedRule.test.ts`

```js
import { warnAboutUnresolvedRule } from './warnAboutUnresolvedRule';

describe('warnAboutUnresolvedRule', () => {
  it('warns on a missing "&"', () => {
    const spy = jest.spyOn(console, 'error').mockImplementation();

    warnAboutUnresolvedRule('div', {
      color: 'red',
      ':hover': { color: 'blue' },
    });

    expect(spy.mock.calls[0][0]).toMatchInlineSnapshot(`
      "@griffel/react: A rule was not resolved to CSS properly. Please check your \`makeStyles\` or \`makeResetStyles\` calls for following:
        makeStyles({
          [slot]: {
            \\"div\\": {
              \\"color\\": \\"red\\",
              \\":hover\\": {
                \\"color\\": \\"blue\\"
              }
            }
          }
        })

      It looks that you're are using a nested selector, but it is missing an ampersand placeholder where the generated class name should be injected.
      Try to update a property to include it i.e \\"div\\" => \\"&div\\"."
    `);
  });
});
```

@dzearing WDYT?

layershifter (approved):

Issue #33, opened 2022-02-16
vite: Add plugin for Vite


zoontek:
Hello 👋 

This project seems very promising! Do you have plan to support Vite with a custom plugin?

Documentation: https://vitejs.dev/guide/api-plugin.html
`vanilla-extract` plugin (for reference): https://github.com/seek-oss/vanilla-extract/tree/master/packages/vite-plugin

layershifter:
Hey, yes we plan to have more integrations with bundlers and frameworks. Having Vite plugin is something that we want.

The implementation should be similar to the existing Webpack loader: call Babel and reuse bundler's module resolution. I haven't checked Vite docs yet and don't know what APIs are exposed, but it should it be doable.

layershifter:
The closest example/reference for us is [`@linaria/rollup`](https://github.com/callstack/linaria/blob/master/packages/rollup/src/index.ts), but it has the problem that I mentioned (https://github.com/callstack/linaria/issues/888). In fact, we need something similar in Rollup/Vite:

https://github.com/microsoft/griffel/blob/b4c7acc77e5e7b4883ca32249418a3d93dc43e77/packages/webpack-loader/src/webpackLoader.ts#L69-L80

layershifter:
Okay, it seems that [`this.resolve()`](https://rollupjs.org/guide/en/#thisresolve) in Rollup is close to what we need 😸

zoontek:
I will check when I have some time 🙂

layershifter:
> I will check when I have some time 🙂

@zoontek thanks. FYI: I will cleanup my branch and will post as a draft PR on the next week.

SahilTara:
Just wanted to bump on this issue since the linaria pr linked seems to have been merged.

layershifter:
@SahilTara indeed, but we will need to move to Linaria v4 first (#184).

layershifter:
> Any update on the vite plugin?

As [wyw-in-js](https://github.com/Anber/wyw-in-js) is out, I can rollout the plugin finally 🎉 

I reworked original PR (#41) and it can be shipped once the issue will be fixed (https://github.com/Anber/wyw-in-js/issues/27). As @Anber will be offline for some time, I expect that we will be able to ship it after NY.

layershifter:
FYI #41 was merged and plugin will be available on NPM today. It does not support CSS extraction, yet (#494).

PR #309, opened 2023-01-07
chore: upgrade to Linaria v4

### LINKED ISSUE #104: babel-plugin: multiple modules produce the same import
#103 fixed handling of multiple exports of the same module, but there is an edge case with multiple modules:

```ts
import { makeStyles as makeStylesA } from 'module-a'
import { makeStyles as makeStylesB } from 'module-b'

// ⬇️⬇️⬇️

import { __styles } from 'module-a'
import { __styles } from 'module-b'
```

This will explode as `__styles` is duplicated identifier. We should use local imports for this case, i.e. result should be:

```tsx
import { __styles } from 'module-a'
import { __styles as __stylesB } from 'module-b'
```

#### References

- https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#scope
- https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#generating-a-uid

### LINKED ISSUE #184: babel-preset/webpack-loader: use wyw-in-js
Linearia v4 has been released, there will be no v3 fixes going forward.

### LINKED ISSUE #313: babel-preset: handles relative paths as assets
In the fixture below `assetWithHash` contains a `url()` that uses relative path (not imported path).

https://github.com/microsoft/griffel/blob/eb26eafe951623518da5a6fecf8142bab00d85b8/packages/babel-preset/__fixtures__/assets/code.ts#L7-L12

⬇️⬇️⬇️

https://github.com/microsoft/griffel/blob/eb26eafe951623518da5a6fecf8142bab00d85b8/packages/babel-preset/__fixtures__/assets/output.ts#L27

`url()` there should not be transformed there, but it gets 💥
It seems that without migration to Linaria v4 (#184) that uses absolute paths it will be impossible to handle this scenario properly.

layershifter:
## Summary

This PR upgrades `@griffel/babel-preset` and `@griffel/webpack-loader` to use Linaria v4. Linaria v4 introduced a concept of processors to handle different CSS-in-JS implementations i.e. "implement your processor (transform) and use existing tooling to run it".

The information is obtainer from `package.json` files:

```json
{
  "linaria": {
    "tags": {
      "makeStyles": "@griffel/linaria-processor/make-styles",
      "makeResetStyles": "@griffel/linaria-processor/make-reset-styles"
    }
  }
}
```

👆 _indicates Linaria to process `makeStyles` & `makeResetStyles` with matching transforms_

Another benefit is that Linaria v4 supports async resolving of dependencies that was a blocker for other bundlers (#33).

# BREAKING CHANGES

### `@griffel/babel-preset`

- `modules` & `projectRoot` options removed from configs, use `package.json` to define processors (see Summary ⬆️ )
- `EvalCache` & `Module` are not longer exported and should not be used directly anymore
- requires `@griffel/linaria-processor` installed

### `@griffel/webpack-loader`

- `modules` & `projectRoot` options removed from configs, use `package.json` to define processors (see Summary ⬆️ )
- requires `@griffel/linaria-processor` installed

### `@griffel/babel-preset`

All internals of `@griffel/babel-preset` were removed: module evaluation is gone, parts related to assets handling were moved to `@griffel/linaria-processor`.

- `error-argument-count`: not needed, check is not longer there
- `import-custom-module`: aliases are configured now in `package.json` files
- `import-custom-name`: aliases are configured now in `package.json` files
- `non-existing-module-call`: removed, `@linaria/shaker` now works differently
- `require-custom-module`: removed, not needed

Updated fixtures:
- `asset`: file paths handled properly, fixes #313
- `config-evaluation-rules`: we don't have own evaluation part anymore, matches to what Linaria does
- `duplicated-imports`: handles multiple imports as it should be 💪 Fixes #104

### `@griffel/webpack-loader`

All internals of `@griffel/webpack-loader` were removed, now it's just a tiny wrapper around `@linaria/webpack-loader`.

Following fixtures were removed:
- `config-modules`: aliases are configured now in `package.json` files
- `error-argument-count`: not needed, check is not longer there
- `error-syntax`: not needed

New fixture `react-component` to test a more real life example.

#### Related issues

Fixes #104.
Fixes #184.
Fixes #313.

#### TODOs

- [x] Test on Windows machine
- [ ] Test in products
- [x] Pass proper import sources in processors
- [x] ~~🔴  **BLOCKER** https://github.com/callstack/linaria/issues/1214~~
- [x] ~~🔴  **BLOCKER** https://github.com/callstack/linaria/issues/1209~~
- [x] ~~🔴 **BLOCKER:** https://github.com/callstack/linaria/issues/1186~~
- [x] ~~🔴 **BLOCKER:** https://github.com/callstack/linaria/pull/1188~~
- [x] ~~🔴 **BLOCKER:** https://github.com/callstack/linaria/issues/1189~~
- [x] ~~💣 **BLOCKER:** https://github.com/callstack/linaria/issues/1193~~


tsconfig.base.json
```diff
@@ -21,6 +21,9 @@
       "@griffel/e2e-utils": ["e2e/utils/src/index.ts"],
       "@griffel/eslint-plugin": ["packages/eslint-plugin/src/index.ts"],
       "@griffel/jest-serializer": ["packages/jest-serializer/src/index.ts"],
+      "@griffel/linaria-processor": ["packages/linaria-processor/src/index.ts"],
+      "@griffel/linaria-processor/make-styles": ["packages/linaria-processor/src/MakeStylesProcessor.ts"],
+      "@griffel/linaria-processor/make-reset-styles": ["packages/linaria-processor/src/MakeResetStylesProcessor.ts"],
```
layershifter, line 26: Aliases that match export maps in `@griffel/linaria-processor`, needed for Jest.

packages/babel-preset/__fixtures__/require/output.ts
```diff
@@ -1,5 +1,5 @@
-const react_make_styles_1 = require('@griffel/react');
-export const useStyles = react_make_styles_1.__styles(
```
layershifter, line 2: `require()` and `export` 🙈

package.json
```diff
@@ -117,19 +118,19 @@
...
+    "@linaria/shaker": "^4.2.7",
+    "@linaria/tags": "^4.3.0",
+    "@linaria/utils": "^4.3.0",
+    "@linaria/webpack-loader": "^4.1.11",
     "@swc/core": "^1.3.19",
     "@typescript-eslint/utils": "^5.20.0",
     "ajv": "^8.4.0",
     "browserslist": "^4.19.1",
     "csstype": "^3.0.10",
-    "enhanced-resolve": "^5.8.2",
```
layershifter, line 132: We don't need to handle resolutions on our end anymore 💅

packages/babel-preset/src/index.test.ts
```diff
@@ -13,11 +13,7 @@ pluginTester({
   babelOptions: {
     parserOpts: {
       plugins: ['typescript'],
-    },
-  },
-  pluginOptions: {
-    babelOptions: {
-      presets: ['@babel/typescript'],
+      sourceType: 'unambiguous',
```
layershifter, line 16: This will be probably defined in Linaria itself. Forces Babel to detect a type of module, so it will emit `import`/`require` properly.

packages/babel-preset/src/index.ts
```diff
@@ -1,17 +1,20 @@
...
-export type { Evaluator, EvalRule } from '@linaria/babel-preset';
+export type { Evaluator, EvalRule } from '@linaria/utils';
 export type { BabelPluginOptions };
 
-export default function griffelPreset(babel: ConfigAPI, options: BabelPluginOptions) {
-  return {
-    plugins: [[transformPlugin, options]],
-  };
+export default function griffelPreset(babel: ConfigAPI, options: BabelPluginOptions = {}) {
+  const { babelOptions, evaluationRules } = validateOptions(options);
```
layershifter, line 14: I decided to not expose all Linaria options to better control APIs.

PR #310, opened 2023-01-09
chore: upgrade Nx to 13.10.6

layershifter:


apps/benchmark/src/implementations/inline-styles/Provider.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 
-const InlineStylesProvider: React.FC = ({ children }) => {
+const InlineStylesProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
```
ling1726, line 3: is this relevant to the NX upgrade ? seems more like React 18 typings changed

ling1726: oh I see you bumped react types to 18 too

PR #315, opened 2023-01-10
chore(webpack-extraction-plugin): refactor to support aliased imports

layershifter:
This PR refactors Babel plugin that replace functions and extracts CSS in `@griffel/webpack-extraction-plugin`. The goal is to use references from imports instead of hardcoded names (`__resetStyles` & `__styles`).

The refactor is needed as we will use transforms from Linaria itself (#309) instead of ours. `@linaria/babel-preset` uses `@babel/helper-module-imports` to generate imports that don't collide in the scope (_which is better_):

```js
// Before
import { __css } from '@griffel/react';
// After
import { __css as _css } from '@griffel/react';
import { __css as _css2 } from '@griffel/react';
```

Internally the plugin is almost the same, but now it takes references via `path.scope.getBinding()` instead of trying to target hardcoded identifiers directly. This allows to be compatible with upcoming changes.

- New fixture (`babel/alias/code.ts`) was added to verify the behavior
- Plugin uses `@babel/helper-module-imports` to emit imports, fixtures have been updated

### DIFF
--- a/packages/webpack-extraction-plugin/__fixtures__/webpack/mixed/output.ts
+++ b/packages/webpack-extraction-plugin/__fixtures__/webpack/mixed/output.ts
@@ -1,9 +1,11 @@
-import { __resetStyles, __styles, __resetCSS, __css } from '@griffel/react';
-export const useClasses = __css({
+import { __css as _css } from '@griffel/react';
+import { __resetCSS as _resetCSS } from '@griffel/react';
+import { __resetStyles, __styles } from '@griffel/react';
+export const useClasses = _css({
   root: {
     sj55zd: 'fe3e8s9',
   },
 });
-export const useClassName = __resetCSS('rjefjbm', 'r7z97ji');
+export const useClassName = _resetCSS('rjefjbm', 'r7z97ji');
 
-require('./code.griffel.css!=!../../../virtual-loader/index.js?style=%2F**%20%40griffel%3Acss-start%20%5Bd%5D%20**%2F%0A.fe3e8s9%7Bcolor%3Ared%3B%7D%0A%2F**%20%40griffel%3Acss-end%20**%2F%0A%2F**%20%40griffel%3Acss-start%20%5Br%5D%20**%2F%0A.rjefjbm%7Bcolor%3Ared%3Bpadding-left%3A4px%3B%7D.r7z97ji%7Bcolor%3Ared%3Bpadding-right%3A4px%3B%7D%0A%2F**%20%40griffel%3Acss-end%20**%2F%0A!./code.ts');
+require('./code.griffel.css!=!../../../virtual-loader/index.js?style=%2F**%20%40griffel%3Acss-start%20%5Br%5D%20**%2F%0A.rjefjbm%7Bcolor%3Ared%3Bpadding-left%3A4px%3B%7D.r7z97ji%7Bcolor%3Ared%3Bpadding-right%3A4px%3B%7D%0A%2F**%20%40griffel%3Acss-end%20**%2F%0A%2F**%20%40griffel%3Acss-start%20%5Bd%5D%20**%2F%0A.fe3e8s9%7Bcolor%3Ared%3B%7D%0A%2F**%20%40griffel%3Acss-end%20**%2F%0A!./code.ts');
--- a/packages/webpack-extraction-plugin/__fixtures__/webpack/multiple/output.ts
+++ b/packages/webpack-extraction-plugin/__fixtures__/webpack/multiple/output.ts
@@ -1,4 +1,3 @@
-import { __css as _css2 } from '@griffel/react';
 import { __css as _css } from '@griffel/react';
 import { __styles } from '@griffel/react';
 export const stylesA = _css({
@@ -6,7 +5,7 @@ export const stylesA = _css({
     sj55zd: 'fe3e8s9',
   },
 });
-export const stylesB = _css2({
+export const stylesB = _css({
   root: {
     De3pzq: 'fcnqdeg',
   },
--- a/packages/webpack-extraction-plugin/__fixtures__/webpack/multiple/output.ts
+++ b/packages/webpack-extraction-plugin/__fixtures__/webpack/multiple/output.ts
@@ -1,10 +1,12 @@
-import { __styles, __css } from '@griffel/react';
-export const stylesA = __css({
+import { __css as _css2 } from '@griffel/react';
+import { __css as _css } from '@griffel/react';
+import { __styles } from '@griffel/react';
+export const stylesA = _css({
   root: {
     sj55zd: 'fe3e8s9',
   },
 });
-export const stylesB = __css({
+export const stylesB = _css2({
   root: {
     De3pzq: 'fcnqdeg',
   },
--- a/packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.test.ts
+++ b/packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.test.ts
@@ -23,8 +23,39 @@ pluginTester({
       },
     }),
 
-  fixtures: fixturesDir,
-  tests: [],
+  tests: [
+    // 🎩 Tip: use "only: true" to run a single test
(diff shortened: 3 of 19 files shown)


packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
```diff
@@ -1,4 +1,6 @@
 import { NodePath, PluginObj, PluginPass, types as t } from '@babel/core';
+// @ts-expect-error Missing typings
+import { addNamed } from '@babel/helper-module-imports';
```
layershifter, line 3: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63871

packages/webpack-extraction-plugin/__fixtures__/webpack/mixed/output.ts
```diff
@@ -1,9 +1,11 @@
...
+export const useClasses = _css({
   root: {
     sj55zd: 'fe3e8s9',
   },
 });
-export const useClassName = __resetCSS('rjefjbm', 'r7z97ji');
+export const useClassName = _resetCSS('rjefjbm', 'r7z97ji');
 
-require('./code.griffel.css!=!../../../virtual-loader/index.js?style=%2F**%20%40griffel%3Acss-start%20%5Bd%5D%20**%2F%0A.fe3e8s9%7Bcolor%3Ared%3B%7D%0A%2F**%20%40griffel%3Acss-end%20**%2F%0A%2F**%20%40griffel%3Acss-start%20%5Br%5D%20**%2F%0A.rjefjbm%7Bcolor%3Ared%3Bpadding-left%3A4px%3B%7D.r7z97ji%7Bcolor%3Ared%3Bpadding-right%3A4px%3B%7D%0A%2F**%20%40griffel%3Acss-end%20**%2F%0A!./code.ts');
+require('./code.griffel.css!=!../../../virtual-loader/index.js?style=%2F**%20%40griffel%3Acss-start%20%5Br%5D%20**%2F%0A.rjefjbm%7Bcolor%3Ared%3Bpadding-left%3A4px%3B%7D.r7z97ji%7Bcolor%3Ared%3Bpadding-right%3A4px%3B%7D%0A%2F**%20%40griffel%3Acss-end%20**%2F%0A%2F**%20%40griffel%3Acss-start%20%5Bd%5D%20**%2F%0A.fe3e8s9%7Bcolor%3Ared%3B%7D%0A%2F**%20%40griffel%3Acss-end%20**%2F%0A!./code.ts');
```
layershifter, line 11: CSS in the query is same, only order was changed. It does matter as Webpack plugin will sort it later.

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.test.ts
```diff
@@ -23,8 +23,39 @@ pluginTester({
       },
     }),
 
-  fixtures: fixturesDir,
-  tests: [],
+  tests: [
```
layershifter, line 26: The same as in `@griffel/babel-preset`, saves time in debugging.

packages/webpack-extraction-plugin/__fixtures__/webpack/multiple/output.ts
```diff
@@ -1,10 +1,12 @@
-import { __styles, __css } from '@griffel/react';
-export const stylesA = __css({
+import { __css as _css2 } from '@griffel/react';
+import { __css as _css } from '@griffel/react';
```
ling1726, line 2: I don't understand why 2 _css imports are necessary here

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
```diff
@@ -24,6 +26,170 @@ function concatCSSRulesByBucket(bucketA: CSSRulesByBucket = {}, bucketB: CSSRule
...
+
+  if (
+    (functionKind === '__styles' && argumentPaths.length !== 2) ||
+    (functionKind === '__resetStyles' && argumentPaths.length !== 3)
+  ) {
+    const calleePath = callExpressionPath.get('callee');
+
+    throw calleePath.buildCodeFrameError(
+      [
+        `"${functionKind}" function call should have exactly ${argumentPaths.length} arguments.`,
```
ling1726, line 73: does this error make sense?

* call `__styles` with 3 arguments
* '__styles shoud have exactly **3** arguments'

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
```diff
@@ -24,6 +26,170 @@ function concatCSSRulesByBucket(bucketA: CSSRulesByBucket = {}, bucketB: CSSRule
...
+        }
+
+        expressionPath.replaceWith(t.stringLiteral(importDeclarationPath.get('source').node.value));
+        importDeclarationPath.remove();
+      });
+    },
+  });
+}
+
+function updateCalleeName(
```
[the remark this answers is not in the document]
layershifter: > is there a need to call addNamed more than once?

Technically, no. I refactored to not do it in this plugin. However, Webpack or any other bundler will endup using the same function anyway.

[Rollup Playground](https://rollupjs.org/repl/?version=2.50.2&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMCU3QiUyMGNzcyUyMGFzJTIwYTElMjAlN0QlMjBmcm9tJTIwJy4lMkZjc3MnJTVDbmltcG9ydCUyMCU3QiUyMGNzcyUyMGFzJTIwYTIlMjAlN0QlMjBmcm9tJTIwJy4lMkZjc3MnJTVDbiU1Q25jb25zb2xlLmxvZyhhMSgpJTJDJTIwYTIoKSklMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSU3RCUyQyU3QiUyMm5hbWUlMjIlM0ElMjJjc3MuanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyZXhwb3J0JTIwZnVuY3Rpb24lMjBjc3MoKSUyMCU3QiU1Q24lMjAlMjByZXR1cm4lMjAnZm9vJyU1Q24lN0QlNUNuJTIyJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmVzJTIyJTJDJTIybmFtZSUyMiUzQSUyMm15QnVuZGxlJTIyJTJDJTIyYW1kJTIyJTNBJTdCJTIyaWQlMjIlM0ElMjIlMjIlN0QlMkMlMjJnbG9iYWxzJTIyJTNBJTdCJTdEJTdEJTJDJTIyZXhhbXBsZSUyMiUzQW51bGwlN0Q=)

packages/webpack-extraction-plugin/__fixtures__/webpack/multiple/output.ts
layershifter: https://github.com/microsoft/griffel/pull/315#discussion_r1067213802

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
layershifter: Nope and this error should be removed at all: `__styles` can be used in another packages and this could throw. I replaced it with other checks.

ling1726: yeah the existing checks already cover that pretty well

packages/webpack-extraction-plugin/src/babelPluginStripGriffelRuntime.ts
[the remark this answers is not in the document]
ling1726: well TIL that bundlers collapse alias instances

ling1726 (approved):

PR #320, opened 2023-01-20
docs: add sharing feature to "Try it out"

### LINKED ISSUE #319: docs: linkable examples in "Try It Out"
This isn't a high priority request.

Try It Out is a great way to experiment with writing styles in Griffel to see the generate output. I'd like to see Try It Out expanded so I can easily share a link to examples ala [Typescript Playground](https://www.typescriptlang.org/play).

Here's an example TS Playground link: https://www.typescriptlang.org/play?#code/PTAEEkBdQdwewK4BsAmoBGBTUBDGmBnOAW2wEsAzUAcQCdKLMkByA0AFVoE8JoB5BJACwAKBCgCCAA5S4tSJjSQAFmQIBCUIF4NwAF7QA

This came up when I was helping out someone new to Griffel and I wanted to share some examples over chat.

layershifter:
Fixes #319.


apps/website/src/components/Playground/PlaygroundUrl.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  return debouncedValue;
+}
+
+/**
+ * The component that checks for code modifications in the editor and updates URL to store it code there.
+ */
+export function PlaygroundUrl({ template }: { template: string }) {
+  const { code, updateCode } = useActiveCode();
+  const searchParam = useDebounce(code === template ? null : lz.compressToEncodedURIComponent(code), 400);
```
layershifter, line 24: Inspired by TypeScript playground: https://github.com/microsoft/TypeScript-Website/blob/6cc9d3087e5161adc9a0aaef4a2ca18d431501c0/packages/sandbox/src/getInitialCode.ts#L21

ling1726: should it also save to localStorage like ts playground so the user doesn't lose previous code?

apps/website/src/components/Playground/PlaygroundShare.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+
+export function PlaygroundShare() {
+  const [visible, setVisible] = React.useState<boolean>(false);
```
ling1726, line 4:
```suggestion
  const [copied, setCopied] = React.useState<boolean>(false);
```

nit

apps/website/src/components/Playground/PlaygroundShare.tsx
```diff
@@ -0,0 +1,31 @@
...
+  React.useEffect(() => {
+    if (visible) {
+      const timeoutId = setTimeout(() => {
+        setVisible(false);
+      }, 2000);
+
+      return () => clearTimeout(timeoutId);
+    }
+
+    return () => {};
```
ling1726, line 23:
```suggestion
```

apps/website/src/components/Playground/PlaygroundShare.tsx
```diff
@@ -0,0 +1,31 @@
...
+  }, [visible]);
+
+  return (
+    <button className="button button--sm button--secondary" onClick={handleClick}>
+      {visible ? '✅ Copied' : 'Share'}
```
ling1726, line 28: nit: no export icon?

ling1726 (approved):

apps/website/src/components/Playground/PlaygroundUrl.ts
```diff
@@ -0,0 +1,50 @@
+import { useActiveCode } from '@codesandbox/sandpack-react';
+import * as lz from 'lz-string';
+import React from 'react';
```
spmonahan, line 3: nit
```suggestion
import * as React from 'react';
```

spmonahan (approved):
That was super fast! This is awesome 🚀

apps/website/src/components/Playground/PlaygroundShare.tsx
layershifter: It's ESLint rule that requires to have returned values

apps/website/src/components/Playground/PlaygroundShare.tsx
layershifter: I tried multiple icons, does not like great 😭

PR #322, opened 2023-01-25
fix: properly handle @media in __resetStyles()

### LINKED ISSUE #321: webpack-extraction-plugin: @media in __resetStyles() creates broken CSS
`@griffel/webpack-extraction-plugin` produces invalid CSS when `__resetStyles()` contains a `@media` rule.

### Input

```ts
import { __resetStyles } from "@griffel/react";
export const useClassName = /*#__PURE__*/ __resetStyles("r13wlxb8", null, [
  ".r13wlxb8{line-height:0;}",
  "@media screen and (prefers-reduced-motion: reduce){.r13wlxb8{transition-duration:0.01ms;}}"
]);
```

### Expected output

```css
.r13wlxb8 {
  line-height: 0;
}
@media screen and (prefers-reduced-motion: reduce) {
  .r13wlxb8 {
    transition-duration: 0.01ms;
  }
}, /* 👈 "," at end */
```

```
    SyntaxError: CssSyntaxError: Unknown word (3:116)
      1 |
      2 | /** griffel.css **/
    > 3 | .r13wlxb8{line-height:0;}@media screen and (prefers-reduced-motion: reduce){.r13wlxb8{transition-duration:0.01ms;}},
        |                                                                                                                    ^

      179 |
      180 |       resultCSS = fixLineEndings(
    > 181 |         prettier.format(result.cssOutput, {
          |                  ^
      182 |           ...prettierConfig,
      183 |           parser: 'css',
      184 |         }),
```

### Expected output

```css
.r13wlxb8 {
  line-height: 0;
}
@media screen and (prefers-reduced-motion: reduce) {
  .r13wlxb8 {
    transition-duration: 0.01ms;
  }
}
```

layershifter:
Fixes #321.

### DIFF
--- a/packages/webpack-extraction-plugin/src/parseCSSRules.ts
+++ b/packages/webpack-extraction-plugin/src/parseCSSRules.ts
@@ -37,7 +37,7 @@ export function parseCSSRules(css: string) {
 
     if (cssBucketName) {
       const cssRule = serialize([element], stringify);
-      const bucketEntry: CSSBucketEntry = element.type === MEDIA ? [cssRule, cssMeta!] : cssRule;
+      const bucketEntry: CSSBucketEntry = cssBucketName === 'm' ? [cssRule, cssMeta!] : cssRule;
 
       cssRulesByBucket[cssBucketName].push(bucketEntry);
       continue;
--- a/packages/webpack-extraction-plugin/src/parseCSSRules.ts
+++ b/packages/webpack-extraction-plugin/src/parseCSSRules.ts
@@ -1,5 +1,5 @@
 import { styleBucketOrdering } from '@griffel/core';
-import { COMMENT, compile, MEDIA, serialize, stringify } from 'stylis';
+import { COMMENT, compile, serialize, stringify } from 'stylis';
 
 import type { CSSBucketEntry, CSSRulesByBucket, StyleBucketName } from '@griffel/core';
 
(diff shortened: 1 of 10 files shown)


packages/webpack-extraction-plugin/src/parseCSSRules.ts
```diff
@@ -37,7 +37,7 @@ export function parseCSSRules(css: string) {
 
     if (cssBucketName) {
       const cssRule = serialize([element], stringify);
-      const bucketEntry: CSSBucketEntry = element.type === MEDIA ? [cssRule, cssMeta!] : cssRule;
```
layershifter, line 40: This condition was not safe as only rules from `m` bucket have `cssMeta` information.

packages/webpack-extraction-plugin/src/parseCSSRules.ts
```diff
@@ -37,7 +37,7 @@ export function parseCSSRules(css: string) {
...
+      const bucketEntry: CSSBucketEntry = cssBucketName === 'm' ? [cssRule, cssMeta!] : cssRule;
```
miroslavstastny, line 40: The GitHub actions warning is really hidden here. Can we consider to complain more visibly?

layershifter: Good call, did not even notice it. Now the rule will throw

PR #335, opened 2023-02-09
chore: run tests for extraction plugin in production mode

layershifter:
This PR forces to run tests for extraction plugin in production mode, so Webpack will run its optimizations like module concatenation, empty chunks removal, etc. Minimizer are turned off to keep code readable.

### DIFF
--- a/packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.test.ts
+++ b/packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.test.ts
@@ -28,13 +28,16 @@ async function compileSourceWithWebpack(
 }> {
   const defaultConfig: webpack.Configuration = {
     context: __dirname,
-    entry: entryPath,
+    entry: {
+      bundle: entryPath,
+    },
 
-    mode: 'development',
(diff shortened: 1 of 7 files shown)


packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.test.ts
```diff
@@ -28,13 +28,16 @@ async function compileSourceWithWebpack(
 }> {
   const defaultConfig: webpack.Configuration = {
     context: __dirname,
-    entry: entryPath,
+    entry: {
+      bundle: entryPath,
+    },
 
-    mode: 'development',
+    mode: 'production',
```
ling1726, line 35: should we have tests for both modes? 🤔

ling1726 (approved):

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.test.ts
layershifter: I was thinking about that, but all our partners use it only in production mode.

PR #338, opened 2023-02-10
refactor: update how griffel chunk is attached

### LINKED ISSUE #314: extraction-plugin: configure attachment of the chunk to entrypoints
Currently the plugin attaches CSS chunk to the main entrypoint:

https://github.com/microsoft/griffel/blob/eb26eafe951623518da5a6fecf8142bab00d85b8/packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts#L17-L33

This is not ideal and we should allow to attach it in a different way or make it configurable.

layershifter:
Fixes #314.

This PR updates how `@griffel/webpack-extraction-plugin` will attach `griffel` chunk. The current way is updated to be under `unstable_attachToMainEntryPoint` option as still required by one of our partners. However, the plugin itself defaults into the behavior of `SplitChunksPlugin` as it does the same job.

> Note: `unstable_attachToMainEntryPoint` will be removed without a further notice, do not rely on it.

### DIFF
--- a/packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
+++ b/packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
@@ -12,10 +12,8 @@ const OPTIMIZE_CHUNKS_STAGE_BASIC = -10;
 
 export type GriffelCSSExtractionPluginOptions = {
   compareMediaQueries?: GriffelRenderer['compareMediaQueries'];
+  unstable_attachToMainEntryPoint?: boolean;
 };
-
-type IterableElement<TargetIterable> = TargetIterable extends Iterable<infer ElementType> ? ElementType : never;
-type ChunkGroup = IterableElement<Chunk['groupsIterable']>;
 
 const PLUGIN_NAME = 'GriffelExtractPlugin';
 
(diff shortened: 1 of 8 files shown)


packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -52,74 +50,67 @@ type CSSModule = Module & {
...
-      compilation.chunkGraph.connectChunkAndModule(griffelChunk, cssModule);
+      compilation.chunkGraph.connectChunkAndModule(griffelChunk, module);
+    }
+  }
 
-      ensureModuleHasPostOrderIndex(griffelChunk, cssModule);
+  if (!attachToMainEntryPoint) {
+    for (const chunk of matchingChunks) {
+      chunk.split(griffelChunk);
     }
```
layershifter, line 100: https://github.com/webpack/webpack/blob/b67626c7b4ffed8737d195b27c8cea1e68d58134/lib/optimize/SplitChunksPlugin.js#L1480-L1483

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -52,74 +50,67 @@ type CSSModule = Module & {
...
-      if (!isGriffelCSSModule(cssModule as CSSModule)) {
-        continue;
+        matchingChunks.add(chunk);
       }
 
-      // https://github.com/webpack/webpack/blob/8241da7f1e75c5581ba535d127fa66aeb9eb2ac8/lib/Chunk.js#L245-L253
-      compilation.chunkGraph.disconnectChunkAndModule(chunk, cssModule);
-      compilation.chunkGraph.connectChunkAndModule(griffelChunk, cssModule);
+      compilation.chunkGraph.connectChunkAndModule(griffelChunk, module);
+    }
```
layershifter, line 94: The logic is almost the same: https://github.com/webpack/webpack/blob/b67626c7b4ffed8737d195b27c8cea1e68d58134/lib/optimize/SplitChunksPlugin.js#L1513

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -52,74 +50,67 @@ type CSSModule = Module & {
...
 
-    if (typeof cssModules === 'undefined') {
-      continue;
-    }
+      for (const chunk of moduleChunks) {
+        compilation.chunkGraph.disconnectChunkAndModule(chunk, module);
+
+        for (const group of chunk.groupsIterable) {
+          group.setModulePostOrderIndex(module, moduleIndex++);
+        }
```
layershifter, line 88: Is done to avoid ordering issues with `mini-css-extract-plugin`.

ling1726 (approved):

PR #346, opened 2023-02-21
feat: add "insertionPoint" option to createDOMRenderer()

### LINKED ISSUE #332: core: allow to control style tag insertion location
[Internally Griffel ensures its style tags are inserted in the correct order to provide deterministic results](https://github.com/microsoft/griffel/blob/main/packages/core/src/renderer/getStyleSheetForBucket.ts#L65). However, when used in an application that lazy loads code and styles, Griffel's style tags may become interleaved with the application's style tags, resulting in non-deterministic results.

This manifests in real-world applications that use lazy loading of React components (and CSS modules for application styles in this case but I don't think that's necessarily relevant) and Fluent UI React v9 for controls. In this case application authors want to override default Fluent styles with their own CSS but because their components are lazily loaded and they cannot control where Griffel styles are loaded in the document they cannot be guaranteed their styles will have greater specificity than Fluent's built in Griffel styles.

I've created a rough demo of this on Codesandbox: [Griffel Lazy Load](https://codesandbox.io/s/griffel-lazy-load-example-j53dfe)

With the default example, by inspecting the source, you can see how some application CSS is loaded in the document first, then Griffel styles, then more app CSS.

I'd like an API for Griffel that allows me to control where Griffel's style tags are inserted so I can ensure they are in the correct place relative to my application's styles.

layershifter:
Fixes #332.

This PR adds `insertionPoint` option. If specified, a renderer will insert created style tags after this element:

```js
import { createDOMRenderer } from '@griffel/react';

const insertionPoint = document.head.querySelector('#foo');
const renderer = createDOMRenderer(document, {
  insertionPoint,
});
```

```html
<html>
  <head>
    <style id="foo" />
    <!-- Style elements created by Griffel will be inserted after "#foo" element -->
    <style data-make-styles-bucket="d" />
    <style id="bar" />
  </head>
</html>
```

### DIFF
--- a/packages/core/src/renderer/getStyleSheetForBucket.test.ts
+++ b/packages/core/src/renderer/getStyleSheetForBucket.test.ts
@@ -10,6 +10,33 @@ function createFakeDocument(): Document {
 }
 
 describe('getStyleSheetForBucket', () => {
+  it('creates elements order', () => {
+    const target = createFakeDocument();
+    const renderer = createDOMRenderer();
+
+    getStyleSheetForBucket('l', target, null, renderer);
+    getStyleSheetForBucket('r', target, null, renderer);
+    getStyleSheetForBucket('a', target, null, renderer);
+    getStyleSheetForBucket('d', target, null, renderer);
+
+    expect(target.head.children).toMatchInlineSnapshot(`
+      HTMLCollection [
+        <style
+          data-make-styles-bucket="r"
+        />,
+        <style
+          data-make-styles-bucket="d"
+        />,
+        <style
+          data-make-styles-bucket="l"
+        />,
+        <style
+          data-make-styles-bucket="a"
+        />,
+      ]
+    `);
+  });
+
   it('sets "data-make-styles-bucket" attribute in order', () => {
     const target = createFakeDocument();
     const renderer = createDOMRenderer();
@@ -122,10 +149,10 @@ describe('getStyleSheetForBucket', () => {
           data-test="B"
         />,
         <style
-          data-make-styles-bucket="d"
+          data-make-styles-bucket="r"
         />,
         <style
-          data-make-styles-bucket="r"
+          data-make-styles-bucket="d"
         />,
         <style
           data-test="C"
--- a/packages/core/src/renderer/getStyleSheetForBucket.test.ts
+++ b/packages/core/src/renderer/getStyleSheetForBucket.test.ts
@@ -1,6 +1,6 @@
-import { getStyleSheetForBucket, styleBucketOrdering } from './getStyleSheetForBucket';
-import { createDOMRenderer } from './createDOMRenderer';
 import { DATA_BUCKET_ATTR } from '../constants';
+import { createDOMRenderer } from './createDOMRenderer';
+import { getStyleSheetForBucket, styleBucketOrdering } from './getStyleSheetForBucket';
 
 function createFakeDocument(): Document {
   const doc = document.implementation.createDocument('http://www.w3.org/1999/xhtml', 'html', null);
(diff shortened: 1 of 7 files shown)


packages/core/src/renderer/getStyleSheetForBucket.ts
```diff
@@ -45,25 +45,31 @@ const styleBucketOrderingMap = styleBucketOrdering.reduce((acc, cur, j) => {
...
 
-    if (target && tag) {
-      const elementSibling = findElementSibling(target, bucketName, renderer, metadata);
-      target.head.insertBefore(tag, elementSibling);
+    if (targetDocument && tag) {
+      let beforeElement: Node | null = findElementSibling(targetDocument, bucketName, renderer, metadata);
+
+      if (beforeElement === null && insertionPoint) {
+        beforeElement = insertionPoint.nextSibling;
+      }
```
ling1726, line 70:
```suggestion
      let beforeElement: Node | null = findElementSibling(targetDocument, bucketName, renderer, metadata) ?? insertionPoint?.nextSibling;
```

nit: one liner?

packages/core/src/renderer/getStyleSheetForBucket.ts
```diff
@@ -45,25 +45,31 @@ const styleBucketOrderingMap = styleBucketOrdering.reduce((acc, cur, j) => {
  */
 export function getStyleSheetForBucket(
   bucketName: StyleBucketName,
-  target: Document | undefined,
+  targetDocument: Document | undefined,
+  insertionPoint: HTMLElement | null,
   renderer: GriffelRenderer,
```
ling1726, line 50: nit: refactor this to use a DTO now? it would make things a bit clearer what's going on, particularly tests

ling1726 (approved):

spmonahan (approved):
I think this will work great for my case. Thanks for turning it around so quickly!

packages/core/src/renderer/getStyleSheetForBucket.ts
```diff
@@ -45,25 +45,31 @@ const styleBucketOrderingMap = styleBucketOrdering.reduce((acc, cur, j) => {
...
     renderer.stylesheets[stylesheetKey] = stylesheet;
 
-    if (target && tag) {
-      const elementSibling = findElementSibling(target, bucketName, renderer, metadata);
-      target.head.insertBefore(tag, elementSibling);
+    if (targetDocument && tag) {
+      let beforeElement: Node | null = findElementSibling(targetDocument, bucketName, renderer, metadata);
+
+      if (beforeElement === null && insertionPoint) {
+        beforeElement = insertionPoint.nextSibling;
```
behowell, line 69: I'm not sure this will work as intended? `findElementSibling` returns `null` when the new griffel bucket should be _after_ all other griffel buckets. 

The first griffel bucket inserted will appear after `insertionPoint` as expected. However, if you later add a bucket that should come last, `findElementSibling` will return `null`, and the new one will be inserted directly after `insertionPoint`, and _before_ any other griffel buckets.

behowell: One option is to have the `insertionPoint` point API instead be an "insert before this node" API. That has a few benefits:
1. The fix to the above problem is simple, change this line to `beforeElement = insertionPoint;` (or a one-liner definition `const beforeElement = findElementSibling(...) || insertionPoint;`)
2. It reduces the chances that style nodes inserted by other libraries will be interleaved with griffel styles. Either the other library inserts at the top of `<head>`, in which case it'll always be before griffel's styles. Or the other library inserts at the bottom of `<head>`, in which case it'll always be after `insertionPoint`.

packages/core/src/renderer/getStyleSheetForBucket.test.ts
```diff
@@ -93,4 +93,44 @@ describe('getStyleSheetForBucket', () => {
...
+        />,
+        <style
+          data-test="B"
+        />,
+        <style
+          data-make-styles-bucket="d"
+        />,
+        <style
+          data-make-styles-bucket="r"
+        />,
```
behowell, line 129: Related to my comment in getStyleSheetForBucket.ts, I think this order is wrong? `r` should come before `d` right?

packages/core/src/renderer/getStyleSheetForBucket.ts
ling1726: Wow good catch @behowell 👍

I agree that `1.`  will work as expected. It seems that we went with the insertBefore convention already with `findElementSibling` and it makes sense to keep it that way.

Thinking about it a bit more, if we still wante the the insertAfter behaviour, then the `findElementSibling` should return the element _after which to insert the new element_. It's a convention thing 🤷, currently it's internal but once we set the convention with `insertionPoint` we can actually never change it easily 😅.

Also having said that we should rename `findElementSibling` to be more accurate to what kind of sibling it's returning. I thought I originally named the thing pretty clearly, but it's actually quite 💩💩 and we've fallen into our own trap.

packages/core/src/renderer/getStyleSheetForBucket.test.ts
layershifter: @behowell, indeed the order is wrong. Good catch 👍

packages/core/src/renderer/getStyleSheetForBucket.ts
layershifter: > One option is to have the insertionPoint point API instead be an "insert before this node" API. That has a few benefits:

@behowell there are valid concerns, however I prefer to keep the same API as at least someone has already (https://github.com/microsoft/griffel/issues/332#issuecomment-1438574739), in this case Emotion's behavior looks nicest to me.

I didn't pay attention for this in the previous implementation 🙈 I reworked it in 27ad3cf and it should work as expected, thanks for heads up 👍

behowell: Sounds good! Thanks for making the update.

behowell (approved):

ling1726 (approved):

PR #356, opened 2023-03-27
chore(extraction-plugin): redo chunk attachment

layershifter:
This PR redoes chunk management in the plugin to prevent "phantom chunks" that we were creating before.

- Griffel chunk will be created only if needed
- A hack with `GriffelDummyModule` was removed
- We will reuse `SplitChunksPlugin` if it's enabled to avoid redundant module movements
- New fixture added for `optiminizatin.splitChunks = false`


packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -115,48 +107,50 @@ export class GriffelCSSExtractionPlugin {
...
+        // Performs module movements if SplitChunksPlugin is not present
+        compilation.hooks.optimizeChunks.tap({ name: PLUGIN_NAME, stage: OPTIMIZE_CHUNKS_STAGE_ADVANCED }, () => {
+          moveCSSModulesToGriffelChunk(compilation);
+        });
+      }
 
+      // In this case we detach Griffel chunk from all other chunks and connect it with the main entrypoint
       if (this.attachToMainEntryPoint) {
-        compilation.hooks.afterChunks.tap(PLUGIN_NAME, () => {
+        compilation.hooks.optimizeChunks.tap({ name: PLUGIN_NAME, stage: 10 }, () => {
```
ling1726, line 130:
```suggestion
        compilation.hooks.optimizeChunks.tap({ name: PLUGIN_NAME, stage:  OPTIMIZE_CHUNKS_STAGE_ADVANCED}, () => {
```

ling1726 (approved):

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -115,48 +107,50 @@ export class GriffelCSSExtractionPlugin {
...
-        compilation.chunkGraph.disconnectChunkAndModule(griffelChunk, dummyModule);
-      });
-
       compilation.hooks.processAssets.tap(
         {
           name: PLUGIN_NAME,
           stage: Compilation.PROCESS_ASSETS_STAGE_PRE_PROCESS,
         },
         assets => {
+          const griffelChunk = compilation.namedChunks.get('griffel');
```
ling1726, line 148: is the check for CSS assets under this new early return still necessary? there should always be assets if a griffel chunk exists?

layershifter: I would keep it to make TS happy and `.find()` can return `undefined`

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -115,48 +107,50 @@ export class GriffelCSSExtractionPlugin {
...
+      };
+    }
+
     compiler.hooks.compilation.tap(PLUGIN_NAME, compilation => {
-      // A chunk where we will move all CSS modules
-      const griffelChunk = compilation.addChunk('griffel');
-      // Set up a dummy module to prevent the chunk from getting cleaned up by RemoveEmptyChunksPlugin
-      const dummyModule = new GriffelDummyModule();
+      if (!compiler.options.optimization.splitChunks) {
+        // Performs module movements if SplitChunksPlugin is not present
```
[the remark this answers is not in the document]
layershifter: I update all comments in this file, should be clearer now.

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
layershifter: I moved these taps into a single call.

PR #358, opened 2023-04-03
feat: Support CSS container queries

### LINKED ISSUE #270: core: add support for @container
`@container` is supported in the latest Chrome, Edge and Safari.
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries

ling1726:
Adds support for `@container` queries. Container queries will be inserted in a bucket that is always after `@media` buckets.

This means that **styles in @container queries will always win over @media queries**.

There is no support for nested container queries or sorting the container query conditions yet.

Fixes #270


layershifter:
Let's update docs to mention `@container` support:
- README for `@griffel/react`
- https://griffel.js.org/react/api/make-styles#at-rules
- https://github.com/microsoft/griffel/tree/main/packages/react#pseudo--class-selectors-at-rules-global-styles
- Add example to playground

layershifter (approved):

PR #361, opened 2023-04-04
Add a rule to enforce makeStyles calls only happen in .styles.ts files

christiango:
We've seen issues internally where our webpack bundle times have regressed when integrating the griffel CSS extraction plugins and loaders. We found that having all makeStyles in a .styles.ts file lets us limit the scope of where these plugins and loaders run to just files ending in .styles.ts which dramatically improved performance of our bundle tasks.

This code was heavily adapted from existing code by Yuanbo Xue in the Microsoft Teams repository.


layershifter (approved):

christiango:
@layershifter any chance this will be available in an upcoming npm publish? Would like to replace our internal implementation with this one.

layershifter:
@christiango published in `@griffel/eslint-plugin@1.2.0`.

PR #375, opened 2023-05-17
fix(core|babel-preset): deduplicate CSS rules in output

### LINKED ISSUE #372: core/babel-preset: duplicated rules with at-rules
The following snippet with flipping properties in RTL causes CSS rule duplication:

```ts
import { makeStyles, shorthands } from "@griffel/core";

export default makeStyles({
  root: {
    "@media(forced-colors: active)": {
      paddingLeft: "8px",
      paddingRight: "8px"
    },
    "@supports (display: flex)": {
      paddingLeft: "4px",
      paddingRight: "4px"
    }
  }
});
```

This affects both CSS extraction and transformed code:

```js
// Transformed code
  {
    root: {
      cvzfql: ["f13wceqr", "fmtq1hn"],
      g42fl4: ["fmtq1hn", "f13wceqr"],
      B6rh290: ["f16pi78z", "fem5dos"],
      rth2jb: ["fem5dos", "f16pi78z"]
    }
  },
  {
    m: [
      [
        "@media (min-width: 600px){.f13wceqr{padding-left:4px;}.fmtq1hn{padding-right:4px;}}",
        {
          m: "(min-width: 600px)"
        }
      ],
      [
        "@media (min-width: 600px){.fmtq1hn{padding-right:4px;}.f13wceqr{padding-left:4px;}}",
        {
          m: "(min-width: 600px)"
        }
      ]
    ],
    t: [
      "@supports (display: flex){.f16pi78z{padding-left:4px;}.fem5dos{padding-right:4px;}}",
      "@supports (display: flex){.fem5dos{padding-right:4px;}.f16pi78z{padding-left:4px;}}"
    ]
  }
);
```

```css
// Extracted CSS
@media (forced-colors: active) {
  .f1xnl3gg {
    padding-left: 8px;
  }

  .fyzjxl6 {
    padding-right: 8px;
  }
}

@media (forced-colors: active) {
  .fyzjxl6 {
    padding-right: 8px;
  }

  .f1xnl3gg {
    padding-left: 8px;
  }
}
```

[Playground](https://griffel.js.org/try-it-out/?code=JYWwDg9gTgLgBAbziAhgawKYGUYE8A2GAzgDRxEAW0MFKAdgCZFwC%2BcAZlBCHAOQACAcyjB27DPgD0AY2gZeAbgBQSjAA9IsOAwzsUAV3zxUmHAWIAKBErhwuEGAC5EN2334gMDYCgvto0l4AtLL40ETOKNIwwABuGACUvM7Wbm5gKAzedIIAMrpOfAAcYGq8JK5pGVnAOQBKwIIUhbwlZRVpLB2sFSwJykA)

layershifter:
Fixes #372.

The fix contains two parts:
- **Sorting of classes inside rules (only for Atomic CSS)**
  This mandatory to handle RTL flipping on directional properties. Sorting is done via a Stylis plugin.

  ```css
  /* Before: two identical CSS rules */
  @media {
    .pl { padding-left: 10px }
    .pr { padding-right: 10px }
  }
  @media {
    .pr { padding-right: 10px }
    .pl { padding-left: 10px }
  }

  /* After: single CSS rule */
  @media {
    .pr { padding-right: 10px }
    .pl { padding-left: 10px }
  }
  ```
- **Fixes to deduplication in Babel plugin**
  `dedupeCSSRules()` did not handle rules with meta information i.e. media queries.


packages/babel-preset/src/transformPlugin.ts
```diff
@@ -107,20 +102,6 @@ function isRequireDeclarator(
...
- */
-function dedupeCSSRules(cssRules: CSSRulesByBucket): CSSRulesByBucket {
-  (Object.keys(cssRules) as StyleBucketName[]).forEach(styleBucketName => {
-    cssRules[styleBucketName] = cssRules[styleBucketName]!.filter(
-      (rule, index, rules) => rules.indexOf(rule) === index,
-    );
-  });
-
-  return cssRules;
-}
```
layershifter, line 122: Moved to a separate file.

packages/babel-preset/src/utils/dedupeCSSRules.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export function dedupeCSSRules(cssRulesByBucket: CSSRulesByBucket): CSSRulesByBucket {
+  return Object.fromEntries(
+    Object.entries(cssRulesByBucket).map(([styleBucketName, cssBucketEntries]) => {
+      if (styleBucketName === 'm') {
+        return [
+          styleBucketName,
+          cssBucketEntries.filter(
+            (entryA, index, entries) => entries.findIndex(entryB => entryA[0] === entryB[0]) === index,
+          ),
```
layershifter, line 15: Opened for better ideas on making that array unique.

packages/core/src/runtime/stylis/sortClassesInAtRulesPlugin.ts
```diff
@@ -0,0 +1,18 @@
...
+} from 'stylis';
+import type { Middleware } from 'stylis';
+
+export const sortClassesInAtRulesPlugin: Middleware = element => {
+  switch (element.type) {
+    case MEDIA:
+    case SUPPORTS:
+    case LAYER:
+      if (Array.isArray(element.children)) {
+        element.children.sort((a, b) => (a.props[0] > b.props[0] ? 1 : -1));
```
layershifter, line 15: As we have there only two rules such primitive sort should be good enough.

packages/babel-preset/src/utils/dedupeCSSRules.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    const rules: CSSRulesByBucket = {
+      d: ['.foo { color: red; }', '.foo { color: red; }', '.baz { color: pink }'],
+    };
+
+    expect(dedupeCSSRules(rules)).toMatchObject({
+      d: ['.foo { color: red; }', '.baz { color: pink }'],
+    });
+  });
+
+  it('dedupes rules with metadata', () => {
```
ling1726, line 15: nit: would be worth having a test case with same rule but different media query+metadata

packages/babel-preset/src/utils/dedupeCSSRules.ts
ling1726: this is during build time so speed might not be an issue. However, if you did want to make this faster from a compute time PoV, you could sort the array. Doing a sort operation and removing duplicates after will be cheaper than the current way you are doing it.

Also a bigger arch comment that also might not be so important but internally it should be pretty easy to maintain order in the CSS buckets if you just did ordered insertion with a binary search, it would make insertion into the bucket slightly slower, but mean less work afterwards.

This is more or less a larger nit out of intereset 🤷, no expectation to do any of it

packages/babel-preset/src/utils/dedupeCSSRules.ts
```diff
@@ -0,0 +1,22 @@
...
+        ];
+      }
+
+      return [styleBucketName, [...new Set(cssBucketEntries)]];
```
ling1726, line 19: why a set here? it's essentially just a set of array references, do you need reference comparison later?

packages/core/src/runtime/compileCSSRules.ts
```diff
@@ -0,0 +1,28 @@
+import { compile, middleware, prefixer, rulesheet, serialize, stringify } from 'stylis';
+
+import { globalPlugin } from './stylis/globalPlugin';
+import { sortClassesInAtRulesPlugin } from './stylis/sortClassesInAtRulesPlugin';
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+function noop() {}
+
+export function compileCSSRules(cssRules: string, sortClassesInAtRules = false): string[] {
```
ling1726, line 9: why make the sort optional at all? it's not configurable in userland

ling1726 (approved):
Approved with some questions/nits

packages/core/src/runtime/compileCSSRules.ts
layershifter: Makes sense, made it required

PR #380, opened 2023-06-20
Added some more info about class ordering to the mergeClasses documentation.

GeoffCoxMSFT:
Called out that the resulting class name order might be different than the order passed to mergeClasses.  Noted that the final stylesheet is all that really matters.

### DIFF
--- a/apps/website/docs/react/api/merge-classes.md
+++ b/apps/website/docs/react/api/merge-classes.md
@@ -43,8 +43,18 @@ function Component(props) {
 
 :::note
 
-Compared to native CSS order of arguments in `mergeClasses()` determines the results. This lets you control styles with JavaScript.
+Compared to native CSS, the order of arguments in `mergeClasses()` affects the results. This lets you control styles with JavaScript.
 
+Each parameter passed to `mergeClasses()` is processed in turn.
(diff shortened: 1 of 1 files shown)


layershifter:
@GeoffCoxMSFT thanks for changes, we worked in offline with @ling1726 to improve it. Can you please provide feedback for the changes?

apps/website/docs/react/api/merge-classes.md
```diff
@@ -29,24 +29,61 @@ function Component(props) {
...
+    }
+```
+
+Compared to native CSS, the order of arguments in `mergeClasses()` affects the results. This lets you control styles with JavaScript.
 
 :::
 
+In CSS the order of the class names doesn't affect the application of styles. If you have issues with style specificity or ordering, look at the generated CSS rather than the class names.
+
+Griffel perform ordering of CSS rules, check [the order of pseudo classes](react/guides/atomic-css#lvfha-order-of-pseudo-classes).
```
GeoffCoxMSFT, line 67:
```suggestion
Griffel performs CSS rule ordering, check [the order of pseudo classes](react/guides/atomic-css#lvfha-order-of-pseudo-classes).
```

layershifter (approved):

PR #385, opened 2023-07-07
fix(shadow-dom): update sorting logic

layershifter:
Fixes an issue sorting logic with `@media` queries in `@griffel/shadow-dom`. New test cases added.


packages/shadow-dom/src/findInsertionPoint.ts
```diff
@@ -29,9 +29,13 @@ export function findInsertionPoint(
     const styleElement = styleSheets[i];
 
     if (comparer(styleElement) > 0) {
-      return styleElement;
+      return styleSheets[i + 1] || null;
```
ling1726, line 32:
```suggestion
      return styleSheets[i + 1] ?? null;
```

nit

ling1726 (approved):

PR #388, opened 2023-07-13
feat: Prefix according to browser support matrix

### LINKED ISSUE #379: feature(core): `transform` is expanded to 3 different vendor prefixes
When using the css [transform](https://developer.mozilla.org/en-US/docs/Web/CSS/transform) property, Griffel will expand this to include vendor prefixing for:

* webkit
* moz
* ms (i.e 11 I guess?)

This results in a style rule that will have 4 times its length (including the actual transform) and negatively impacts bundlesize

ling1726:
Copies the source from the [stylis prefixer](https://github.com/thysultan/stylis/blob/master/src/Prefixer.js) and updates it so that prefixing is only applied according the the browser support matrix. We follow the [browser support matrix](https://github.com/microsoft/fluentui/blob/master/rfcs/shared/build-system/06-browser-support-for-v9.md) of Fluent UI

Fixes #379

### DIFF
--- a/packages/core/src/runtime/compileAtomicCSSRule.test.ts
+++ b/packages/core/src/runtime/compileAtomicCSSRule.test.ts
@@ -14,6 +14,53 @@ const defaultOptions: Pick<
 };
 
 describe('compileAtomicCSSRule', () => {
+  it.only.each([
+    ['transform', 'none'],
+    ['flex-grow', '1'],
+    ['flex-shrink', '1'],
+    ['flex-basis', 'auto'],
+    ['flex-direction', 'column'],
+    ['align-self', 'center'],
+    ['align-content', 'center'],
+    ['align-items', 'center'],
+    ['order', '1'],
+    ['justify-content', 'center'],
+    ['display', 'flex'],
+    ['display', 'inline-flex'],
+    ['display', 'grid'],
+    ['display', 'inline-grid'],
+    ['transition', 'margin-right 4s'],
+    ['writing-mode', 'vertical-lr'],
+    ['columns', '2'],
+    ['text-size-adjust', 'none'],
+    ['text-decoration', 'none'],
+    ['filter', 'blur(5px)'],
+    ['position', 'sticky'],
+    ['clip-path', 'circle(40%)'],
+    ['width', 'fit-content'],
+    ['width', 'min-block-size'],
+    ['background-clip', 'text'],
+    ['animation', '3s linear 1s slidein;'],
+    ['animation-delay', '3s'],
+    ['animation-direction', 'normal'],
+    ['animation-duration', '3s'],
+    ['animation-fill-mode', 'both'],
+    ['animation-iteration-count', '10'],
+    ['animation-name', 'foo'],
+    ['animation-play-state', 'running'],
+    ['animation-timing-function', 'linear'],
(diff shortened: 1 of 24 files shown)


packages/core/src/runtime/prefixer.ts
```diff
@@ -0,0 +1,161 @@
+/* eslint-disable no-fallthrough */
+import {
```
ling1726, line 2: Most of the changelist is updating tests - main feature to review is here

packages/core/src/runtime/compileAtomicCSSRule.test.ts
```diff
@@ -160,10 +207,10 @@ describe('compileAtomicCSSRule', () => {
...
         }),
       ).toMatchInlineSnapshot(`
-      Array [
-        ".fui-FluentProvider .foo .focus:hover{color:red;}",
-      ]
-    `);
+              Array [
+                ".fui-FluentProvider .foo .focus:hover{color:red;}",
+              ]
+          `);
```
layershifter, line 213: Revert this?

packages/core/src/runtime/compileAtomicCSSRule.test.ts
```diff
@@ -14,6 +14,53 @@ const defaultOptions: Pick<
 };
 
 describe('compileAtomicCSSRule', () => {
+  it.only.each([
```
layershifter, line 17:
```suggestion
  it.each([
```

packages/core/src/runtime/compileAtomicCSSRule.test.ts
```diff
@@ -178,11 +225,11 @@ describe('compileAtomicCSSRule', () => {
...
-      Array [
-        "body .foo{padding-left:10px;}",
-        "body .rtl-foo{padding-right:10px;}",
-      ]
-    `);
+              Array [
+                "body .foo{padding-left:10px;}",
+                "body .rtl-foo{padding-right:10px;}",
+              ]
+          `);
```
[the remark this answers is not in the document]
ling1726: This was applied with the formatter

packages/core/src/runtime/compileAtomicCSSRule.test.ts
ling1726: This was applied with the formatter, I applied `yarn nx format`and check passed

packages/core/src/runtime/compileAtomicCSSRule.test.ts
ling1726: good catch, fixed

packages/core/src/runtime/prefixer.ts
```diff
@@ -0,0 +1,161 @@
...
+      if (!match(value, /grab/)) {
+        return (
+          replace(replace(replace(value, /(zoom-|grab)/, WEBKIT + '$1'), /(image-set)/, WEBKIT + '$1'), value, '') +
+          value
+        );
+      }
+    // background, background-image
+    case 5495:
+    case 3959:
+      return replace(value, /(image-set\([^]*)/, WEBKIT + '$1' + '$`$1');
```
layershifter, line 58: Can you please silence warnings there?

layershifter (approved):

PR #392, opened 2023-07-17
feat(core): support variadic arguments in border* shorthands

### LINKED ISSUE #393: feature(core): border shorthands should support variadic arguments
`border*` shorthands functions should support the different order of arguments.

To be clear, it's following cases:

- `border('solid')`
- `border('solid', '2px')`
- `border('solid', '2px', 'red')`

layershifter:
This PR adds functionality to `border*` shorthands functions to support the different order of arguments. For example,

```diff
border('2px')
+border('solid')
+border('none')
border('2px', 'solid')
+border('solid', '2px')
border('2px', 'solid', 'red')
+border('solid', '2px', 'red')
```

Fixes #393.

### DIFF
--- a/packages/core/src/shorthands/border.ts
+++ b/packages/core/src/shorthands/border.ts
@@ -1,8 +1,10 @@
 import type { GriffelStyle } from '@griffel/style-types';
+
 import { borderWidth } from './borderWidth';
 import { borderStyle } from './borderStyle';
 import { borderColor } from './borderColor';
 import type { BorderColorInput, BorderStyleInput, BorderWidthInput } from './types';
+import { isBorderStyle } from './utils';
 
 type BorderStyle = Pick<
   GriffelStyle,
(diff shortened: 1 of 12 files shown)


packages/core/src/shorthands/border.ts
```diff
@@ -29,15 +31,28 @@ export function border(width: BorderWidthInput, style: BorderStyleInput, color:
...
+ *  border('solid')
  *  border('2px', 'solid')
+ *  border('solid', '2px')
  *  border('2px', 'solid', 'red')
+ *  border('solid', '2px', 'red')
  *
  * See https://developer.mozilla.org/en-US/docs/Web/CSS/border
  */
-export function border(...values: [BorderWidthInput, BorderStyleInput?, BorderColorInput?]): BorderStyle {
+export function border(
```
ling1726, line 42: does it make more sense to use some kind of runtime check to determine the values?

There is no point IMO is 'half' supporting variadic arguments since the CSS spec states that they can be in any order

https://developer.mozilla.org/en-US/docs/Web/CSS/border#syntax

ling1726 (approved):

PR #397, opened 2023-07-18
chore: use modern APIs in tests

layershifter:
This modifies tests to use non-deprecated APIs from React 18. Also adds some new assertions to existing tests.


packages/react/src/renderToStyleElements.test.tsx
```diff
@@ -43,14 +45,17 @@ describe('renderToStyleElements (DOM)', () => {
...
-        <ExampleComponent />
-      </RendererProvider>,
-    );
+    act(() => {
+      root.render(
+        <RendererProvider renderer={renderer}>
+          <ExampleComponent />
+        </RendererProvider>,
+      );
+    });
```
layershifter, line 56: This test runs in DOM env, it should use DOM APIs.

PR #407, opened 2023-07-24
feat: insertionFactory API for @griffel/core

layershifter:
This PR adds `insertionFactory` API to `@griffel/core` and related functions (`makeStyles`, `makeResetStyles`, etc.).

Every factory should return `insertStyles` function:

```ts
import type { GriffelInsertionFactory } from '@griffel/react'

const insertionFactory: GriffelInsertionFactory = () => {
  return function insertStyles() {
    /* do smth */
  }
}
```

While customization of insertion can be done by custom renderers (https://griffel.js.org/react/api/create-dom-renderer), the goal of this API is to allow customization of insertion process. For example, instead of inserting styles immediately it can be done via `useInsertionEffect()`.

```ts
import type { GriffelInsertionFactory } from '@griffel/react'

const insertionFactory: GriffelInsertionFactory = () => {
  return function insertStyles(renderer, cssRules) {
    React.useInsertionEffect(
      () => {
        renderer.insertRules(cssRules)
      },
      [renderer, cssRules]
  }
}
```

### DIFF
--- a/packages/core/src/makeStyles.ts
+++ b/packages/core/src/makeStyles.ts
@@ -1,15 +1,20 @@
 import { debugData, isDevToolsEnabled, getSourceURLfromError } from './devtools';
+import { insertionFactory } from './insertionFactory';
 import { resolveStyleRulesForSlots } from './resolveStyleRulesForSlots';
 import { reduceToClassNameForSlots } from './runtime/reduceToClassNameForSlots';
 import type { CSSClassesMapBySlot, CSSRulesByBucket, GriffelRenderer, StylesBySlots } from './types';
+import type { GriffelInsertionFactory } from './types';
 
 export interface MakeStylesOptions {
   dir: 'ltr' | 'rtl';
   renderer: GriffelRenderer;
 }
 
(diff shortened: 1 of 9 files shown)


packages/core/src/makeStyles.ts
```diff
@@ -1,15 +1,20 @@
...
   renderer: GriffelRenderer;
 }
 
-export function makeStyles<Slots extends string | number>(stylesBySlots: StylesBySlots<Slots>) {
-  const insertionCache: Record<string, boolean> = {};
+export function makeStyles<Slots extends string | number>(
+  stylesBySlots: StylesBySlots<Slots>,
+  factory: GriffelInsertionFactory = insertionFactory,
+) {
+  const insertStyles = factory();
```
ling1726, line 17: would be nice to add tests for insertion factory usage

ling1726 (approved):

packages/core/src/makeStyles.ts
layershifter: I added tests for `insertionFactory()` itself and its connection with `makeStyles` 👍

PR #412, opened 2023-07-27
Inherit more webpack resolve options

MLoughry:
The Outlook project has `webpack` build directly from Typescript source using `esbuild-loader`; to accomplish this, we define a `"source"` condition on all of our various package `"exports"` and specify it in `resolve.conditionNames` in `webpack.config.js`. The current behavior to lock this option to simply `["require"]` is a blocker to integrating the loader into our config.

### DIFF
--- a/change/@griffel-webpack-loader-b9e23b96-1848-4762-8b00-bb466cbbe407.json
+++ b/change/@griffel-webpack-loader-b9e23b96-1848-4762-8b00-bb466cbbe407.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Inherit more resolve options from webpack config",
+  "comment": "feat: Add 'inheritResolveOptions' to configure resolve options in Webpack",
   "packageName": "@griffel/webpack-loader",
   "email": "miclo@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@griffel-webpack-loader-b9e23b96-1848-4762-8b00-bb466cbbe407.json
+++ b/change/@griffel-webpack-loader-b9e23b96-1848-4762-8b00-bb466cbbe407.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Inherit more resolve options from webpack config",
+  "packageName": "@griffel/webpack-loader",
+  "email": "miclo@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/webpack-loader/README.md
+++ b/packages/webpack-loader/README.md
@@ -159,7 +159,9 @@ module.exports = {
 ```
 
 ### Configuring webpack resolve options
-If your `@griffel/react` modules import other files (eg., a set of common mixins or colors for your app), the loader resolves these using `enhanced-resolve`. By default, it inherits the settings `resolve.alias`, `resolve.modules`, and `resolve.modules` from youe webpack config, while using its own default values for `resolve.extensions` and `resolve.conditionNames`. If you want to change this behavior, you can choose which `resolve` options are inherited from your webpack config.
+If your `@griffel/react` modules import other files (eg., a set of common mixins or colors for your app), the loader resolves these using `enhanced-resolve`. By default, it inherits the settings `resolve.alias`, `resolve.modules`, and `resolve.modules` from your Webpack config, while using its own default values for `resolve.extensions` and `resolve.conditionNames`.
+
+If you want to change this behavior, you can choose which [`resolve` options](https://webpack.js.org/configuration/resolve/) are inherited from your Webpack config.
 ```js
 module.exports = {
   module: {
--- a/packages/webpack-loader/README.md
+++ b/packages/webpack-loader/README.md
@@ -10,6 +10,7 @@ A loader for Webpack 5 that performs build time transforms for [`@griffel/react`
 - [Usage](#usage)
   - [Handling Griffel re-exports](#handling-griffel-re-exports)
   - [Configuring Babel settings](#configuring-babel-settings)
+  - [Configuring webpack resolve options](#configuring-webpack-resolve-options)
   - [Configuring module evaluation](#configuring-module-evaluation)
 - [Troubleshooting](#troubleshooting)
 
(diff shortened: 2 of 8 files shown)


packages/webpack-loader/src/webpackLoader.ts
```diff
@@ -56,21 +56,18 @@ export function webpackLoader(
...
   // There is this.resolve(), but it's asynchronous. Another option is to read the webpack.config.js, but it won't work
   // for programmatic usage. This API is used by many loaders/plugins, so hope we're safe for a while
   const resolveOptionsFromWebpackConfig: webpack.ResolveOptions = this._compilation?.options.resolve || {};
 
   const resolveSync = enhancedResolve.create.sync({
-    ...resolveOptionsDefaults,
     alias: resolveOptionsFromWebpackConfig.alias,
     modules: resolveOptionsFromWebpackConfig.modules,
     plugins: resolveOptionsFromWebpackConfig.plugins,
+    conditionNames: [...(resolveOptionsFromWebpackConfig.conditionNames ?? []), 'require'],
```
ling1726, line 69: Yeah I can see the problem you'd have with the setup you described in the PR description. However, changing the behaviour to do a complete inheritance by default of resolve options will be a breaking change for existing users. Here's an example: 

The current set of extensions include `.json`, if this is not configured in the webpack config - it can break when Linaria tries to resolve modules that import json files. In this case even though webpack does not configure `json`in resolve options, it will not fail because it will prefer to use `import`. However `Linaria` will then blow up because it can't resolve a JSON file

In this case I'd propose two alternatives to support your use case:

1. add a loader option to select what resolve options users would like to inherit:

`inheritResolveOptions: ['alias', 'modules', 'plugins', 'conditionNames]`

This way inheritance is possible without requiring a major version bump of the package

2. Allow users to create their own `resolveSync` function which will achieve the same with maybe greater control over the resolver use by Linaria. https://github.com/microsoft/griffel/blob/bd37c4fe0b21b33c1d548f7abf1b8c58dd1e5280/packages/webpack-loader/src/webpackLoader.ts#L59-L74

In addition to inheriting resolve options, you'd also need to configure the evaluation rules  so that you transform the TS code to CJS which is required by Linaria https://github.com/microsoft/griffel/blob/main/packages/webpack-loader/README.md#configuring-module-evaluation

MLoughry: I've implemented the first suggestion.

packages/webpack-loader/README.md
```diff
@@ -157,6 +158,27 @@ module.exports = {
 };
 ```
 
+### Configuring webpack resolve options
+If your `@griffel/react` modules import other files (eg., a set of common mixins or colors for your app), the loader resolves these using `enhanced-resolve`. By default, it inherits the settings `resolve.alias`, `resolve.modules`, and `resolve.modules` from youe webpack config, while using its own default values for `resolve.extensions` and `resolve.conditionNames`. If you want to change this behavior, you can choose which `resolve` options are inherited from your webpack config.
```
layershifter, line 162:
```suggestion
If your `@griffel/react` modules import other files (eg., a set of common mixins or colors for your app), the loader resolves these using `enhanced-resolve`. By default, it inherits the settings `resolve.alias`, `resolve.modules`, and `resolve.modules` from your Webpack config, while using its own default values for `resolve.extensions` and `resolve.conditionNames`.

If you want to change this behavior, you can choose which [`resolve` options](https://webpack.js.org/configuration/resolve/) are inherited from your Webpack config.
```

nit

packages/webpack-loader/src/webpackLoader.ts
```diff
@@ -4,11 +4,25 @@ import * as path from 'path';
 import * as webpack from 'webpack';
 
 import { transformSync, TransformResult, TransformOptions } from './transformSync';
+import type { JSONSchema7 } from '@typescript-eslint/utils/dist/json-schema';
```
layershifter, line 7:
```suggestion
import type { JSONSchema7 } from 'json-schema';
```

packages/webpack-loader/src/webpackLoader.ts
```diff
@@ -4,11 +4,25 @@ import * as path from 'path';
...
+const optionsSchema: JSONSchema7 = {
+  ...configSchema,
+  properties: {
+    ...configSchema.properties,
+    inheritResolveOptions: {
+      type: 'array',
+      items: { type: 'string', enum: ['alias', 'modules', 'plugins', 'conditionNames', 'extensions'] },
+    },
+  },
+};
```
layershifter, line 24: nit: let's put it to `packages/webpack-loader/src/schemas.ts` file

change/@griffel-webpack-loader-b9e23b96-1848-4762-8b00-bb466cbbe407.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Inherit more resolve options from webpack config",
```
layershifter, line 3:
```suggestion
  "comment": "feat: Add 'inheritResolveOptions' to configure resolve options in Webpack",
```

layershifter (approved):
Let's also add a unit test for this case:

#### `packages/webpack-loader/__fixtures__/webpack-resolve-options/code.ts`

```ts
import color from './color';

export const styles = makeStyles({
  root: {
    backgroundColor: color,
  },
});
```

#### `packages/webpack-loader/__fixtures__/webpack-resolve-options/color.jsx`

```js
const color = 'blue';

export default color;
```

#### `packages/webpack-loader/__fixtures__/webpack-resolve-options/output.ts`

```ts
import color from './color';
export const styles = __styles(
  {
    root: {
      De3pzq: 'f1bh81bl',
    },
  },
  {
    d: ['.f1bh81bl{background-color:blue;}'],
  },
);
```

#### `packages/webpack-loader/src/webpackLoader.test.ts`

```diff
  // Asserts that aliases are resolved properly in Babel plugin
  testFixture('webpack-aliases', {
    webpackConfig: {
      resolve: {
        alias: {
          'non-existing-color-module': path.resolve(__dirname, '..', '__fixtures__', 'webpack-aliases', 'color.ts'),
        },
      },
    },
  });

+ // Asserts that "inheritResolveOptions" are handled properly
+ testFixture('webpack-resolve-options', {
+   loaderOptions: {
+     inheritResolveOptions: ['extensions'],
+   },
+   webpackConfig: {
+      resolve: {
+       extensions: ['.ts', '.jsx'],
+     },
+   },
+ });
```

layershifter (approved):

layershifter:
@MLoughry thanks, released in `@griffel/webpack-loader@2.1.21` 🚢

PR #418, opened 2023-08-11
eslint-plugin: Add autofix for no-shorthands

NotWoods:
When migrating from Fluent v8 to v9, replacing shorthands with the equivalent function is somewhat tedious. We can leverage the existing linter rule to autofix all shorthands!

I wasn't sure how to instruct the fixer to import `shorthands` but this is pretty easily resolved in the editor.


packages/eslint-plugin/src/rules/no-shorthands.ts
```diff
@@ -77,6 +77,64 @@ const UNSUPPORTED_CSS_PROPERTIES: Record<keyof CSS.StandardShorthandProperties,
...
+const splitShorthandIntoParts: ShorthandToArguments = value => {
+  if (typeof value !== 'string') {
+    if (value === 0) {
+      value = '0';
+    } else {
+      value = `${value}px`;
+    }
+  }
+  return value.match(CSS_FUNCTION_OR_VALUE);
+};
```
layershifter, line 98: Can you please move out this function to `utils` and add base unit tests for it?

packages/eslint-plugin/src/rules/no-shorthands.ts
```diff
@@ -77,6 +77,64 @@ const UNSUPPORTED_CSS_PROPERTIES: Record<keyof CSS.StandardShorthandProperties,
...
+
+// Transforms shorthand string into args for griffel shorthands.<name>() function.
+const SHORTHAND_FUNCTIONS: Partial<Record<keyof CSS.StandardShorthandProperties, ShorthandToArguments>> = {
```
layershifter, line 101:
```suggestion
const SHORTHAND_FUNCTIONS: Partial<Record<keyof typeof UNSUPPORTED_CSS_PROPERTIES, ShorthandToArguments>> = {
```

nit: let's reuse the type that we already have

layershifter (approved):
That's cool ❤️ 
Please do a rebase and address minor comments, then we should be good to go 🚢

layershifter:
Released in `@griffel/eslint-plugin@1.3.0` 🚢

PR #420, opened 2023-08-14
feat: experimental code splitting for CSS

### LINKED ISSUE #419: webpack-extraction-plugin: support for code split CSS
Currently CSS extraction forces all styles to be inside one style sheet (one file). That simplifies the implementation, but we want to be able to only load the CSS that is used in loaded modules instead of loading whole CSS from a project.

There main constraint is that all stylesheets need to be sorted in the same way as in a single file (https://griffel.js.org/react/guides/atomic-css#lvfha-order-of-pseudo-classes) i.e. follow the approach with style buckets.

This means that only CSS from `d` bucket is safe, all other buckets are considered unsafe.

### Why is this hoisting required?

If the order is not consistent ➡️ applied CSS won't be consistent and results for different experiences are dependant on what chunks were loaded and when. Here is an example:

```css
/** css-1.css **/

.baz { color: "green" }
.baz:active { color: "magenta" }
```

```css
/** css-2.css **/
.baz:hover { color: "pink" }
```

```html
<link rel="./css-1.css" />
<link rel="./css-2.css" />
```

In this example the resulting order will be:

```css
.baz { color: "green" }
.baz:active { color: "magenta" }
.baz:hover { color: "pink" } /** 💣 :hover is later than :active **/
```

While it **should** be:

```css
.baz { color: "green" }
.baz:hover { color: "pink" }
.baz:active { color: "magenta" } /** ✅ :active is after :hover **/
```

layershifter:
Fixes #419.


packages/webpack-extraction-plugin/src/constants.ts
```diff
@@ -4,8 +4,8 @@ export const PLUGIN_NAME = 'GriffelExtractPlugin';
 export const GriffelCssLoaderContextKey = Symbol(`${PLUGIN_NAME}/GriffelCssLoaderContextKey`);
 
 export interface GriffelLoaderContextSupplement {
-  registerExtractedCss(css: string): void;
-  getExtractedCss(): string;
+  createCSSImport(type: 'safe' | 'unsafe', css: string): string;
+  getCSSOutput(type: 'safe' | 'unsafe'): string;
```
layershifter, line 8: Revert

packages/webpack-extraction-plugin/src/parseCSSRules.ts
```diff
@@ -17,6 +17,16 @@ export function parseCSSRules(css: string) {
...
     if (element.type === COMMENT) {
+      // TODO: better handling of comments
+
+      if (element.value.indexOf('/** @griffel:css-safe') === 0) {
+        continue;
+      }
+
+      if (element.value.indexOf('/** @griffel:css-unsafe') === 0) {
+        continue;
+      }
```
layershifter, line 28: Revert

packages/webpack-extraction-plugin/src/webpackLoader.ts
```diff
@@ -76,7 +78,20 @@ function webpackLoader(
...
+            `/** @griffel:css-end **/`,
+            '',
+          ].join('\n')
+        : '';
+
+      const unsafeCSS = entries.reduce((acc, [cssBucketName, cssBucketRules]) => {
+        if (cssBucketName === 'd') {
+          return acc;
+        }
+
```
layershifter, line 94: Revert

packages/webpack-extraction-plugin/src/webpackLoader.ts
```diff
@@ -48,6 +46,10 @@ function webpackLoader(
     return;
   }
 
+  if (!this[GriffelCssLoaderContextKey]) {
+    throw new Error('GriffelCSSExtractionPlugin is not configured, please check your webpack config');
+  }
+
```
layershifter, line 52: Ship to master

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -182,16 +200,29 @@ export class GriffelCSSExtractionPlugin {
...
+            );
+
+            return `import ${stringifiedRequest};\n`;
           },
-          getExtractedCss() {
-            const css = cssByModuleMap.get(resourcePath) ?? '';
-            cssByModuleMap.delete(resourcePath);
+          getCSSOutput(type: 'safe' | 'unsafe') {
+            const css = cssByModuleMap.get(loaderContext.resourcePath + ':' + type) ?? '';
+            cssByModuleMap.delete(loaderContext.resourcePath + ':' + type);
```
layershifter, line 225: Revert

packages/webpack-extraction-plugin/virtual-loader/index.js
```diff
@@ -5,7 +5,9 @@ const { GriffelCssLoaderContextKey } = require('../src/constants');
  * @return {String}
  */
 function virtualLoader() {
-  return this[GriffelCssLoaderContextKey]?.getExtractedCss() ?? '';
+  const type = /** @type {'safe'|'unsafe'} */ (this.resourceQuery.slice(1));
+
+  return this[GriffelCssLoaderContextKey]?.getCSSOutput(type) ?? '';
```
layershifter, line 10: Revert

PR #421, opened 2023-08-15
Allow developers to specify custom resolve options.

MLoughry:
Integrating my last change in #412, I ran into an issue because `@linaria/babel-preset` (at least at its current version of `3.0.0-beta.23`) throws an error when given an ESM module. Since we don't want to use CJS as the basis of our whole build, we need to specify custom `conditionNames` for this loader in particular.


MLoughry:
@layershifter When you get a chance, can you please review?

packages/webpack-loader/src/webpackLoader.test.ts
```diff
@@ -269,6 +269,18 @@ describe('webpackLoader', () => {
...
+  // Asserts that "webpackResolveOptions" are handled properly
+  testFixture('webpack-resolve-options', {
+    loaderOptions: {
+      webpackResolveOptions: { conditionNames: ['source', 'require'] },
+    },
+    webpackConfig: {
+      resolve: {
+        conditionNames: ['source', 'import'],
+      },
+    },
```
layershifter, line 281:
```suggestion
    loaderOptions: {
      webpackResolveOptions: {
        extensions: ['.ts', '.jsx'],
      },
    },
```

Let's make the assertion real as we don't have `package.json` files in this fixture to test against.

change/@griffel-webpack-loader-741e9ad7-6c83-458e-9add-a8541efa0939.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add the ability to specify custom webpack resolve options",
```
layershifter, line 3:
```suggestion
  "comment": "feat: add webpackResolveOptions to specify custom resolve options",
```

layershifter (approved):

layershifter:
Released in `@griffel/webpack-loader@2.2.0` 🚢

PR #426, opened 2023-08-22
Stop inlining the CSS in the virtual CSS module request

MLoughry:
This change should have no impact on production build output. However, one issue I've noticed hen trying to migrate Outlook to use this is that our bundle-diff tool gets unwieldy with the massive module requests that embed the entire extracted CSS into the request.

Instead, this change stores the extracted CSS in a map in the plugin, and uses an enhancement to the loader context to read and write the CSS in the map. As a result, we can eliminate the `?styles=<css>` part of the module request.

### DIFF
--- a/change/@griffel-webpack-extraction-plugin-2722342d-e7ad-42c7-b370-cc43d50217de.json
+++ b/change/@griffel-webpack-extraction-plugin-2722342d-e7ad-42c7-b370-cc43d50217de.json
@@ -1,6 +1,6 @@
 {
-  "type": "minor",
-  "comment": "Stop inlining the styles in the virtual CSS module request",
+  "type": "patch",
+  "comment": "chore: stop inlining the styles in the virtual CSS module request",
   "packageName": "@griffel/webpack-extraction-plugin",
   "email": "miclo@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@griffel-webpack-extraction-plugin-2722342d-e7ad-42c7-b370-cc43d50217de.json
+++ b/change/@griffel-webpack-extraction-plugin-2722342d-e7ad-42c7-b370-cc43d50217de.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Stop inlining the styles in the virtual CSS module request",
+  "packageName": "@griffel/webpack-extraction-plugin",
+  "email": "miclo@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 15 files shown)


packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -177,6 +176,22 @@ export class GriffelCSSExtractionPlugin {
...
+      NormalModule.getCompilationHooks(compilation).loader.tap(PLUGIN_NAME, (loaderContext, module) => {
+        (loaderContext as SupplementedLoaderCotext)[GriffelCssLoaderContextKey] = {
+          registerExtractedCss(css: string) {
+            cssByModuleMap[module.resource] = css;
+          },
+          getExtractedCss() {
+            return cssByModuleMap[module.resource] ?? '';
+          },
+        };
+      });
```
layershifter, line 193:
```suggestion
      const cssByModuleMap = new Map<string, string>();

      NormalModule.getCompilationHooks(compilation).loader.tap(PLUGIN_NAME, (loaderContext, module) => {
        const resourcePath = module.resource;

        (loaderContext as SupplementedLoaderCotext)[GriffelCssLoaderContextKey] = {
          registerExtractedCss(css: string) {
            cssByModuleMap.set(resourcePath, css);
          },
          getExtractedCss() {
            const css = cssByModuleMap.get(resourcePath) ?? '';
            cssByModuleMap.delete(resourcePath);

            return css;
          },
        };
      });
```

Let's use there `Map` and dispose it once CSS was used.

change/@griffel-webpack-extraction-plugin-2722342d-e7ad-42c7-b370-cc43d50217de.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Stop inlining the styles in the virtual CSS module request",
```
layershifter, line 3:
```suggestion
  "type": "patch",
  "comment": "chore: stop inlining the styles in the virtual CSS module request",
```

nit: it's not adding any functionality

layershifter (commented):
LGTM, I will do a round of testing in products and will come back with outcome.

layershifter:
This PR can cause chunks to shuffle, that's fine as there are no changes in generated output. Changes can happen due different size of modules (now size is smaller). Examples are below:

### Before
image


### After

image

---

Sources are not modified, changes in size are only due shorter imports.

layershifter (approved):

PR #432, opened 2023-09-04
fix(style-types): allow CSS variables in animationName

### LINKED ISSUE #417: Animation types have errors when animating CSS custom properties
When animating CSS custom properties Typescript raises an error:

> Type '{ "--my-var": string; }' is not assignable to type 'GriffelResetStylesCSSProperties'.
  Object literal may only specify known properties, and '"--my-var"' does not exist in type 'GriffelResetStylesCSSProperties'.ts(2322)

This issue affects both `makeStyles` and `makeResetStyles`. I believe it originates from #413 as this code worked prior to the type change.

Repro: https://codesandbox.io/s/griffel-css-animation-myhrq4?file=/src/index.tsx

NOTE: repro will only run in a Chrome and Edge at the time of this writing (and [maybe Safari](https://ishoudinireadyyet.com/), but haven't tested there).

This code is taking advantage of the fact that [registered custom properties are animatable](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_animated_properties#animating_custom_properties). Technically custom properties can be used in animations without registering them, but animations are discrete in that case.

Would be nice for Griffel types to support custom properties in animations.

layershifter:
Fixes an issue with `animationName` (fixes #417) to allow usage of CSS variables in `animationName`. The fix is done for modern typings as legacy typings already allow it.

To make it strict, new syntax with template literals. Legacy types now used for TS 4.3 and earlier.

### DIFF
--- a/packages/style-types/src/makeStyles.ts
+++ b/packages/style-types/src/makeStyles.ts
@@ -27,5 +27,8 @@ type GriffelCSSPseudos = {
   [Property in CSS.Pseudos]?: GriffelStylesStrictCSSObject | GriffelCSSObjectCustom;
 };
 
-export type GriffelAnimation = Record<'from' | 'to' | string, GriffelStylesCSSProperties>;
+export type GriffelAnimation = Record<
+  'from' | 'to' | string,
+  GriffelStylesCSSProperties & { [Property in `--${string}`]: string }
+>;
 export type GriffelStyle = GriffelStylesStrictCSSObject | GriffelCSSObjectCustom;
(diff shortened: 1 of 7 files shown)


packages/style-types/src/makeStyles.ts
```diff
@@ -27,5 +27,8 @@ type GriffelCSSPseudos = {
   [Property in CSS.Pseudos]?: GriffelStylesStrictCSSObject | GriffelCSSObjectCustom;
 };
 
-export type GriffelAnimation = Record<'from' | 'to' | string, GriffelStylesCSSProperties>;
+export type GriffelAnimation = Record<
+  'from' | 'to' | string,
+  GriffelStylesCSSProperties & { [Property in `--${string}`]: string }
```
ling1726, line 32: These changes will only support css variables in animation properties for non-legacy types correct?

ling1726 (approved):

Issue #424, opened 2023-08-18
How to combine class strings + griffel atomic classes


benkeen:
Simple scenario + easy to solve, I just wasn't sure of the best-practice for this use-case. 

My (existing) component accepts a `className` prop which is a string potentially containing potentially multiple classes. My component is using Griffel and constructs its own classes. 

What's the recommended way to combine them?

spmonahan:
I think you want [`mergeClasses()`](https://griffel.js.org/react/api/merge-classes).

I put together an example based on your description: https://codesandbox.io/s/mergeclasses-example-48dk65

benkeen:
Oh! Apologies, I didn't think that method permitted that. Terrific, thanks for the help :)

spmonahan:
Glad it worked!

@layershifter this might be a good example to add to the docs. What do you think?

layershifter:
@spmonahan yes, we should add a note to `mergeClasses()` docs about this (https://griffel.js.org/react/api/merge-classes). @benkeen @spmonahan would you like to submit a PR? 🐱

benkeen:
Sure, np! I can knock something up tonight. :)

PR #433, opened 2023-09-05
Update mergeClasses doc to explain combining griffel classes with custom classes

benkeen:
Small doc update to provide an explicit example of combining the generated scoped classes from Griffel with arbitrary space-delimited class strings. 

As per discussion here: https://github.com/microsoft/griffel/issues/424

### DIFF
--- a/apps/website/docs/react/api/merge-classes.md
+++ b/apps/website/docs/react/api/merge-classes.md
@@ -112,6 +112,25 @@ function App(props) {
 }
 ```
 
+### Combining Griffel classes with custom class strings
+
+You may find yourself in a position where your component accepts classes via props or have global classes that you need applied to the same elements being styled by Griffel. For this, you can also use `mergeStyles`: 
+
+```jsx
+const useClasses = makeStyles({
+  root: { color: 'red' }
+});
+
+function App(props) {
+  const classes = useClasses();
+  const rootClass = mergeClasses(classes.root, "custom-global-class1 custom-global-class2");
+
+  return (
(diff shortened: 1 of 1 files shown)


apps/website/docs/react/api/merge-classes.md
```diff
@@ -112,6 +112,25 @@ function App(props) {
...
+
+function App(props) {
+  const classes = useClasses();
+  const rootClass = mergeClasses(classes.root, "custom-global-class1 custom-global-class2");
+
+  return (
+     <div className={rootClass}>...</div>
+  )
+}
+```
```
layershifter, line 132: I just realized that we have already an example for this case above "Incorrect usages":

image

I suggest to modify that section and add above it:

```
### Merging plain CSS classes

Your component may accept classes via props or have static classes that you need applied to the same elements being styled by Griffel. For this, you can also use `mergeClasses`:
```

WDYT?

benkeen: Sounds good!

layershifter: @benkeen would you be able to do the change? 🐱

benkeen: Yup absolutely! Sorry - last night was busy. Will update it tonight.

benkeen: Updated. Definitely better, thanks!

layershifter (approved):

PR #434, opened 2023-09-05
feat: integrate stylelint to css-lint rule

ling1726:
Follow up of #430 and integrates stylelint to lint atomic css rules in eslint.

Uses [synckit](https://github.com/un-ts/synckit) to make the styleint calls synchronous. The method behind synckit [is explained a bit here](https://samthor.au/2021/block-nodejs-main-thread/)

Exports the `css-lint` rule as unstable for now


packages/eslint-plugin/package.json
```diff
@@ -12,9 +12,11 @@
     "@babel/core": "^7.12.13",
     "@typescript-eslint/utils": "^5.47.0",
     "csstype": "^3.1.2",
+    "synckit": "^0.8.5",
     "tslib": "^2.1.0"
   },
   "peerDependencies": {
-    "eslint": "^7.5.0 || ^8.0.0"
+    "eslint": "^7.5.0 || ^8.0.0",
+    "stylelint": "15.x"
```
layershifter, line 20: This probably should be a direct dependency, actually.

packages/eslint-plugin/src/rules/css-lint/.stylelintrc.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * A test stylelint config for tests
+ */
+
+module.exports = {
+  rules: {
+    'selector-anb-no-unmatchable': [true],
```
layershifter, line 7: "anb"?

packages/eslint-plugin/src/rules/css-lint/.stylelintrc.js
```diff
@@ -0,0 +1,9 @@
+/**
```
layershifter, line 1: Let's name this file directly to include `-test`. Also please check that it does not leak to `dist` folder.

packages/eslint-plugin/src/rules/css-lint/.stylelintrc.js
layershifter: Edit: i see, it's a valid rule.

layershifter (approved):

packages/eslint-plugin/package.json
ling1726: well it's probably never going to be put into a bundle, so I don't mind

PR #435, opened 2023-09-07
fix(core): trim ">" selector

### LINKED ISSUE #431: Selector `>` and `> ` should generate same class name
### Actual behavior:
[griffel playground](https://griffel.js.org/try-it-out/?code=JYWwDg9gTgLgBAbziAhgawKYGUYE8A2GAzgDRxEAW0MFKAdgCZFwC%2BcAZlBCHAOQACAcyjB27DPgD0AY2gZeAbgBQSjAA9IsOAwzsUAV3zxUmHAWIAKBErhwuEGAC5EN23wBkAPgB0MYjF5nJFl8aGdeYQwMOl5WEldbXi84X39AxDgQsL5I6NiWVxZ4lgBKZSA):
 
image

Note that `>` and `> ` selector generates different classNames.

### Expected behavior: 
Note that `>` and `> ` selector generates the same className

layershifter:
Fixes #431.

This PR updates hash generation for properties and classes to trim selectors before hash will be generated. Currently only the case with `> ` is handled.

New tests added for matching scenarios.

### DIFF
--- a/packages/core/src/runtime/utils/trimSelector.ts
+++ b/packages/core/src/runtime/utils/trimSelector.ts
@@ -1,0 +1,6 @@
+/**
+ * Trims selectors to generate consistent hashes.
+ */
+export function trimSelector(selector: string): string {
+  return selector.replace(/>\s+/g, '>');
+}
--- a/change/@griffel-core-514d4515-c4ba-47ce-81d3-9c6c3426e82e.json
+++ b/change/@griffel-core-514d4515-c4ba-47ce-81d3-9c6c3426e82e.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix(core): trim \">\" selector",
+  "packageName": "@griffel/core",
+  "email": "olfedias@microsoft.com",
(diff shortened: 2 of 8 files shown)


packages/core/src/runtime/utils/trimSelector.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Trims selectors to generate consistent hashes.
+ */
+export function trimSelector(selector: string): string {
```
ling1726, line 4: for me it's a little bit more than a nit since what you're doing here isn't actually trimming. Youre're specifically targeting the spaces after the `>` combinator.

I'd rename this to `stripExtraWhitespace` and add some extra documentation why that specifically is necessary

ling1726 (approved):

packages/core/src/runtime/utils/trimSelector.ts
layershifter: Potentially we could have more cases, that's why I went with more general naming.

PR #437, opened 2023-09-08
test(postcss-syntax): add e2e wiring @griffel/postcss-syntax into stylelint

ling1726:
Adds an `e2e/stylelint` project that validates the [`@griffel/postcss-syntax`](https://github.com/microsoft/griffel/tree/main/packages/postcss-syntax) + [stylelint](https://stylelint.io/) integration end to end.

The test packs the local `@griffel/postcss-syntax` and `@griffel/babel-preset` packages, installs `stylelint` into a temporary project, configures it with the Griffel PostCSS [custom syntax](https://stylelint.io/developer-guide/syntaxes), and lints `*.styles.ts` fixtures covering three cases:

- **`green.styles.ts`** — a valid file reports no problems.
- **`error.styles.ts`** — an unmatchable `:nth-child(0)` selector is reported via stylelint's built-in `selector-anb-no-unmatchable` rule.
- **`disabled.styles.ts`** — the same violation is suppressed through a `griffel-csslint-disable` comment directive (emitted as `stylelint-disable-line`).

Run it with:

```sh
yarn nx run @griffel/e2e-stylelint:test
```

### DIFF
--- a/e2e/stylelint/README.md
+++ b/e2e/stylelint/README.md
@@ -1,0 +1,22 @@
+# @griffel/e2e-stylelint
+
+End-to-end test that verifies [`@griffel/postcss-syntax`](../../packages/postcss-syntax) can be
+wired into [stylelint](https://stylelint.io/) as a
+[custom syntax](https://stylelint.io/developer-guide/syntaxes) to lint Griffel CSS-in-JS files.
+
+The test packs the local `@griffel/postcss-syntax` and `@griffel/babel-preset` packages, installs
+`stylelint` into a temporary project, and runs `stylelint` (configured with the Griffel custom
+syntax) against `*.styles.ts` fixtures:
+
+- **`green.styles.ts`** — a valid file, asserts stylelint reports no problems.
+- **`error.styles.ts`** — a `makeStyles` call with an unmatchable `:nth-child(0)` selector, asserts
+  stylelint reports the built-in `selector-anb-no-unmatchable` rule for the generated CSS.
+- **`disabled.styles.ts`** — the same violation, but suppressed with a `griffel-csslint-disable`
+  comment directive; asserts the directive emits a `stylelint-disable-line` and stylelint reports no
+  problems.
+
+Run it with:
+
+```sh
+yarn nx run @griffel/e2e-stylelint:test
+```
--- a/e2e/stylelint/eslint.config.mjs
+++ b/e2e/stylelint/eslint.config.mjs
@@ -1,0 +1,16 @@
+import baseConfig from '../../eslint.config.mjs';
+
+export default [
+  {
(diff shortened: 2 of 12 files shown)


packages/postcss-syntax/src/parse.ts
```diff
@@ -5,16 +5,18 @@ import { BabelPluginOptions } from '@griffel/babel-preset';
...
 
-export interface ParserOptions extends PostCSSParserOptions, BabelPluginOptions {}
+export interface ParserOptions extends PostCSSParserOptions {
+  griffelPreset?: BabelPluginOptions;
+}
 
 export const parse = (css: string | { toString(): string }, opts?: ParserOptions) => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  const { from: filename = 'postcss-syntax.styles.ts', map, ...griffelPluginOptions } = opts ?? {};
+  const { from: filename = 'postcss-syntax.styles.ts', griffelPreset } = opts ?? {};
```
ling1726, line 13: postcss passes some untyped properties here - it will fail the griffel config validator, so I've put it all the preset configuration in a separate property

packages/postcss-syntax/src/stringify.ts
```diff
@@ -4,7 +4,6 @@ import { GRIFFEL_SRC_RAW } from './constants';
 export const stringify: postcss.Stringifier = root => {
   const originalSource = root.raw(GRIFFEL_SRC_RAW);
   if (originalSource) {
-    console.log(originalSource);
```
ling1726, line 7: oopsie

packages/stylelint-config/src/stylelint.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    const { results } = await stylelint.lint({ code, config });
+    const warnings = results.map(result => result.warnings).flat();
+    expect(warnings).toMatchInlineSnapshot(`
+      Array [
+        Object {
+          "column": 11,
+          "endColumn": 24,
+          "endLine": 5,
+          "line": 5,
```
ling1726, line 35: The starting line is always correct, however the rest of the locations are wrong because most of the stylelint default rules calculate additional offsets internally for better reporting when linting real CSS files.

This doesn't work too well with the griffel postcss syntax since we lose the mapping to the entire style slot. Example:

https://github.com/stylelint/stylelint/blob/0b686f289ab436eac64aaa78eb1646d38a153294/lib/rules/selector-anb-no-unmatchable/index.js#L91C8-L100

That rule even uses a third party parser for the selector and doesn't really use postcss at all

layershifter (dismissed):

layershifter (approved):

PR #450, opened 2023-10-06
feat: add support for style resets in makeStyles

### LINKED ISSUE #442: Support removing property from existing set of styles
If a component is already setting the font-family, there is no way to completely remove the font-family property. The only option is to set the font-family to a new, valid value. There is currently no way to completely remove the font-family property.

layershifter:
Fixes #442.

This PR adds support for `RESET` values in `makeStyles()` calls, for example:

```tsx
import { makeStyles, RESET } from '@griffel/react'

  root: {
    color: RESET,
  }
});

function App() {
  // 💡 "classes.root" will be an empty string
  const classes = useClasses();

  return <div className={classes.root} />
}
```

In this case, nothing will be applied to a `div`. This feature becomes useful for merging cases, for example:

```tsx
import { makeStyles, mergeClasses, RESET } from '@griffel/react';

const useClassesA = makeStyles({
  root: {
    color: 'red',
    backgroundColor: 'blue',
  },
});
const useClassesB = makeStyles({
  root: {
    color: RESET,
  },
});

function Component() {
  const classesA = useClassesA();
  const classesB = useClassesB();

  // 💡 After merging, the `color` property will be removed
  const className = mergeClasses(classesA.root, classesB.root);

  // 💡 If `RESET` value is set, it won't be added to the class
  return <div className={classesA} />;
}
```

### Note on `makeStyles()` & `makeResetStyles()` combo

```tsx
import { makeStyles, makeResetStyles, mergeClasses, RESET } from '@griffel/react';

const useClassesA = makeResetStyles({
  color: 'red',
  backgroundColor: 'blue',
});
const useClassesB = makeStyles({
  root: {
    color: RESET,
  },
});

function Component() {
  const classesA = useClassesA();
  const classesB = useClassesB();

  // 🔴  After merging, the `color` property from `makeResetStyles()` will be still applied
  const className = mergeClasses(classesA, classesB.root);
}
```

`RESET` values work only for `makeStyles()` calls i.e. it won't remove/reset values created by `makeResetStyles()`.


packages/babel-preset/src/transformPlugin.ts
```diff
@@ -126,7 +126,15 @@ function buildCSSEntriesMetadata(
     Object.entries(classnamesMapping).map(([slot, cssClassesMap]) => {
       return [
         slot,
-        Object.values(cssClassesMap).flatMap(cssClasses => (Array.isArray(cssClasses) ? cssClasses : [cssClasses])),
+        Object.values(cssClassesMap).reduce<string[]>((acc, cssClasses) => {
+          if (typeof cssClasses === 'string') {
+            acc.push(cssClasses ?? '');
```
ling1726, line 131:
```suggestion
            acc.push(cssClasses);
```

ling1726, line 131:
```suggestion
            acc.push(cssClasses);
```

this shouldn't be necessary

packages/babel-preset/src/transformPlugin.ts
```diff
@@ -143,6 +151,10 @@ function buildCSSEntriesMetadata(
       return [
         slot,
         cssClasses.map(cssClass => {
+          if (cssClass === null) {
+            return '';
+          }
```
ling1726, line 156:
```suggestion
```

This shouldn't be necessary, the `classesBySlot` object is created to ignore null values

apps/website/docs/react/api/make-styles.md
```diff
@@ -260,6 +260,37 @@ const useClasses = makeStyles({
 });
 ```
 
+### `null` values
```
ling1726, line 263: There should also be a playground for this example

packages/core/src/runtime/resolveStyleRules.ts
```diff
@@ -88,8 +88,16 @@ export function resolveStyleRules(
 
     const value = styles[property as keyof GriffelStyle];
 
-    // eslint-disable-next-line eqeqeq
-    if (value == null) {
+    if (typeof value === 'undefined') {
+      continue;
+    }
+
+    if (value === null) {
```
ling1726, line 95: As discussed offline, let's try to use a special symbol so we don't start introducing different behaviour for undefined vs null

apps/website/docs/react/api/make-styles.md
layershifter: As it relies on `mergeClasses()`, I am unsure about playground example

packages/core/src/runtime/resolveStyleRules.ts
layershifter: @ling1726 after discussion with @george-cz, I went with a a string casted to a CSS value.

ling1726 (approved):

PR #451, opened 2023-10-07
update make-static-styles.md

### LINKED ISSUE #443: docs: makeStaticStyles doesn't mention it returns a React hook
I've seen a lot of developer confusion around the fact that Griffel creates React hooks. Most developers don't realize it produces hooks. The lint rule helps, but now some developers think Griffel simply creates things that need the name "use" and still aren't hooks.

[`makeStaticStyles`](https://griffel.js.org/react/api/make-static-styles) doesn't mention that it creates a React hook, so some devs on our team were tripped up and assumed it broke the pattern and did not return a React hook.

It should probably also mention if shorthands are OK or not.

sanketmp:
fixes #443 
- Mentioned that `makeStaticStyles` returns a React hook.
#

Note:
I couldn't figure out if `makeStaticStyles` supports shorthands or not. I have tried this on my codesandbox and it was not working. If I am going the wrong way or if it's expected behavior that shorthands are not Okay. Then I will add this as well. Please review.
Ref: https://codesandbox.io/s/shorthands-check-n342ls

### DIFF
--- a/apps/website/docs/react/api/make-static-styles.md
+++ b/apps/website/docs/react/api/make-static-styles.md
@@ -8,6 +8,8 @@ import OutputTitle from '@site/src/components/OutputTitle'
 
 Creates styles with a global selector. This is especially useful for CSS resets, for example [normalize.css](https://github.com/necolas/normalize.css/).
 
+`makeStaticStyles` returns [a React hook](https://reactjs.org/docs/hooks-intro.html) that should be called inside a component.
+
 ## Defining styles with objects
 
 ```js


sanketmp:
@layershifter I have added a note about shorthands. Please review.

and what is CLA bot about?

apps/website/docs/react/api/make-static-styles.md
```diff
@@ -100,3 +100,44 @@ function App() {
...
+      <button
+        className={mergeClasses(
+          props.primary && classes.primary,
+          props.circular && classes.circular
+        )}
+      />
+    </>
+  );
+}
+```
```
layershifter, line 143:
````suggestion
```tsx
import { makeStyles, makeStaticStyles, shorthands } from '@griffel/react';

const useStaticStyles = makeStaticStyles({
  body: {
    color: 'red',
    padding: '5px',
  },
});

  primaryText: {
    color: 'blue',
    ...shorthands.padding('5px'),
  },
});

export default function App(props) {
  useStaticStyles();
  const classes = useClasses();

  return <p className={props.primaryText}>Hello world</p>;
}
```
````

Let's simplify the example to not include `mergeClasses()` as it's not related to the problem :)

**Comment by layershifter** [comment]
> and what is CLA bot about?

You can find all details in the comment https://github.com/microsoft/griffel/pull/451#issuecomment-1755476031.

**Review by layershifter — APPROVED** [review]

File: apps/website/docs/react/api/make-static-styles.md (hunk 1 above)
Comment (sanketmp) [reply apps/website/docs/react/api/make-static-styles.md:143]: Okay. 👍

**Comment by sanketmp** [comment]

File: apps/website/docs/react/api/make-static-styles.md [hunk 2]
```diff
@@ -98,3 +100,32 @@ function App() {
...
+
+`makeStaticStyles` supports the use of [shorthands](/react/api/shorthands) and does not have the same limitation as `makeStyles()`.
+
+:::
+
+## Usage with `makeStyles`
+
+```tsx
+import { makeStyles, makeStaticStyles, shorthands } from '@griffel/react';
+const useStaticStyles = makeStaticStyles({
```
layershifter, line 114:
```suggestion
import { makeStyles, makeStaticStyles, shorthands } from '@griffel/react';

const useStaticStyles = makeStaticStyles({
```

apps/website/docs/react/api/make-static-styles.md
```diff
@@ -98,3 +100,32 @@ function App() {
...
+  body: {
+    color: 'red',
+    padding: '5px',
+  },
+});
+const useClasses = makeStyles({
```
layershifter, line 120:
```suggestion
});

const useClasses = makeStyles({
```

apps/website/docs/react/api/make-static-styles.md
```diff
@@ -98,3 +100,32 @@ function App() {
...
+  primaryText: {
+    color: 'blue',
+    ...shorthands.padding('5px'),
+  },
+});
+export default function App(props) {
```
layershifter, line 126:
```suggestion
});

export default function App(props) {
```

PR #454, opened 2023-10-17
feat(webpack-extraction-plugin): add compat for Rspack

layershifter:
This PR updates `@griffel/webpack-extraction-plugin` to support Rspack.

----

### Changes to E2E

- New E2E added for Rspack
- Improved snapshot diff in E2E tests

### Changes to `@griffel/webpack-loader`

No changes need, works as it is

### Changes to `@griffel/webpack-extraction-plugin`

All specific places are marked with `@ Rspack compat` comment.

The biggest problem is that Rspack does not support functions for `test` in `splitChunks.cacheGroups`. Because of that we can't detect generated CSS modules via `isCSSModule()` 😭 That's the reason why I have switched for Rspack to the previous approach with a fake CSS file as then we can use a regex for `test` assertion. All other changes are related this problem.


packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
```diff
@@ -134,58 +142,85 @@ export class GriffelCSSExtractionPlugin {
...
+      }
+
+      // WHAT?
+      //  Disconnects Griffel chunk from other chunks, so Griffel chunk cannot be loaded async. Also connects with
+      //  the main entrypoint in config.
+      // WHY?
+      //  This is scenario required by one of MS teams. Will be removed in the future.
+      if (this.attachToMainEntryPoint) {
+        // @ Rspack compat
+        // We don't support this scenario for Rspack yet.
```
ling1726, line 210: wouldn't this mean that with rspack the extracted css can be loaded async and cause style flashing?

ling1726 (approved):

packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
layershifter: Nope, that trick is a specific for Teams and allows to load async instead as it breaks connections with other chunks.

PR #463, opened 2023-11-08
feat(core): add prefixing for "backdrop-filter"

### LINKED ISSUE #460: feature: add prefixing for backdrop-filter
`backdrop-filter` is supported in Safari only with a `-webkit` prefix (https://caniuse.com/?search=backdrop-filter), we need to add prefixing for it.

layershifter:
Fixes #460.

### DIFF
--- a/packages/core/src/runtime/stylis/prefixerPlugin.ts
+++ b/packages/core/src/runtime/stylis/prefixerPlugin.ts
@@ -37,6 +37,7 @@ export function prefix(value: string, length: number, children?: Element[]): str
     // backdrop-filter, background-clip: text
     case 4215:
       // Additional check on "backdrop-(f)ilter" to prevent conflict with "background-clip"
+      // https://github.com/thysultan/stylis/issues/292
       if (charat(value, 9) === 102) {
         return WEBKIT + value + value;
       }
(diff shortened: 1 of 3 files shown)


ling1726 (approved):

packages/core/src/runtime/stylis/prefixerPlugin.ts
```diff
@@ -34,11 +34,16 @@ export function prefix(value: string, length: number, children?: Element[]): str
     case 4599:
     case 4855:
       return WEBKIT + value + value;
-    // background-clip: text
+    // backdrop-filter, background-clip: text
     case 4215:
-      switch (charat(value, length + 1)) {
-        case 116:
-          return WEBKIT + value + value;
+      // Additional check on "backdrop-(f)ilter" to prevent conflict with "background-clip"
```
ling1726, line 39: please link the issue from stylis

layershifter, line 39:
```suggestion
      // Additional check on "backdrop-(f)ilter" to prevent conflict with "background-clip"
      // https://github.com/thysultan/stylis/issues/292
```

PR #480, opened 2024-01-08
feat(shadow-dom): add `insertionPoint` to `createShadowDOMRenderer()` options

spmonahan:
Exposes the `insertionPoint` API for shadow DOM similar to vanilla Griffel.

[See Fluent UI Contrib for more details](https://github.com/microsoft/fluentui-contrib/issues/59).

### DIFF
--- a/packages/shadow-dom/src/createShadowDOMRenderer.ts
+++ b/packages/shadow-dom/src/createShadowDOMRenderer.ts
@@ -3,7 +3,7 @@ import type { GriffelRenderer, StyleBucketName } from '@griffel/core';
 
 import { createFallbackRenderer } from './createFallbackRenderer';
 import type { ExtendedCSSStyleSheet, GriffelShadowDOMRenderer } from './types';
-import { findInsertionPoint } from './findInsertionPoint';
+import { findInsertionPoint, findShadowRootInsertionPoint } from './findInsertionPoint';
 
 const SUPPORTS_CONSTRUCTABLE_STYLESHEETS: boolean = (() => {
   try {
@@ -18,7 +18,7 @@ export interface CreateShadowDomRendererOptions {
   /**
    * If specified, a renderer will insert created CSSStyleSheets after this CSSStyleSheet.
    */
-  insertionPoint?: ExtendedCSSStyleSheet;
+  insertionPoint?: CSSStyleSheet;
 }
 
 let rendererId = 0;
@@ -92,10 +92,20 @@ export function createShadowDOMRenderer(shadowRoot: ShadowRoot, options: CreateS
             metadata,
 
             styleSheet => {
-              const targetStyleSheet = findInsertionPoint(renderer, styleSheet, insertionPoint);
+              const targetStyleSheet = findInsertionPoint(renderer, styleSheet);
+              const shadowRootTargetSheet = findShadowRootInsertionPoint(
+                renderer,
+                shadowRoot,
+                targetStyleSheet,
+                insertionPoint,
+              );
 
               renderer.adoptedStyleSheets = insertBefore(renderer.adoptedStyleSheets, styleSheet, targetStyleSheet);
-              shadowRoot.adoptedStyleSheets = insertBefore(shadowRoot.adoptedStyleSheets, styleSheet, targetStyleSheet);
+              shadowRoot.adoptedStyleSheets = insertBefore(
+                shadowRoot.adoptedStyleSheets,
+                styleSheet,
+                shadowRootTargetSheet,
+              );
             },
           );
 
--- a/packages/shadow-dom/src/createShadowDOMRenderer.ts
+++ b/packages/shadow-dom/src/createShadowDOMRenderer.ts
@@ -93,12 +93,7 @@ export function createShadowDOMRenderer(shadowRoot: ShadowRoot, options: CreateS
 
             styleSheet => {
               const targetStyleSheet = findInsertionPoint(renderer, styleSheet);
-              const shadowRootTargetSheet = findShadowRootInsertionPoint(
-                renderer,
-                shadowRoot,
-                targetStyleSheet,
-                insertionPoint,
-              );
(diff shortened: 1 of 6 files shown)


packages/shadow-dom/src/createShadowDOMRenderer.ts
```diff
@@ -14,6 +14,13 @@ const SUPPORTS_CONSTRUCTABLE_STYLESHEETS: boolean = (() => {
   }
 })();
 
+export interface CreateShadowDomRendererOptions {
+  /**
+   * If specified, a renderer will insert created CSSStyleSheets after this CSSStyleSheet.
+   */
+  insertionPoint?: ExtendedCSSStyleSheet;
```
ling1726, line 21: Why does this need to be an `ExtendedCSSStyleSheet` should a normal `CSSStylesheet ` be enough?

packages/shadow-dom/src/findInsertionPoint.ts
```diff
@@ -29,10 +30,14 @@ export function findInsertionPoint(
     const styleElement = styleSheets[i];
 
     if (comparer(styleElement) > 0) {
-      return styleSheets[i + 1] ?? null;
+      return styleSheets[i + 1] ?? insertionPoint ?? null;
```
ling1726, line 33: It would be good for this to be consistent with vanilla griffel - the `insertionPoint` is where insertion starts, so all stylesheets should be inserted after th `insertionPoint` but before the insertionPoint's next sibling.

It might make the `insertBefore` handling a bit more tedious but users can expect it to function the same way

packages/shadow-dom/src/findInsertionPoint.test.ts
```diff
@@ -65,6 +85,26 @@ describe('findInsertionPoint', () => {
...
+
+      expect(findInsertionPoint(renderer, styleSheet, insertionPoint)).toBe(insertionPoint);
+    });
+
+    it('handles insertionPoint with other (non-Griffel) stylesheets', () => {
+      const insertionPoint = new CSSStyleSheet() as ExtendedCSSStyleSheet;
+      const other1 = new CSSStyleSheet() as ExtendedCSSStyleSheet;
+      const other2 = new CSSStyleSheet() as ExtendedCSSStyleSheet;
+
+      const renderer = createRendererMock([other1, insertionPoint, other2]);
```
ling1726, line 102: This is actually incorrect - since the `renderer` adoptedStylesheets would actually never contain non-Griffel created stylesheets. The other stylesheets would only be present in the shadow root's adoptedStylesheets collection.

packages/shadow-dom/src/createShadowDOMRenderer.ts
spmonahan: It doesn't need to be `ExtendedCSSStyleSheet` but the shadow DOM renderer already [exposes this as the type of its `adoptedStyleSheets`](https://github.com/microsoft/griffel/blob/main/packages/shadow-dom/src/types.ts#L8) and [`findInsertionPoint()` uses this type](https://github.com/microsoft/griffel/blob/main/packages/shadow-dom/src/findInsertionPoint.ts#L11) so using it here feels more consistent to me.

Of course, currently we don't use any of the extended parts so we don't need to use this type but I think the API consistency is nice. That said, don't feel strongly about this.

packages/shadow-dom/src/findInsertionPoint.ts
```diff
@@ -39,3 +39,32 @@ export function findInsertionPoint(
 
   return null;
 }
+
+export function findShadowRootInsertionPoint(
```
spmonahan, line 43: I decided that since we're really maintaining two arrays and `insertionPoint` is irrelevant to the `renderer` it was better so separate the find logic.

Assuming this approach gets a 👍 I'll add tests for this.

packages/shadow-dom/src/createShadowDOMRenderer.ts
spmonahan: Done!

packages/shadow-dom/src/findInsertionPoint.test.ts
spmonahan: Removed

packages/shadow-dom/src/findInsertionPoint.ts
layershifter: @spmonahan I like the idea, I have pushed a small update to tests, so we don't need to repeat the logic ([3658c87](https://github.com/microsoft/griffel/pull/480/commits/3658c87d36df090c8987d848c65a257b65e87112)).

packages/shadow-dom/src/findInsertionPoint.ts
```diff
@@ -39,3 +39,32 @@ export function findInsertionPoint(
...
+  renderer: GriffelShadowDOMRenderer,
+  shadowRoot: ShadowRoot,
+  rendererTargetStyleSheet: ExtendedCSSStyleSheet | null,
+  insertionPoint?: CSSStyleSheet,
+): CSSStyleSheet | null {
+  const styleSheets = shadowRoot.adoptedStyleSheets;
+
+  if (renderer.adoptedStyleSheets.length < 1) {
+    return insertionPoint ? styleSheets[styleSheets.indexOf(insertionPoint) + 1] : null;
+  }
```
layershifter, line 53: I tried to understand when it's needed, but failed... Can you please a UT for this scenario? As so far it looks redundant.

packages/shadow-dom/src/findInsertionPoint.ts
spmonahan: Nice! Thanks for that!

packages/shadow-dom/src/findInsertionPoint.ts
spmonahan: You're right, it is redundant!

When I was working through this I had it as an "early out" conditional but this case is handled by the `while` loop just as well.

PR #500, opened 2024-01-29
feat: no invalid shorthand argument rule

NotWoods:
Adding a linter rule for invalid shorthand values like those warned agaisnt in [the docs](https://griffel.js.org/react/api/shorthands):
Caution: The most of the functions follow syntax in matching CSS properties, but each value should a separate argument

I also improved the `no-shorthands` rule to handle template strings and `!important`, then shared logic between the two rules.


layershifter (approved):

layershifter:
Love it ❤️ Will trigger a release after the merge 🚀

PR #511, opened 2024-03-04
fix(style-types): use custom fallback type

### LINKED ISSUE #510: `csstype` bump to 3.1.3 causing BREAKING CHANGEs on type-level
While bumping griffel build dependencies within fluent core, I run into unexpected type errors.

I was able to troubleshoot it deeper and I can confirm that the culprit of this issue is `csstype` package bump from `3.1.2` to `3.1.3` which happened [in `@griffel/core@1.15.2` ( and related packages to that release )](https://github.com/microsoft/griffel/blob/main/packages/core/CHANGELOG.md#1152).

Also other folks run into this issue: https://github.com/frenic/csstype/issues/189

I dont have context if that package minor bump is required, but if not I'd recommend to revert to `3.1.2` until the package authors will release a fixed version.

**Reproduction:**

**Repro:**

checkout: https://github.com/microsoft/fluentui/pull/30694

`yarn workspace @fluentui/react-positioning type-check`

```
src/createArrowStyles.ts:93:34 - error TS2769: No overload matches this call.
  Overload 1 of 6, '(width: BorderWidthInput, style: BorderStyleInput, color: BorderColorInput): BorderBottomStyle', gave the following error.
    Argument of type 'BorderBottomWidth<GriffelStylesCSSValue> | readonly BorderBottomWidth<GriffelStylesCSSValue>[]' is not assignable to parameter of type 'BorderWidthInput'.
      Type 'readonly BorderBottomWidth<GriffelStylesCSSValue>[]' is not assignable to type 'BorderWidthInput'.
        The type 'readonly BorderBottomWidth<GriffelStylesCSSValue>[]' is 'readonly' and cannot be assigned to the mutable type 'BorderWidth<GriffelStylesCSSValue>[]'.
  Overload 2 of 6, '(style: BorderStyleInput, width: BorderWidthInput, color: BorderColorInput): BorderBottomStyle', gave the following error.
    Argument of type 'BorderBottomWidth<GriffelStylesCSSValue> | readonly BorderBottomWidth<GriffelStylesCSSValue>[]' is not assignable to parameter of type 'BorderStyleInput'.
      Type '0' is not assignable to type 'BorderStyleInput'.

93       ...shorthands.borderBottom(borderWidth, borderStyle, borderColor),
                                    ~~~~~~~~~~~
```

layershifter:
Fixes #510.

### DIFF
--- a/packages/core/src/shorthands/types.test.ts
+++ b/packages/core/src/shorthands/types.test.ts
@@ -1,0 +1,25 @@
+import { type GriffelStyle } from '@griffel/style-types';
+import { borderBottom } from './borderBottom';
+
+describe('types tests', () => {
+  it('borderBottom', () => {
+    type Options = {
+      width?: GriffelStyle['borderBottomWidth'];
+      style?: GriffelStyle['borderBottomStyle'];
+      color?: GriffelStyle['borderBottomColor'];
+    };
+    const options: Options = {
+      width: '2px',
+      style: 'solid',
+      color: 'red',
+    };
+
+    const { width = '1px', style = 'solid', color = 'red' } = options;
+
+    expect(borderBottom(width, style, color)).toEqual({
+      borderBottomWidth: '2px',
+      borderBottomStyle: 'solid',
+      borderBottomColor: 'red',
+    });
+  });
+});
--- a/packages/style-types/src/shared.ts
+++ b/packages/style-types/src/shared.ts
@@ -1,3 +1,4 @@
 export type GriffelStylesCSSValue = string | 0;
 
 export type ValueOrArray<T> = T | Array<T>;
+export type Fallback<T> = { [P in keyof T]: T[P] | NonNullable<T[P]>[] };
--- a/packages/style-types/src/shared.ts
+++ b/packages/style-types/src/shared.ts
@@ -1,8 +1,9 @@
 export type GriffelStylesCSSValue = string | 0;
 
-export type ValueOrArray<T> = T | Array<T>;
 /**
  * instead of using `CSS.PropertiesFallback` from csstype that remaps properties to `readonly` union we need to use our cutom Fallback mapped type in order to avoid type issues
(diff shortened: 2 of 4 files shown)


packages/core/src/shorthands/types.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    };
+    const options: Options = {
+      width: '2px',
+      style: 'solid',
+      color: 'red',
+    };
+
+    const { width = '1px', style = 'solid', color = 'red' } = options;
+
+    expect(borderBottom(width, style, color)).toEqual({
```
Hotell, line 19: wondering if we should introduce test for all possible cases that can occur in order to mitigate future breaking changes on csstype level ?

Hotell (commented):
verified that this fix resolves issue with using private interface property types within fluent

packages/style-types/src/shared.ts
```diff
@@ -1,3 +1,4 @@
 export type GriffelStylesCSSValue = string | 0;
 
 export type ValueOrArray<T> = T | Array<T>;
+export type Fallback<T> = { [P in keyof T]: T[P] | NonNullable<T[P]>[] };
```
Hotell, line 4: can we document what's the purpose of this ?

```suggestion
/**
 * instead of using `CSS.PropertiesFallback` from csstype that remaps properties to `readonly` union we need to use our cutom Fallback mapped type in order to avoid type issues
 */
export type Fallback<T> = { [P in keyof T]: T[P] | NonNullable<T[P]>[] };
```

packages/core/src/shorthands/types.test.ts
layershifter: The test for other cases already exists, https://github.com/microsoft/griffel/blob/main/e2e/typescript/src/assets/fixture.ts.

ling1726 (approved):

PR #515, opened 2024-03-11
chore: remove usage of shorthands.* in E2E tests

layershifter:
Related to #531.

### DIFF
--- a/e2e/nextjs/src/assets/pages/index.tsx
+++ b/e2e/nextjs/src/assets/pages/index.tsx
@@ -1,4 +1,4 @@
-import { makeStyles, shorthands } from '@griffel/react';
+import { makeStyles } from '@griffel/react';
 import Head from 'next/head';
 import * as React from 'react';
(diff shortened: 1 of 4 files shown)


packages/webpack-extraction-plugin/src/webpackLoader.ts
```diff
@@ -97,31 +97,20 @@ function webpackLoader(
...
-          [
-            `/** @griffel:css-start [${cssBucketName}] **/`,
-            cssBucketRules.flatMap(entry => normalizeCSSBucketEntry(entry)).join(''),
-            `/** @griffel:css-end **/`,
-            '',
-          ].join('\n')
+          cssBucketRules
+            .map(entry => {
+              return [
+                `/** @griffel:css-start [${cssBucketName}] [${JSON.stringify(entry[1])}] **/`,
```
layershifter, line 107: This should be optimized

PR #516, opened 2024-03-11
Improve speed of serializer test function

rocketBANG:
The `test` function for the serializer plugin is called many times (from jest docs)
> Pay attention to efficiency in test because pretty-format calls it often.
https://github.com/jestjs/jest/tree/main/packages/pretty-format#test

Using the serializer for many elements causes a large slowdown due to the fact it constructs a large regex every time it's called
image (9)

I believe this can be improved because
1. Both `DEBUG_RESET_CLASSES` and `DEFINITION_LOOKUP_TABLE` are maps so we should be able to lookup a class without doing a string search
1. classes will always be space separated in the string, so we can split the string by spaces to isolate individual classes

This PR changes the `test` function to use a faster implementation
Simple benchmark test - https://jsperf.app/nevedi
image (6)

### DIFF
--- a/change/@griffel-jest-serializer-584b8140-267b-4569-be55-89a684191ee4.json
+++ b/change/@griffel-jest-serializer-584b8140-267b-4569-be55-89a684191ee4.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix: improve performance of serializer test function",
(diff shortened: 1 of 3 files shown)


layershifter (approved):
Love it, thx ❤️

Issue #502, opened 2024-02-08
`shorthands.flex(1)` should produce `flex: 1 1 0%` instead of `flex: 1 1 0px`


stefan-schweiger:
When you set a css property to `flex: 1` in any of the major browsers the result is`flex: 1 1 0%`. But the griffel shorthand `shorthands.flex(1)` produces `flex: 1 1 0px`.

What's maybe even weirder is that [the standard](https://www.w3.org/TR/css-flexbox-1/#propdef-flex) actually defines that `flex: <positive-number>` is the equivalent of `flex: <positive-number> 1 0;`. But somehow all browser vendors decided that this is just wrong and produce `flex: 1 1 0%` as mentioned above.

Either way following reality or following the standard griffel produces an incorrect output. Ideally it should follow how the major browser vendors do it to have the same developer experience as specifying `flex: 1`, because otherwise this can be a real pitfall.

stefan-schweiger:
https://github.com/w3c/csswg-drafts/issues/5742

layershifter:
@stefan-schweiger `shorthands.flex()` have been deprecated as CSS shorthands are supported natively, check #531 😉

PR #538, opened 2024-04-16
make `shorthands.flex` work the same as `flex` shorthand in css

jligeza:
Flex basis should be `0%` instead of `0px`, this is how web browsers work.

### DIFF
--- a/packages/core/src/shorthands/flex.test.ts
+++ b/packages/core/src/shorthands/flex.test.ts
@@ -29,7 +29,7 @@ describe('flex', () => {
     expect(flex(2)).toEqual({
       flexGrow: 2,
       flexShrink: 1,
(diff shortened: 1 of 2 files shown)


jligeza:

jligeza:
More info here:
https://github.com/microsoft/griffel/issues/502

ling1726:
@jligeza  there is active work to actually enable shorthands in Griffel (#544), most of the the shorthand factories (including flex) will be deprecated in the near future

layershifter:
@jligeza thanks for contributing, however `shorthands.flex()` have been deprecated, we are not going to fix it.

https://github.com/microsoft/griffel/issues/502#issuecomment-2092536984

PR #539, opened 2024-04-17
chore: add support for priority in CSS rules

layershifter:
### What's done?

This PR adds support for additional `priority` property in `meta` of CSS rules, this allows to order CSS rules based on it.

For example:

```ts
const rules = [
  '.bar { color: blue }',
  ['.foo { color: red }', { p: -1 }],
]
```

Will produce following CSS:

```css
.foo { color: red }
.bar { color: blue }
```

These changes are needed to support CSS shorthands.

### Minor notes

- `sortCSSRules()` have been refactored to have clearer implementation
- Snapshots in `sortCSSRules()` have been updated to be clearer

### DIFF
--- a/packages/core/src/renderer/getStyleSheetForBucket.ts
+++ b/packages/core/src/renderer/getStyleSheetForBucket.ts
@@ -44,7 +44,7 @@ const styleBucketOrderingMap = styleBucketOrdering.reduce((acc, cur, j) => {
   return acc;
 }, {} as Record<StyleBucketName, number>);
 
-export function getStyleSheetKey(bucketName: StyleBucketName, media: string, priority: number): string {
+export function getStyleSheetKey(bucketName: StyleBucketName, media: string, priority: number | string): string {
   return (bucketName === 'm' ? bucketName + media : bucketName) + priority;
 }
 
@@ -52,7 +52,7 @@ export function getStyleSheetKeyFromElement(styleEl: HTMLStyleElement): string {
   const bucketName = styleEl.getAttribute(DATA_BUCKET_ATTR) as StyleBucketName;
   const priority = styleEl.getAttribute(DATA_PRIORITY_ATTR) ?? '0';
 
-  return (bucketName === 'm' ? bucketName + styleEl.media : bucketName) + priority;
+  return getStyleSheetKey(bucketName, styleEl.media, priority);
 }
 
 /**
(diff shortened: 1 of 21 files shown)


packages/core/src/renderer/getStyleSheetForBucket.ts
```diff
@@ -38,12 +38,23 @@ export const styleBucketOrdering: StyleBucketName[] = [
...
 
+export function getStyleSheetKey(bucketName: StyleBucketName, media: string, priority: number): string {
+  return (bucketName === 'm' ? bucketName + media : bucketName) + priority;
+}
+
+export function getStyleSheetKeyFromElement(styleEl: HTMLStyleElement): string {
+  const bucketName = styleEl.getAttribute(DATA_BUCKET_ATTR) as StyleBucketName;
+  const priority = styleEl.getAttribute(DATA_PRIORITY_ATTR) ?? '0';
+
+  return (bucketName === 'm' ? bucketName + styleEl.media : bucketName) + priority;
```
ling1726, line 55:
```suggestion
  return getStyleSheetKey(bucketName, styleEl.media, Number(priority))
```

 nit

packages/core/src/renderer/getStyleSheetForBucket.ts
```diff
@@ -123,6 +153,12 @@ function findInsertionPoint(
   while (index >= 0) {
     const styleElement = styleElements.item(index);
 
+    if (isSameInsertionKey(styleElement, targetBucket, metadata)) {
```
ling1726, line 156: There's no need for branching logic here, the priority comparer is just another aspec of the comparing?

```ts
  const priorityComparer = (el: HTMLStyleElement) => priority - Number(el.getAttribute('data-priority'));
  let comparer: (el: HTMLStyleElement) => number = el => {
    if (isSameInsertionKey(el, targetBucket, metadata)) {
      return priorityComparer(el);
    }

    return targetOrder - styleBucketOrderingMap[el.getAttribute(DATA_BUCKET_ATTR) as StyleBucketName];
  }
```

ling1726 (approved):

packages/core/src/renderer/getStyleSheetForBucket.ts
layershifter: yeah, you're correct 👍

PR #546, opened 2024-04-23
chore: add script for updating shorthands mappings

layershifter:
- **feat: add support for style resets in makeStyles**
- **feat: add support for shorthands**
- **add script**


tools/update-shorthands/src/assignShorthandPriority.ts
```diff
@@ -0,0 +1,18 @@
+import { type CSSShorthands } from './types';
+
+export function assignShorthandPriority(shorthandProperties: CSSShorthands): CSSShorthands {
+  const propertiesWithPriority = { ...shorthandProperties };
+
+  Object.keys(shorthandProperties)
```
ling1726, line 6: maybe assign the default `-1` priorities in this function? The fact that the default priorities were assigned in the filter function confused me

ling1726 (approved):

tools/update-shorthands/src/assignShorthandPriority.ts
layershifter: Makes sense, will redo it.

PR #557, opened 2024-05-16
feat: add "classNameHashSalt" for AOT

layershifter:
This PR adds `classNameHashSalt` option for Babel plugin & Webpack loader.

The option will be properly documented once it will be implemented also for runtime.

### DIFF
--- a/packages/vite-plugin/project.json
+++ b/packages/vite-plugin/project.json
@@ -10,6 +10,7 @@
     },
     "test": {
       "executor": "nx:run-commands",
+      "dependsOn": [{ "target": "build", "dependencies": true }],
       "options": {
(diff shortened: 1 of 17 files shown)


packages/vite-plugin/project.json
```diff
@@ -10,6 +10,7 @@
     },
     "test": {
       "executor": "nx:run-commands",
+      "dependsOn": [{ "target": "build", "dependencies": true }],
```
ling1726, line 13: what's this change for?

ling1726 (approved):

packages/vite-plugin/project.json
layershifter: unrelated to this PR, reverted

PR #565, opened 2024-05-28
chore: add `eslint-plugin-react-compiler` to React package

spmonahan:
Adds the [React Compiler eslint plugin](https://react.dev/learn/react-compiler#installing-eslint-plugin-react-compiler) to enforce the [Rules of React](https://react.dev/reference/rules) via linting.

See related: https://github.com/microsoft/fluentui/pull/31457

### DIFF
--- a/packages/react/src/RendererContext.tsx
+++ b/packages/react/src/RendererContext.tsx
@@ -28,6 +28,8 @@ const RendererContext = React.createContext<GriffelRenderer>(createDOMRenderer()
  * @public
  */
 export const RendererProvider: React.FC<RendererProviderProps> = ({ children, renderer, targetDocument }) => {
+  'use no memo';
+
   if (canUseDOM()) {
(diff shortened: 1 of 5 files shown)


packages/react/src/RendererContext.tsx
```diff
@@ -28,6 +28,8 @@ const RendererContext = React.createContext<GriffelRenderer>(createDOMRenderer()
  * @public
  */
 export const RendererProvider: React.FC<RendererProviderProps> = ({ children, renderer, targetDocument }) => {
+  'use no memo';
```
spmonahan, line 31: We should probably address the underlying issue but I've opted to simply disable for now to keep this PR focused on landing the lint rule, not fixing anything it surfaces.

Have a first draft of a potential fix here, though I think it needs work: #566 566

layershifter (approved):
Oops, somehow missed this. Anyway, bumped plugin and rebased. Thx 👍

PR #566, opened 2024-05-28
fix: remove eslint exception in `RendererProvider`

spmonahan:
Changes the code to strictly follow the [Rules of Hooks](https://react.dev/reference/rules/rules-of-hooks).

Previously the code conditionally called a hook which prevents React Compiler from making any potential optimizations. Now the conditional has been moved inside the hook.

However, this code is still potentially problematic as we are relying on [`useMemo()` for correctness rather than as a performance optimization](https://react.dev/reference/react/useMemo):

msedge_dtpiVAZAuS

Depends on #565

### DIFF
--- a/packages/react/src/RendererContext.tsx
+++ b/packages/react/src/RendererContext.tsx
@@ -28,18 +28,13 @@ const RendererContext = React.createContext<GriffelRenderer>(createDOMRenderer()
  * @public
  */
 export const RendererProvider: React.FC<RendererProviderProps> = ({ children, renderer, targetDocument }) => {
-  'use no memo';
-
-  if (canUseDOM()) {
-    // This if statement technically breaks the rules of hooks, but is safe because the condition never changes after
-    // mounting.
-    // eslint-disable-next-line react-hooks/rules-of-hooks
-    React.useMemo(() => {
+  React.useMemo(() => {
+    if (canUseDOM()) {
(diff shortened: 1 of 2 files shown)


packages/react/src/RendererContext.tsx
```diff
@@ -28,18 +28,13 @@ const RendererContext = React.createContext<GriffelRenderer>(createDOMRenderer()
...
  */
 export const RendererProvider: React.FC<RendererProviderProps> = ({ children, renderer, targetDocument }) => {
-  'use no memo';
-
-  if (canUseDOM()) {
-    // This if statement technically breaks the rules of hooks, but is safe because the condition never changes after
-    // mounting.
-    // eslint-disable-next-line react-hooks/rules-of-hooks
-    React.useMemo(() => {
+  React.useMemo(() => {
```
layershifter, line 31: Actually, it could be a nice fit for `useInsertionEffect()`. @spmonahan if you interested, feel free to take a look 😉

layershifter (approved):

packages/react/src/RendererContext.tsx
spmonahan: Yeah, looks like a possibility :)

My focus on React Compiler changes so far has been to just grant any existing code an exception so we can lint without errors and then come back and revisit all the exceptions later. Basically, don't break anything that is currently working ;)

PR #586, opened 2024-07-25
feat(core): add "classNameHashSalt" option

### LINKED ISSUE #453: feature: add support for adding prefix to CSS to scope down the styling
We have applications overlap in one page.

```
<html>
  <App-a>
  <!-- inserting griffel.css -->
  </App-a>
  <App-b>
  <!-- inserting <style data-make-styles-bucket="d"></style> -->
  </App-b>
</html>
```

Since the hashed className .fqdk4by may both existing in griffel.css and data-make-styles-bucket. They may cause global pollution or something else. So we need to add prefix for className to make CSS scoped.

Maybe solution like selectorPrefix https://fela.js.org/docs/latest/advanced/renderer-configuration

layershifter:
Fixes #453.

### DIFF
--- a/apps/website/docs/react/api/create-dom-renderer.md
+++ b/apps/website/docs/react/api/create-dom-renderer.md
@@ -28,7 +28,7 @@ A salt that will be added for generated hashed classes. Check [microsoft/griffel
 
 :::caution Use with caution
 
-Should be the same for all renderers in the same application (bundle).
+There cannot be more than **one single** hash salt in the same application (bundle) 
 
 :::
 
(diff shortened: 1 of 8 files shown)


apps/website/docs/react/api/create-dom-renderer.md
```diff
@@ -22,6 +22,16 @@ function App(props) {
 }
 ```
 
+### classNameHashSalt
+
+A salt that will be added for generated hashed classes. Check [microsoft/griffel#453](https://github.com/microsoft/griffel/issues/453) for an example use case.
+
+:::caution Use with caution
+
+Should be the same for all renderers in the same application (bundle).
```
ling1726, line 31:
```suggestion
There cannot be more than **one single** hash salt in the same application (bundle) 
```

ling1726 (approved):

PR #587, opened 2024-07-29
feat: Support stylelint rule disabling for style slots

ling1726:
Adds comment directive parsing in the location preset so that that comments starting with `griffel-` are collected with styles metadata.

```ts
const useStyles = makeStyles({
 // griffel-csslint-disable color-no-hex   
 slot: {
    color: '#fff'
  },
});

// griffel-csslint-disable color-no-hex
const useResetStyles = makeResetStyles({
  color: '#fff'
})
```

Updates the parser to transform `griffel-csslint-disable` to `styleint-disable` comments in output CSS for attached style slots. The above example outputs to

```css
.fxxx { color: red } /* stylelint-disable-line color-no-hex */
.rxxx {  color: red } /* stylelint-disable-line color-no-hex */
```

### DIFF
--- a/packages/postcss-syntax/src/location-preset.ts
+++ b/packages/postcss-syntax/src/location-preset.ts
@@ -2,13 +2,30 @@ import type { PluginObj, PluginPass, types as t, ConfigAPI } from '@babel/core';
 import { declare } from '@babel/helper-plugin-utils';
 import type { BabelPluginOptions } from '@griffel/babel-preset';
 
+export type CommentDirective = [/** directive */ string, /** value */ string];
+export type CommentDirectivesBySlot = Record</** slot */ string, CommentDirective[]>;
+export type CommentDirectivesByHookDeclarator = Record</** hook declarator */ string, CommentDirectivesBySlot>;
+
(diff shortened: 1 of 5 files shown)


packages/postcss-syntax/src/location-preset.ts
```diff
@@ -113,12 +141,46 @@ const plugin = declare<LocationPluginOptions, PluginObj<LocationPluginState>>((a
...
+  if (!leadingComments) {
+    return null;
+  }
+
+  const entries = leadingComments
+    // We don't support comment blocks
+    .filter(comment => comment.type === 'CommentLine')
+    .map(comment => {
+      const commentValue = comment.value.trim();
+      if (!commentValue.startsWith('griffel-')) {
```
ling1726, line 172: The hardcoded strings are never used more than once per module in this PR. If we do ever support more directives, consider moving to the main babel preset with a more elaborate code around adding/parsing directives

layershifter (approved):

YuanboXue-Amber (approved):

PR #595, opened 2024-08-11
feat(eslint-plugin): add top-level-styles eslint rule

benkeen:
Hey folks! This adds a new `top-level-styles` eslint rule to forbid using the `makeStyles`, `makeResetStyles` and `makeStaticStyles` methods anywhere other than in the top-level of a file. 

It's kind of opinionated, so I understand if you don't want to accept the PR, but it seems like a reasonable linting rule addition to the "recommended" lint settings, even if it's more restrictive for developers. 

## Rationale

We're using Fluent 9 + Griffel in our app + recently ran into a variety of build errors once we imported a new set of packages.  Several package authors were misusing the Griffel methods (`makeStyles` in particular) and passing in runtime variables. In every case, they'd create a helper method that was doing the job of calling makeStyles.  

By forcing developers to always call these methods in the root scope/top level of a file, it makes it less likely people will feed in runtime vars. Seems like a sensible pattern to enforce and it's a linting rule we want to pass on to package authors to use so we don't have to keep bugging them to fix their Griffel usages. :) 

---------------

Edit: _acccttuuuallly_, if you do decide to accept the PR + make it part of the recommended linting defaults, let me know and I'll add in a configuration option to let users define their own package name(s), just in case they're re-exporting the griffel methods. Should probably get that in for the first version.

### DIFF
--- a/change/@griffel-eslint-plugin-dd20b1e1-c272-409a-95f9-19ca824187ce.json
+++ b/change/@griffel-eslint-plugin-dd20b1e1-c272-409a-95f9-19ca824187ce.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add top-level-styles eslint rule",
+  "packageName": "@griffel/eslint-plugin",
+  "email": "ben.keen@gmail.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@griffel-eslint-plugin-dd20b1e1-c272-409a-95f9-19ca824187ce.json
+++ b/change/@griffel-eslint-plugin-dd20b1e1-c272-409a-95f9-19ca824187ce.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Add top-level-styles eslint rule",
+  "comment": "feat: add top-level-styles eslint rule",
   "packageName": "@griffel/eslint-plugin",
   "email": "ben.keen@gmail.com",
   "dependentChangeType": "patch"
--- a/packages/eslint-plugin/README.md
+++ b/packages/eslint-plugin/README.md
@@ -23,7 +23,7 @@ Add `@griffel` to the plugins section and `plugin:@griffel/recommended` to the e
 
 ## Shareable configurations
 
-This plugin exports recommended configuration that enforce good practices, but you can use only uses that are required for your use case:
+This plugin exports recommended configuration that enforce good practices, but you can choose to use only the ones that are necessary for your use case:
 
 ```json
 {
@@ -48,4 +48,6 @@ You can find more info about enabled rules in the [Supported Rules section](#sup
 | [`@griffel/no-invalid-shorthand-arguments`](./src/rules/no-invalid-shorthand-arguments.md) | All shorthands must not use space separators, and instead pass in multiple arguments                            | ✅  |
 | [`@griffel/no-shorthands`](./src/rules/no-shorthands.md)                                   | Enforce usage of CSS longhands                                                                                  | ✅  |
 | [`@griffel/styles-file`](./src/rules/styles-file.md)                                       | Ensures that all `makeStyles()` and `makeResetStyles()` calls are placed in a `.styles.js` or `.styles.ts` file | ❌  |
+limitations) | ❌ |
 | [`@griffel/pseudo-element-naming`](./src/rules/pseudo-element-naming.md)                   | Ensures that all Pseudo Elements start with two colons                                                          | ✅  |
+| [`@griffel/top-level-styles`](./src/rules/top-level-styles.md)                    | Ensures that all `makeStyles()`, `makeResetStyles()` and `makeStaticStyles()` calls are written in the top level of a file to discourage developer misuse.
--- a/packages/eslint-plugin/README.md
+++ b/packages/eslint-plugin/README.md
@@ -47,7 +47,6 @@ You can find more info about enabled rules in the [Supported Rules section](#sup
 | [`@griffel/hook-naming`](./src/rules/hook-naming.md)                                       | Ensure that hooks returned by the `makeStyles()` function start with "use"                                      | ❌  |
 | [`@griffel/no-invalid-shorthand-arguments`](./src/rules/no-invalid-shorthand-arguments.md) | All shorthands must not use space separators, and instead pass in multiple arguments                            | ✅  |
 | [`@griffel/no-shorthands`](./src/rules/no-shorthands.md)                                   | Enforce usage of CSS longhands                                                                                  | ✅  |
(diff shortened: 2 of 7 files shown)


packages/eslint-plugin/README.md
```diff
@@ -48,4 +48,6 @@ You can find more info about enabled rules in the [Supported Rules section](#sup
 | [`@griffel/no-invalid-shorthand-arguments`](./src/rules/no-invalid-shorthand-arguments.md) | All shorthands must not use space separators, and instead pass in multiple arguments                            | ✅  |
 | [`@griffel/no-shorthands`](./src/rules/no-shorthands.md)                                   | Enforce usage of CSS longhands                                                                                  | ✅  |
 | [`@griffel/styles-file`](./src/rules/styles-file.md)                                       | Ensures that all `makeStyles()` and `makeResetStyles()` calls are placed in a `.styles.js` or `.styles.ts` file | ❌  |
+limitations) | ❌ |
```
spmonahan, line 51: Looks like a formatting issue

msedge_cKmJPXr4G7

packages/eslint-plugin/src/rules/top-level-styles.md
```diff
@@ -0,0 +1,64 @@
+# Ensure `makeStyles`, `makeResetStyles` and `makeStaticStyles` are only called at the top-level of a file
+
+This is an opinionated rule to discourage developers from misusing the `makeStyles`, `makeResetStyles` and `makeStaticStyles`
```
spmonahan, line 3: Is this really an opinionated rule? In dev mode misusing `makeStyles` [throws an error](https://github.com/microsoft/griffel/blob/main/packages/react/src/makeStyles.ts#L14).

benkeen: I think that check just confirms it's not within a component, correct? We want to forbid it within any method at all. 

I regard it as opinionated because _technically_ these methods could be used anywhere  - as long as the dev doesn't use pass dynamic vars. But forcing it only to be used in the top-level makes it a lot less likely they do that.

spmonahan: I wonder if the rule should be "no runtime parameters" in that case? Not sure how feasible that is though 😀

benkeen: I know, hey! Would've been nice to address the root cause, but that didn't seem plausible... `no-runtime-parameters` might make developers a little overconfident :)

change/@griffel-eslint-plugin-dd20b1e1-c272-409a-95f9-19ca824187ce.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Add top-level-styles eslint rule",
```
layershifter, line 3:
```suggestion
  "comment": "feat: add top-level-styles eslint rule",
```

packages/eslint-plugin/README.md
```diff
@@ -48,4 +48,6 @@ You can find more info about enabled rules in the [Supported Rules section](#sup
...
 | [`@griffel/pseudo-element-naming`](./src/rules/pseudo-element-naming.md)                   | Ensures that all Pseudo Elements start with two colons                                                          | ✅  |
+| [`@griffel/top-level-styles`](./src/rules/top-level-styles.md)                    | Ensures that all `makeStyles()`, `makeResetStyles()` and `makeStaticStyles()` calls are written in the top level of a file to discourage developer misuse.
```
layershifter, line 53:
```suggestion
| [`@griffel/top-level-styles`](./src/rules/top-level-styles.md)                    | Ensures that all `makeStyles()`, `makeResetStyles()` and `makeStaticStyles()` calls are written in the top level of a file to discourage developer misuse. | ❌  |
```

A missing cell?

packages/eslint-plugin/src/rules/top-level-styles.md
layershifter, line 3:
```suggestion
This is a rule to discourage developers from misusing the `makeStyles`, `makeResetStyles` and `makeStaticStyles` 
```

Agree with @spmonahan, there are no opinions, `makeStyles()` & co. calls should be top-level 🐱

packages/eslint-plugin/src/rules/top-level-styles.md
```diff
@@ -0,0 +1,64 @@
...
+
+export const getStyles = () => {
+    const useStyles = makeStyles({
+        root: {
+            backgroundColor: 'red',
+        },
+    });
+
+    return useStyles;
+}
```
layershifter, line 22:
```suggestion
export const getStyles = () => {
  const useStyles = makeStyles({
    root: {
      backgroundColor: 'red',
    },
  });

  return useStyles;
};
```

nit: formating

packages/eslint-plugin/src/rules/top-level-styles.md
```diff
@@ -0,0 +1,64 @@
...
+  getStyles () {
+    const styles = makeStyles({
+        root: {
+            backgroundColor: 'red',
+        },
+    });
+
+    return styles;
+  }
+}
```
layershifter, line 39:
```suggestion
export class MyClass {
  getStyles() {
    const styles = makeStyles({
      root: {
        backgroundColor: 'red',
      },
    });

    return styles;
  }
}
```

nit

packages/eslint-plugin/src/rules/top-level-styles.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+
```
layershifter, line 42:
```suggestion
```

packages/eslint-plugin/src/rules/top-level-styles.ts
```diff
@@ -0,0 +1,53 @@
...
+
+      'FunctionDeclaration CallExpression, ArrowFunctionExpression CallExpression, FunctionExpression CallExpression':
+        function (node) {
+          if (!isMakeStylesImported) {
+            return;
+          }
+          const methodName = getMakeStylesCallExpression(node, 'makeStyles', 'makeStaticStyles', 'makeResetStyles');
+          if (methodName) {
+            console.log('callExpression');
+
```
layershifter, line 41:
```suggestion
```

oops

benkeen:
Great feedback, thanks @spmonahan & @layershifter! I'll get it updated soon.

packages/eslint-plugin/src/rules/top-level-styles.ts
benkeen: Oops!

packages/eslint-plugin/src/rules/top-level-styles.md
benkeen: Text updated.

benkeen:
Updated, thanks @layershifter.

layershifter (approved):
Thanks @benkeen ❤️

PR #618, opened 2024-11-11
Deprecation notices for all deprecated shorthands

malladi-s:
Deprecated tag is needed for all overloads of deprecated shorthand functions to communicate that a shorthand is deprecated

Because of https://github.com/microsoft/TypeScript/issues/407, deprecation notices are not coming through today when there are multiple overloads of a deprecated shorthand function
image

When there are no overloaded functions, deprecation notices come through 
image

### DIFF
--- a/change/@griffel-core-e0dc7909-f321-4c3e-b9f8-0a2a280e37cf.json
+++ b/change/@griffel-core-e0dc7909-f321-4c3e-b9f8-0a2a280e37cf.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "All deprecated shorthands functions are tagged properly",
+  "packageName": "@griffel/core",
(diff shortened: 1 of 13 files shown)


malladi-s:
@microsoft-github-policy-service agree company="Microsoft"

change/@griffel-core-e0dc7909-f321-4c3e-b9f8-0a2a280e37cf.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "All deprecated shorthands functions are tagged properly",
```
layershifter, line 3:
```suggestion
  "comment": "chore: all deprecated shorthands functions are tagged properly",
```

PR #621, opened 2024-11-14
RFC: Improvements to a compiler to make it fast

layershifter:


Anber (approved):

rfcs/fast-n-furious.md
```diff
@@ -0,0 +1,354 @@
...
+
+This POC also remained as a POC.
+
+## Fast & Furious 🏎️
+
+Our objective is to significantly enhance the compiler's speed, aiming for at least a 10x improvement.
+
+### Changes in tooling
+
+We will adopt Rust as the primary language for the next-gen compiler and use [`napi-rs`](https://github.com/napi-rs/napi-rs) to create Node.js bindings.
```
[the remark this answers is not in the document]
layershifter: That's what @Anber proposed yesterday 🐱 It's what we will need to check once we will have something working. 

The problem with WASM that we faced before with @ling1726 - the cost of moving strings between JS & WASM worlds. Longer strings => worse results, as the following happens:

```js
for (let i = index; i < index + size; ++i)
  s += String.fromCharCode(buffer[i]);
```

> https://stackoverflow.com/a/41372484/6488546

 AFAIK there is nothing better yet than messing up with `Uint8Array`s that was proven to be extremely slow when we re-implemented `@emotion/hash` with WASM and managed to get faster than JS only with a custom memory allocator.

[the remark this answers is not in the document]
ling1726: In order for strings to exist in the wasm shared memory (which is essentially numbers) serialization with shared buffers still need to happen. Once wasm-bindgen compiles the output will throw something like this

```ts
function passStringToWasm0(arg, malloc, realloc) {

    if (typeof(arg) !== 'string') throw new Error(`expected a string argument, found ${typeof(arg)}`);

    if (realloc === undefined) {
        const buf = cachedTextEncoder.encode(arg);
        const ptr = malloc(buf.length, 1) >>> 0;
        getUint8ArrayMemory0().subarray(ptr, ptr + buf.length).set(buf);
        WASM_VECTOR_LEN = buf.length;
        return ptr;
    }

    let len = arg.length;
    let ptr = malloc(len, 1) >>> 0;

    const mem = getUint8ArrayMemory0();

    let offset = 0;

    for (; offset < len; offset++) {
        const code = arg.charCodeAt(offset);
        if (code > 0x7F) break;
        mem[ptr + offset] = code;
    }
   // etc....
```

If wasm-bindgen has fs access - it might be possible to only do serialization into a buffer once, and write the output from the wasm side... but that text encoding will still need to happen to make sure the string exists in wasm in the first place 😢

rfcs/fast-n-furious.md
```diff
@@ -0,0 +1,354 @@
...
+  root: { paddingLeft: spacing(1) },
+});
+```
+
+In this scenario, the context of `input.ts` is insufficient to pre-create CSS rules. The compiler must resolve the `spacing()` function and substitute its call with the actual value.
+
+This introduces complexity. Internally, the compiler functions as a micro-bundler that resolves imports, evaluates expressions, and replaces them with actual values. The simplified lifecycle is below.
+
+```mermaid
+graph TD
```
ling1726, line 62: code snippets in diagrams seem to be broken for me 
image

rfcs/fast-n-furious.md
```diff
@@ -0,0 +1,354 @@
...
+
+#### Module evaluation
+
+We will no longer evaluate modules on the go. Instead, we will evaluate all modules at once, similar to a bundle. We will use an approach inspired by Webpack, where after processing steps, we will evaluate:
+
+```js
+// Output code
+
+const fn = () => {
+  function require(moduleId) {
```
ling1726, line 334:
```suggestion
  function __require_wd40_module(moduleId) {
```

Have I understood this `require` correctly?

rfcs/fast-n-furious.md
```diff
@@ -0,0 +1,354 @@
...
+    /* ---- */
+  }
+
+  function __wd40_module(filename, fn) {
+    /* ---- */
+  }
+
+  __wd40_module('/theme.ts', function (module, exports, require) {
+    exports.spacing = value => value + 'px';
```
ling1726, line 343: this example uses a simple function transform for spacing. If this function became more complicated to use imported constants do I understand correctly that it could become:


```js
__wd40_module('theme.ts', function (module, exports, require) {
  exports.spacing = value => {
    value + require('/contants.ts').PX;
  };
});
```

rfcs/fast-n-furious.md
```diff
@@ -0,0 +1,354 @@
...
+    D --> C2[collect imports, tree-shake, transform `theme.ts`]
+    C2 --> C3[resolve imports in `theme.ts`]
+    C3 --> E
+```
+
+#### Strict mode (performance mode)
+
+We will introduce a strict mode that disables certain features to enhance compiler speed. For instance, this mode will not support `export *` and CommonJS modules.
+
+**This mode will be implemented first.**
```
ling1726, line 324: Might want to make it clear that non-performant syntaxes like `export *` will be supported eventually?

PR #629, opened 2024-12-18
fix(postcss-syntax): parse styles with newlines

YuanboXue-Amber:
### Before:
When styles contains newlines, postcss-syntax cannot parse it correctly:
The below js
```js
import { makeStyles } from "@griffel/react";

  slot1: {
    background: \`linear-gradient(#e66465, 
      #9198e5)\`,
  },
  slot2: {
    color: "blue",
  },
});
```
generates css:
```css
.f1qmhkic{background:linear-gradient(#e66465, 
      #9198e5);}
.f163i14w{color:blue;}
```

postcss-syntax has an assumption that each line belongs to one slot. So it thinks slot1's style is `.f1qmhkic{background:linear-gradient(#e66465,` and slot2's style is `      #9198e5);}`.

### After:
We replace newline in the css with space. The generated css is:
```css
.f1qmhkic{background:linear-gradient(#e66465,        #9198e5);}
.f163i14w{color:blue;}
```

### DIFF
--- a/packages/postcss-syntax/src/parse.ts
+++ b/packages/postcss-syntax/src/parse.ts
@@ -39,7 +39,7 @@ export const parse = (css: string | { toString(): string }, opts?: ParserOptions
   Object.entries(cssEntries).forEach(([declarator, slots]) => {
     Object.entries(slots).forEach(([slot, rules]) => {
       cssRuleSlotNames.push(`${declarator} ${slot}`);
-      let cssRule = rules.join('');
+      let cssRule = rules.join('').replace(/\n/g, ' ');
 
       const ignoredRules = getIgnoredRulesFromDirectives(commentDirectives[declarator]?.[slot] ?? []);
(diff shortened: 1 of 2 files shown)


packages/postcss-syntax/src/parse.ts
```diff
@@ -39,7 +39,7 @@ export const parse = (css: string | { toString(): string }, opts?: ParserOptions
   Object.entries(cssEntries).forEach(([declarator, slots]) => {
     Object.entries(slots).forEach(([slot, rules]) => {
       cssRuleSlotNames.push(`${declarator} ${slot}`);
-      let cssRule = rules.join('');
+      let cssRule = rules.join('').replace(/\n/g, ' ');
```
ling1726, line 42: Can you pleause use [os.EOL](https://nodejs.org/api/os.html#os_os_eol) to handle windows/nix formats?

PR #641, opened 2025-02-04
feat: postcss-syntax should silence parse errors on demand

ling1726:
Since the postcss-syntax is primarily used for linting with Stylelint, there are two areas where it can be used

1. CI pipeline check
2. IDE integration

For 1. we want to always throw but for 2. where users can be in the process of editing their code, it's not always parseable so we want to be able to silence here.

### DIFF
--- a/packages/postcss-syntax/src/parse.ts
+++ b/packages/postcss-syntax/src/parse.ts
@@ -13,23 +13,30 @@ import * as os from 'os';
 
 export type PostCSSParserOptions = Pick<postcss.ProcessOptions<postcss.Document | postcss.Root>, 'from' | 'map'>;
 
-export interface ParserOptions extends Pick<PostCSSParserOptions, 'from'>, BabelPluginOptions {}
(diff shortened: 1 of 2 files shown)


packages/postcss-syntax/src/parse.ts
```diff
@@ -91,6 +95,33 @@ export const parse = (css: string | { toString(): string }, opts?: ParserOptions
...
+        generateMetadata: true,
+      },
+    });
+
+    return metadata;
+  } catch (error) {
+    console.warn('Failed to parse Griffel styles');
+    console.warn(error);
+    if (silenceParseErrors) {
+      return null;
```
YuanboXue-Amber, line 118: What about console.warn only when silenceParseErrors?

YuanboXue-Amber (approved):

packages/postcss-syntax/src/parse.ts
ling1726: so we still have some way of tracing the parse error

PR #652, opened 2025-03-17
chore: update GH actions & pin versions

layershifter:
- Adds `github.repository_owner` to pre-checks
- Splits `bundle-size`
- Pins versions of third-party actions
- Replaces `dawidd6/action-download-artifact` with `actions/download-artifact`


.github/workflows/bundle-size-comment.yml
```diff
@@ -8,26 +8,25 @@ on:
...
 
       - name: Load PR number
         id: pr_number
         run: echo "::set-output name=id::$(cat pr.txt)"
         working-directory: ./results
 
-      - name: 'Comment on PR'
-        uses: marocchino/sticky-pull-request-comment@v2
+      - name: Comment on PR
+        uses: marocchino/sticky-pull-request-comment@v2.9.1
```
Hotell, line 29: going forward we wanna allow only SHAs + dependabot automation - see https://github.com/microsoft/fluentui/pull/34004/files

.github/workflows/bundle-size-main.yml
Hotell, line 1: nit: `bundle-size-baseline.yml`

.github/workflows/ci.yml
```diff
@@ -24,7 +28,7 @@ jobs:
       - run: yarn install --immutable
       - run: yarn check-dependencies
 
-      - run: yarn nx affected --target=lint --parallel --max-parallel=3
-      - run: yarn nx affected --target=type-check --parallel --max-parallel=3
-      - run: yarn nx affected --target=build --parallel --max-parallel=3
-      - run: yarn nx affected --target=test --parallel --max-parallel=2
+      - run: yarn nx affected --target=lint
```
Hotell, line 31: question:

1. should we enable bail mode ? with this state if 1 task failed everything else will still run taking addional resources. I'd suggested to use `--nxBail`

2. wondering if we can enable all in one mode to utilize better resources  `yarn nx affected -t lint,type-check,build,test` ?

.github/workflows/bundle-size-comment.yml
layershifter: Cool, let's see if it works okay and then adopt in other repos 👍

.github/workflows/ci.yml
layershifter: 1 & 2, sure, nice idea 👍

.github/workflows/bundle-size-comment.yml
dannyvv: Also don't forget to update the 'allow-list' in the action settings.
We should only allow github ones, and a limmited list of actions with Sha there.

PR #661, opened 2025-04-07
Use capital H in GitHub

eliperkins:
GitHub is spelled with a capital H. This fixes user-facing usage of `Github` by changing it to `GitHub`.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -4,7 +4,7 @@ pr: none
 trigger: none
 
 variables:
-  - group: 'GitHub and NPM secrets'
+  - group: 'Github and NPM secrets'
   - name: tags
     value: production,externalfacing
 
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -4,7 +4,7 @@ pr: none
 trigger: none
 
 variables:
-  - group: 'Github and NPM secrets'
+  - group: 'GitHub and NPM secrets'
```
ling1726, line 7: Please revert this change - this could break the release pipeline

ling1726 (dismissed):
Please revert changes in the azure pipelines config

azure-pipelines.yml
eliperkins: What do you think about using a suggestion in the future?

```suggestion
  - group: 'Github and NPM secrets'
```

layershifter (approved):

PR #693, opened 2025-09-02
chore: add permissions to all jobs

layershifter:
- Sets explicit permissions
- Remove deprecated patterns
- Follow best practices

### DIFF
--- a/.github/workflows/bundle-size-baseline.yml
+++ b/.github/workflows/bundle-size-baseline.yml
@@ -1,4 +1,4 @@
-name: Bundle size [baseline]
+name: Bundle size | Baseline
 on:
   push:
     branches:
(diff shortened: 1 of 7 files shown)


.github/workflows/bundle-size-baseline.yml
```diff
@@ -11,16 +11,14 @@ jobs:
...
     permissions:
       contents: 'read'
       actions: 'read'
 
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - name: Derive appropriate SHAs for base and head for `nx affected` commands
-        uses: nrwl/nx-set-shas@dbe0650947e5f2c81f59190a38512cf49126fe6b # v4.3.0
-
```
layershifter, line 23: Not needed by this job as we use `run-many` anyway

PR #836, opened 2026-04-13
fix(webpack-plugin): fix webpack plugin loader dirname path construction

yang991178:
The loader in the webpack plugin seems to have an issue on Windows when the path contains spaces. Specifically, `new URL(import.meta.url).pathname` returns URL-encoded strings (e.g., `%20` for spaces), causing webpack to fail when the project path contains spaces. This PR updates the line to use `fileURLToPath`, which should correctly handle path decoding on all platforms.

In `packages/webpack-plugin/src/webpackLoader.mts`:
* Added import for `fileURLToPath` from `'node:url'`
* Changed `const __dirname = path.dirname(new URL(import.meta.url).pathname);` → `const __dirname = path.dirname(fileURLToPath(import.meta.url));`

### DIFF
--- a/change/@griffel-webpack-plugin-fix-windows-path.json
+++ b/change/@griffel-webpack-plugin-fix-windows-path.json
@@ -2,6 +2,6 @@
   "type": "patch",
   "comment": "fix: use fileURLToPath for Windows path compatibility",
   "packageName": "@griffel/webpack-plugin",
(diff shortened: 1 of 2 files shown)


yang991178:

layershifter (approved):
Thx!

PR #848, opened 2026-04-17
feat(core): add CSS @scope at-rule support

joeflateau:
Add support for the CSS @scope at-rule in makeStyles, enabling scoped style boundaries that prevent styles from leaking into nested components.

Key changes:
- Add 'scope' and 'scopeToken' to AtRules type
- Add isScopeSelector utility (follows isContainerQuerySelector pattern)
- Handle @scope in resolveStyleRules with stable scope token generation
- Handle @scope in compileAtomicCSSRule with root/boundary & replacement
- Include scope in hash key computation and style bucket assignment ('t')
- Support `@scope (&) to (&)` syntax where root gets the atomic class and boundary gets a stable scope token class for nested exclusion

The scope token is a stable class name shared across all atomic rules in a @scope block, allowing elements with that class to act as scope boundaries. Inside @scope blocks, the atomic class selector is replaced with :scope to correctly reference the scope root element.


layershifter:
@joeflateau thx for rising this PR! I was sitting with Claude and I defined the strategy & plan around it. (I don't paste AI slop below, so be patient with my human written text :P)

There are two parts, reducing the scope (of the feature) & implementing it across the library in a fundamentally correct way.

## Reducing scope

The spec (https://drafts.csswg.org/css-cascade-6/#scoped-styles) allows the following:

```css
@scope {}
@scope to (SELECTOR) {}

@scope (SELECTOR) {}
@scope (SELECTOR_A) to (SELECTOR_B) {}
```

- First two are out of scope by design.
- Griffel already bailouts from fully CSS syntax in some forms, so it's okay to do support a need subset.
- `@scope (SELECTOR) {}` does not have sense as it's effectively any `makeStyles()` call by design :cat
- `@scope (SELECTOR_A) to (SELECTOR_B) {}` - we don't want to allow `SELECTOR_A` to be customized as then the following call:
  ```ts
  makeStyles({
    root: { '@scope (.foo) to (.bar)`: { '& img': { color: 'red' }}
  })
  ```
  Will produce:
  ```css
  @scope (.foo) to (.bar) {
    img { color: red }
  }
  ```
  This has no sense as there is no atomic part

### Proposal

- Don't support `:scope`. If we will see a reason, we will add it later
- Expose only `@scope to (SELECTOR) {}` as sugar
  - Internally it will be expanded to `@scope (ATOMIC_CLASS) to (SELECTOR)`
  - Emit errors/warnings on unsupported syntax
- `makeStyles()` & `makeResetStyles()` should support it

## Handling style buckets (i.e. ordering)

Current implementation is putting all `@scope` rules into `t` bucket. That will be a problem and will break any existing order, check [lvha-bug.html](https://github.com/user-attachments/files/26965236/lvha-bug.html)

We will need another solution. I have something in mind, but will share once we will agree on the current part.

joeflateau:
Obsolete — superseded by the merged stack #852, #853, #854, #855, #856, #912, #857, #858, #859, #861. Closing.

PR #853, opened 2026-04-28
test(core): add Vitest browser-mode infra and LVFHA cascade tests

joeflateau:
> [!NOTE]
> Stacked on top of #852. The diff currently includes `nano-staged@~1.0.2` because that PR hasn't merged yet; once it does, this branch will be rebased and the diff will shrink to just the additions below.

## Summary

Introduces real-browser testing for `@griffel/core` alongside the existing jsdom unit tests, plus a first suite covering the [LVFHA pseudo-class cascade](https://griffel.js.org/react/guides/atomic-css/#lvfha-order-of-pseudo-classes). No `@scope` content here — that lands in PR 1+.

## What changes

- `packages/core/vitest.config.ts` — `projects` split: `unit` (jsdom, existing glob) and `browser` (Chromium via `@vitest/browser-playwright`, `**/*.browser.{test,spec}.*`)
- Registers `mouseDown` / `mouseUp` browser commands so tests can hold `:active` while asserting computed styles
- `src/testing/browserHelpers.ts` — shared fixtures (`render`, `getBg`, `applyStyles`, `resetBrowserTestState`, typed `commands` cast). Excluded from the published lib via `tsconfig.lib.json`
- Scopes the `import/no-extraneous-dependencies` eslint override to `src/testing/**` so helpers can import hoisted dev deps
- `src/runtime/lvfha.browser.test.ts` — 6 tests:
  - sanity transition walk that flips every state (idle → focus → hover → active → unwind)
  - `:hover` over `:focus` in both author orders (verifies bucket routing isn't author-order-dependent)
  - `:active` over `:hover` in both author orders
  - `:active` over both `:focus` and `:hover` while held

## Motivation

jsdom can't dispatch real `:hover` / `:focus` / `:active` state, so cascade behavior — particularly the per-pseudo bucket ordering @griffel/core depends on — has been untestable. This adds an isolated browser project so we can pin the contract for the existing makeStyles bucketing today, and gain a place for `@scope` cascade tests to live in subsequent PRs.

## Stack position

| | Branch |
|---|---|
| **this** | `feat/browser-tests` |
| next | `feat/scope-compile-atomic` |
| | `feat/scope-resolver-wiring` |
| | `feat/scope-hydration-and-tests` |
| | `feat/scope-reset` |

## Test plan

- [ ] `nx test @griffel/core` green
- [ ] `nx lint @griffel/core` green
- [ ] CI green


layershifter:
### Code review

Found 1 issue:

1. `@playwright/test` is added to root `devDependencies` but is never imported anywhere in the repo. `@vitest/browser-playwright` requires `playwright` (also added) as its peer dep, not `@playwright/test`. The `@playwright/test` package pulls in the entire standalone Playwright test runner, which is unused bloat here — drop it and keep only `playwright`.

https://github.com/microsoft/griffel/blob/d142cebc0313dbd9b4265e8058fbf078e44c9880/package.json#L57-L59

layershifter:
AI summary:

Pushed cleanup + follow-up tests on top of a rebase to `main`. Commits in order:

1. **`a30a6af` test(core): cleanup deps, COLORS const, rename helpers**
   - Dropped unused `@playwright/test` from root devDeps (only `playwright` is needed as `@vitest/browser-playwright`'s peer dep).
   - Moved `@vitest/browser-playwright` to root devDeps and removed the entire `devDependencies` block from `packages/core/package.json` — workspace hoisting covers `vitest.config.ts` and the browser test file.
   - Consolidated `WHITE`/`YELLOW`/`CYAN`/`ORANGE` into a single `COLORS` const.
   - Renamed `getBg` → `getComputedBackgroundColor`.
   - Relocated `browserHelpers.ts` from `src/testing/` to `src/common/` (alongside the existing `snapshotSerializers.ts` test helper). Added `src/common/**/*.ts` to `tsconfig.spec.json`'s `include`. Dropped the now-stale `src/testing/**` exclude from `tsconfig.lib.json` and the redundant `**/testing/*` glob from the package's eslint config.

2. **`a5b1f43` test(core): cover :link and :visited in LVFHA cascade**
   - Original tests only exercised the F-H-A part of the cascade despite the `lvfha` name. Added two `:link` runtime tests on an unvisited `<a href="#">`: that `:link` applies by default, and that `:focus` wins over `:link` even when `:focus` is authored first (the Griffel-specific guarantee).
   - Renamed inner `describe('LVHA cascade', ...)` → `describe('LVFHA runtime cascade', ...)` to match the actual scope.

3. **`b6c1854` test(core): drop bucket-ordering assertion** — pulled out a DOM-level check that inspected `<style data-make-styles-bucket>` tags; that's an implementation detail and the runtime tests cover the user-observable behavior. `:visited`'s runtime cascade is left uncovered because browsers privacy-restrict `getComputedStyle` for visited links and there's no portable browser-side way to assert it.

4. **`beb1d92` chore: add changelog entry (none) for @griffel/core browser tests**

Verified locally: `nx lint @griffel/core` 0 errors, `nx type-check @griffel/core` clean, `nx test @griffel/core` 435/435 unit tests pass, `vitest --project=browser` 8/8 browser tests pass.

layershifter (approved):

PR #854, opened 2026-04-28
feat(core): wrap atomic rules in @scope when AtRules.scope is set

joeflateau:
> [!NOTE]
> Stacked on top of #852 and #853. The diff currently includes their content because those PRs haven't merged yet; once they do, this branch will be rebased and the diff will shrink to just the additions below.

## Summary

The first crucial chunk of `@scope` support: `compileAtomicCSSRule` learns to wrap atomic rules in an `@scope (.className) <query> { :scope … }` block when `AtRules.scope` is set. Foundation only — the resolver doesn't yet *set* `atRules.scope`; that wiring is in #PR2.

## What changes

- `src/runtime/utils/types.ts` — adds `scope: string` to the `AtRules` slot. Default initializers across the runtime (and one test fixture) gain `scope: ''` to satisfy the new required field; no logic change
- `src/runtime/compileAtomicCSSRule.ts` — when `atRules.scope` is non-empty:
  - replaces `.className` in the rule body with `:scope` (so pseudos attach to the scope root, matching the @scope spec)
  - wraps with `@scope (.className) <query> { … }`
  - emits a separate scope wrap for RTL so the prelude references the direction-specific class
  - outer at-rules (`@media`, `@layer`, `@supports`, `@container`) continue to wrap around the `@scope` block
- `src/runtime/utils/isScopeSelector.ts` — small `isScopeSelector(property)` utility used by the resolver wiring (#PR2)

## Why explicit prelude + `:scope` substitution

Chromium requires an explicit scope-root prelude for `@scope` to apply reliably. The `:scope` substitution mirrors the spec form (`@scope (.x) { :scope:hover { ... } }`) and avoids re-emitting the class selector inside the body, which keeps the rule legible and matches the proximity-tie-break behavior at the spec level.

## Stack position

| | Branch |
|---|---|
| | `feat/browser-tests` (#853) |
| **this** | `feat/scope-compile-atomic` |
| next | `feat/scope-resolver-wiring` |
| | `feat/scope-hydration-and-tests` |
| | `feat/scope-reset` |

## Test plan

- [ ] `nx test @griffel/core` green (existing tests + the new `compileAtomicCSSRule` and `isScopeSelector` cases)
- [ ] `nx build @griffel/core` green
- [ ] CI green

### DIFF
--- a/change/@griffel-core-ffcddcbb-64d8-470c-a308-514f6c997c17.json
+++ b/change/@griffel-core-ffcddcbb-64d8-470c-a308-514f6c997c17.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "feat(core): wrap atomic rules in @scope when AtRules.scope is set",
+  "packageName": "@griffel/core",
+  "email": "olfedias@microsoft.com",
+  "dependentChangeType": "none"
+}
--- a/packages/core/src/runtime/compileAtomicCSSRule.test.ts
+++ b/packages/core/src/runtime/compileAtomicCSSRule.test.ts
@@ -269,6 +269,134 @@ describe('compileAtomicCSSRule', () => {
       `);
     });
   });
+
+  describe('@scope', () => {
+    it('wraps the rule in @scope and substitutes the class with :scope', () => {
+      expect(
+        compileAtomicCSSRule(
+          {
+            ...defaultOptions,
+            property: 'color',
+            value: 'red',
+          },
+          { ...defaultAtRules, scope: 'to (.boundary)' },
+        ),
+      ).toMatchInlineSnapshot(`
+        [
+          "@scope (.foo) to (.boundary){:scope{color:red;}}",
+        ]
+      `);
+    });
+
+    it('preserves pseudos under @scope', () => {
+      expect(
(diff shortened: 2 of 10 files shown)


layershifter:
Pushed three follow-ups on top of `60cce99`:

- **aff7d0e** `fix(core)`: add missing `scope: ''` to `AtRules` fixtures in `getStyleBucketName.test.ts` / `hashClassName.test.ts` / `hashPropertyKey.test.ts` — the new required field broke type-check there.
- **61f202f** `test(core)`: add 7 inline-snapshot cases for the `@scope` branch (basic, pseudo, RTL, `@media`+`@scope`, `@layer`+`@scope`, descendant `& .baz`, `:hover .nested`).
- **480e6ee** `refactor(core)`: drop the post-hoc `.split('.className').join(':scope')` substitution; instead split `createCSSRule` into a smaller `createCSSDeclaration` (rule body only) and pick the `@scope`-wrapped or class-anchored shape inline at construction. Snapshots unchanged.

`nx test`, `type-check`, `lint` all green for `@griffel/core`.

🤖 AI-generated.

layershifter (approved):

PR #855, opened 2026-04-28
feat(core): wire @scope at-rule into resolveStyleRules

joeflateau:
> [!NOTE]
> Stacked on top of #852, #853, #854. The diff currently includes their content because they haven't merged yet; once they do, this branch will be rebased and the diff will shrink to just the additions below.

## Summary

Hooks `@scope to (SELECTOR)` keys in style objects up to the `compileAtomicCSSRule` machinery added in #854. Recognizes the syntax, validates the prelude, and threads `atRules.scope` through nested recursion so each scope-wrapped declaration retains its pseudo-class bucket (`@scope { :hover { … } }` → `h`, etc.) — preserving LVHA ordering by construction.

## What changes

- `src/runtime/resolveStyleRules.ts`:
  - recognizes `@scope to (...)` keys via `isScopeSelector`
  - asserts the syntax (warning + skip on bare `@scope` or `@scope (...)` without `to`)
  - descends into nested rules with `atRules.scope` set to the trimmed query
- `src/runtime/utils/hashPropertyKey.ts` — adds `scope` slot to the property-hash key so otherwise-identical declarations under different scopes get distinct atomic classes
- Test additions:
  - `resolveStyleRules.test.ts` — extensive @scope coverage (basic, nested, RTL, alongside other at-rules, edge cases)
  - `getStyleBucketName.test.ts` — pinning that `@scope { :hover }` lands in `h` rather than a generic `t` bucket
  - `hashPropertyKey.test.ts` / `hashClassName.test.ts` — distinct classes per scope

## Stack position

| | Branch |
|---|---|
| | `feat/browser-tests` (#853) |
| | `feat/scope-compile-atomic` (#854) |
| **this** | `feat/scope-resolver-wiring` |
| next | `feat/scope-hydration-and-tests` |
| | `feat/scope-reset` |

## Test plan

- [ ] `nx test @griffel/core` green (resolver + bucket + hash tests)
- [ ] CI green


layershifter:
### Code review

Found 1 issue:

1. SSR rehydration is not wired for `@scope`. `AT_RULES_HYDRATOR` enumerates only `media|supports|layer`, and the bucket→regex map only assigns it to `t`/`m`. Scoped atomic rules emitted by `compileAtomicCSSRule` (introduced in #854) land in `d`/`h`/`f`/`a` and look like `@scope (.foo) to (.boundary){:scope{...}}`, which neither `AT_RULES_HYDRATOR` nor `STYLES_HYDRATOR` can match — so SSR-rendered `@scope` rules will be re-inserted on the client (double insertion). Same kind of hydrator update PR #84 did for `@layer`.

https://github.com/microsoft/griffel/blob/43e706497b20505ca474e1f9dcebcf4054425b8e/packages/core/src/renderer/rehydrateRendererCache.ts#L7-L16

layershifter:
Pushed a few follow-ups on top of the original commit (AI-generated, paired with @layershifter):

**Runtime**
- `ca7474a` — warn on nested `@scope` and skip the inner block. The previous code silently overwrote the outer scope when one `@scope` was nested inside another; now `resolveStyleRules` detects `atRules.scope` already being set, logs a dev-only error, and drops the inner rule.

**Tests**
- `3705be3` — moved the `@scope` describe from `resolveStyleRules.test.ts` into a dedicated `resolveStyleRules.scope.test.ts`. Pulled the scope assertions in `getStyleBucketName.test.ts` into their own `it()` block with a comment that `@scope` does not affect bucket strategy. Added a `hashPropertyKey` collision test pinning the scope prefix against `supports` and against an empty scope.
- `0b28ee5` — grouped the syntax-warning cases under a `warnings` describe. Replaced the `Object.keys(classesMap).length` assertions with `toMatchInlineSnapshot` of the classes map so the actual hashes are pinned.
- `a9a7756` — removed redundant single-bucket pseudo tests (already covered by `getStyleBucketName.test.ts`) and the cascade-ordering describe (covered by surrounding cases). Replaced remaining `.d!.length` / `toHaveProperty` assertions with full CSS inline snapshots.

**Chore**
- `63485e3` — replaced `import { RESET, UNSUPPORTED_CSS_PROPERTIES } from '..'` with a direct `../constants.js` import in `resolveStyleRules.test.ts`.
- `7e097cf` — removed the stray top-level `test-css.mjs` scratch file.

layershifter (approved):

PR #933, opened 2026-05-06
chore(eslint): bump ESLint to 10 and migrate to eslint-plugin-react-x

layershifter:
## Summary

`eslint-plugin-react@7.37.5` (latest, last published 2025-04-03) caps its `eslint` peer at `^9.7`, which is the only thing blocking us from going to ESLint 10. Switching to the actively-developed TypeScript-focused fork `eslint-plugin-react-x` removes that ceiling.

### Bumps
- `eslint` `9.39.4` → `10.3.0`
- Add `eslint-plugin-react-x` `5.7.3`
- Add `@eslint/js` `10.0.1` — ESLint 10 no longer bundles it; `@nx/eslint-plugin`'s flat configs require it transitively
- Drop `eslint-plugin-react`

### Per-package config changes (`@griffel/{devtools,react,shadow-dom}`)
- Replace `eslint-plugin-react.configs.flat.{recommended,jsx-runtime}` with `eslint-plugin-react-x.configs['recommended-typescript']`
- Add `eslint-plugin-react-x.configs['disable-conflict-eslint-plugin-react-hooks']` to dedupe rules present in both `react-x` and `react-hooks` v7 (`rules-of-hooks`, `exhaustive-deps`, `set-state-in-effect`, `use-memo`, `purity`, etc.)

### Code adjustments
- `packages/eslint-plugin/src/rules/styles-file.ts` — `context.getFilename()` → `context.filename` (ESLint 10 dropped the deprecated getter)
- `packages/jest-serializer/src/index.ts` — drop dead `let result = null` (`no-useless-assignment`, new ESLint 10 default rule)
- `packages/core/src/devtools/react-render-tracker/stackTrace.ts` — file-level disable for `no-useless-assignment` (verbatim port from upstream)
- `packages/react/src/RendererContext.tsx` — extend inline disable to cover `react-x/use-memo` alongside `react-hooks/void-use-memo`
- `packages/react/src/insertionFactory.ts` — rename inline rules-of-hooks disable from `react-hooks/*` to `react-x/*`

### Stacked on #872
Targets `dependabot/npm_and_yarn/eslint-5bae9fd6e4` so the diff stays scoped to this work. Will retarget to main once #872 lands.

## Test plan
- [x] `nx run-many --target=lint --all` — clean
- [x] `nx run-many --target=type-check --all` — clean
- [x] `nx run-many --target=test --all` — passes


layershifter:
Folded into #872

PR #953, opened 2026-05-19
fix(webpack-extraction-plugin): rspack support — skip cacheGroup + file-based CSS transport

BrunoRDS:
## Summary

Two related fixes that together let `@griffel/webpack-extraction-plugin` work end-to-end on rspack with `CssExtractRspackPlugin` and produce a fully styled app.

### 1. Disable Griffel `splitChunks` cacheGroup on rspack

On rspack with `CssExtractRspackPlugin`, when the `cacheGroups['griffel']` test matches Griffel CSS modules (either via the function `isGriffelCSSModule` or the regex `/griffel\.css/`), rspack creates a `griffel` chunk that gets named with the JS filename template (`griffel-<hash>.js`) and `CssExtractRspackPlugin` does NOT emit a corresponding `griffel-<hash>.css`.

Net effect on rspack: all matched Griffel CSS modules are routed into a chunk whose CSS output is never written — dropping ~10 MB of CSS silently and leaving the app shell unstyled.

Until rspack/`CssExtractRspackPlugin` handles renamed/merged CSS chunks the way webpack/`MiniCssExtractPlugin` does, the safest behavior on rspack is to skip registering the cacheGroup entirely. Per-source-file `.griffel.css` files remain intact; the CSS rule sort-order optimization (which only runs against the merged `griffel.css` asset) is forgone on rspack.

### 2. File-based CSS transport on rspack (`?file=` instead of `?style=`)

The previous inline transport encoded the entire CSS payload into the loader request as `?style=<url-encoded css>`. For large `.styles.ts` files the encoded request exceeds Linux's `PATH_MAX` (4096 bytes), which broke BuildXL pip access tracking on Linux CI.

The new transport writes CSS to a content-hashed tempfile under `os.tmpdir()/griffel-css-cache/<sha1>.css` (atomic via staging file + rename, so concurrent loader invocations are safe) and references it via `?file=<encoded path>`. The virtual loader now prefers `?file=` when present and falls back to `?style=` for backwards compatibility.

Additionally, the rspack loader request now uses the source file as the resource (`...${virtualLoaderPath}!${this.resourcePath}?file=...`) instead of the virtual `griffel.css` placeholder, so `css-loader` resolves `url()` references relative to the source file's directory rather than to the plugin's own `node_modules/virtual-loader/` path.

## Webpack behavior

Unchanged — both new code paths are gated by `IS_RSPACK` / `!IS_RSPACK`. The existing test suite (webpack-only fixtures) should continue to pass.

## Verification

Verified end-to-end on Microsoft Teams' `react-web-client` experience with rspack 1.7.6 + `CssExtractRspackPlugin`:

- Before: 304 KB of CSS across 25 chunks; app shell unstyled.
- After: 12.3 MB of CSS across 529 chunks; app shell renders fully styled.
- BuildXL Linux CI builds succeed (previously failed with `DX14304` PATH_MAX warnings and `DX0064` pip failure).

## Test plan

- [ ] Existing webpack fixtures pass unchanged (`config-split-chunks`, `config-no-split-chunks`, etc.).
- [ ] Manual rspack smoke test on a large styles-heavy app produces emitted CSS files and a styled app.
- [ ] Linux CI green on this branch.

### DIFF
--- a/change/@griffel-webpack-extraction-plugin-460d5734-7faf-4eb7-9069-43390beaad21.json
+++ b/change/@griffel-webpack-extraction-plugin-460d5734-7faf-4eb7-9069-43390beaad21.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix(webpack-extraction-plugin): rspack support — skip Griffel splitChunks cacheGroup on rspack (CssExtractRspackPlugin does not emit CSS for renamed chunks) and switch the rspack loader's CSS transport from inline ?style= (exceeds Linux PATH_MAX on large styles files) to a content-hashed tempfile via ?file=.",
+  "packageName": "@griffel/webpack-extraction-plugin",
+  "email": "brunoru@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
+++ b/packages/webpack-extraction-plugin/src/GriffelCSSExtractionPlugin.ts
@@ -150,14 +150,21 @@ export class GriffelCSSExtractionPlugin {
     // WHY?
     //  We need to sort CSS rules in the same order as it's done via style buckets. It's not possible in multiple
     //  chunks.
-    if (compiler.options.optimization.splitChunks) {
+    // @ Rspack compat:
+    //  Tested with rspack 1.7.6 + CssExtractRspackPlugin: when the cacheGroup test matches Griffel CSS modules
+    //  (either via the function form `isGriffelCSSModule` or the regex form `/griffel\.css/`), the resulting
+    //  `griffel` chunk is named with the JS chunk filename template (`griffel-<hash>.js`) and
+    //  CssExtractRspackPlugin does NOT emit a corresponding `griffel-<hash>.css`. Net effect: all matched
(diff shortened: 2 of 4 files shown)


BrunoRDS:
Superseded by #958 + #960 (both merged), which land the same fix more thoroughly upstream:

- **#958** ("fix(rspack): make webpack-extraction-plugin + webpack-plugin work with CssExtractRspackPlugin") collapses the rspack and webpack code paths almost entirely — uses the same `outputFileName!=!virtualLoader!source` request format for both bundlers (which also kills the `?style=<encoded css>` PATH_MAX issue this PR was trying to address) and fixes the cacheGroup `test` regex by switching to a `type: /^css/` pre-filter + identifier-based function test. New `legacy-css-extract-rspack-1` e2e scenario was added that exactly mirrors the failing production setup.
- **#960** ("fix(rspack): keep CSS alive under consumer 'sideEffects': false") adds the `{ test: /\.griffel\.css$/, sideEffects: true }` rule that this PR's downstream consumer was patching manually in their build config.

Diagnosis in #953 was correct (cacheGroup not matching + PATH_MAX on inline CSS transport) but the right fix is closer to the source than this PR proposed. Closing in favor of @layershifter's upstream fixes; will upgrade the downstream consumer to the released versions.

Thanks @layershifter for the fast turnaround!

PR #962, opened 2026-05-21
chore(e2e): use explicit workspace deps for internal @griffel/* packages

layershifter:
## Summary

First slice of the larger workspaces + project-references migration explored in #774. This PR is intentionally minimal — it only touches **private** packages so it can land independently:

- `e2e/eslint`, `e2e/typescript`: add `@griffel/e2e-utils` as `devDependency` (`workspace:*`). Both already import from `@griffel/e2e-utils`; the resolution was implicit via the path alias in `tsconfig.base.json`.
- `e2e/rspack`: tighten the existing `"*"` `e2e-utils` dep to `workspace:*` for consistency.
- Drop `@griffel/e2e-utils` and `@griffel/update-shorthands` aliases from `tsconfig.base.json` — no consumer needs them now that the deps are declared explicitly.

Published `@griffel/*` packages keep their path aliases untouched; those migrate in follow-up PRs.

## Test plan

- [x] `yarn nx run-many --target=type-check --all` — passes (20/20)
- [x] `yarn nx run @griffel/e2e-eslint:test` — passes
- [x] `yarn nx run @griffel/e2e-typescript:test` — passes
- [x] `yarn check-dependencies` — clean

### DIFF
--- a/e2e/eslint/package.json
+++ b/e2e/eslint/package.json
@@ -2,5 +2,8 @@
   "name": "@griffel/e2e-eslint",
   "version": "0.0.0",
   "private": true,
-  "type": "module"
+  "type": "module",
+  "devDependencies": {
+    "@griffel/e2e-utils": "workspace:*"
+  }
 }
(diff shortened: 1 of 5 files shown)


layershifter:
Slicing the migration into phases doesn't pan out — closing in favor of the single migration in #774.

PR #963, opened 2026-05-21
chore(style-types): migrate to per-package dist/ build pipeline

layershifter:
## Summary

Phase 2 of the workspaces + project-references migration (stacks on top of #962). Picks the smallest leaf package and replaces the workspace-rooted `@nx/js:tsc` build with a per-package `tsc` + asset-copy pipeline.

- Build output: `packages/style-types/dist/` (was `dist/packages/style-types`).
- New `tools/copy-pkg-assets.mjs` copies `package.json` (stripped of `scripts` + `devDependencies`), `README.md`, and `LICENSE.md` into the per-package `dist/`.
- `package.json` gets a published-style `exports` map: `@griffel/source` (no-op today, forward-compat with the `customConditions` in later phases), `types` pointing at `src/index.ts` so consumers can resolve types without first building, `default` at `dist/index.js` for runtime, and a `files: ["dist/", README, LICENSE, CHANGELOG]` field.
- Drop `@griffel/style-types` from `tsconfig.base.json` paths — consumers resolve via the yarn workspace symlink + the package.json above.
- `.gitignore`: drop the leading slash from `dist` so per-package `dist/` directories are also ignored.

## Test plan

- [x] `yarn nx run-many --target=type-check --all` — 20/20 pass
- [x] `yarn nx run-many --target=build --all` — 15/15 pass
- [x] `yarn nx run-many --target=test --all` — 16/16 pass (excluding the pre-existing e2e/rspack flake)
- [x] `yarn check-dependencies` — clean

## Stack

This PR builds on #962 — the diff above #962 is the style-types-only delta. Don't merge until #962 lands.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 # See http://help.github.com/ignore-files/ for more about ignoring files.
 
 # compiled output
-/dist
+dist
 /tmp
 /out-tsc
 /packages/*/dist/bundle-size
--- a/e2e/eslint/package.json
+++ b/e2e/eslint/package.json
@@ -2,5 +2,8 @@
   "name": "@griffel/e2e-eslint",
   "version": "0.0.0",
   "private": true,
-  "type": "module"
+  "type": "module",
+  "devDependencies": {
+    "@griffel/e2e-utils": "workspace:*"
+  }
 }
--- a/e2e/rspack/package.json
+++ b/e2e/rspack/package.json
@@ -5,7 +5,7 @@
   "type": "module",
   "sideEffects": false,
   "dependencies": {
-    "@griffel/e2e-utils": "*",
+    "@griffel/e2e-utils": "workspace:*",
     "@griffel/react": "^1.7.4",
     "@griffel/webpack-extraction-plugin": "^0.5.19",
     "@griffel/webpack-loader": "^2.2.28",
--- a/e2e/typescript/package.json
+++ b/e2e/typescript/package.json
@@ -2,5 +2,8 @@
   "name": "@griffel/e2e-typescript",
   "version": "0.0.0",
   "private": true,
-  "type": "module"
+  "type": "module",
+  "devDependencies": {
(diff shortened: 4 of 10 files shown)


layershifter:
### Code review

Found 1 issue:

1. Published tarball will ship a broken `types` resolution. The outer `packages/style-types/package.json` declares `"types": "./src/index.ts"` (and the `@griffel/source` / `types` export conditions also point at `./src/index.ts`), but `files` only includes `dist/`, `README.md`, `LICENSE.md`, `CHANGELOG.md`. Beachball publishes from the package root using the outer `package.json`, so the tarball will not contain any `src/`. Verified locally with `npm pack --dry-run`: the tarball contains `dist/index.d.ts` but no `src/`, while the included `package.json` still points `types` at `./src/index.ts`. Every downstream consumer will get "cannot find type definitions" once this ships.

https://github.com/microsoft/griffel/blob/4314a61ecebb1009414d28ec69b2fb3693ef6b59/packages/style-types/package.json#L10-L27

Fix options: point `types` and the `types` / non-source export conditions at the emitted `./dist/index.d.ts` (declarations are already produced by `tsconfig.lib.json`), or add `src/` to `files` if the goal is to keep ship-the-source semantics for `@griffel/source` consumers. The `dist/package.json` written by `tools/copy-pkg-assets.mjs` is also not used at publish time (npm reads the outer one), so fixing the outer file is what matters.

Other things considered but not flagged: the dead `dist/package.json` (harmless because beachball doesn't publish from `dist/`), the `@griffel/update-shorthands` path-mapping removal (no callers found), the `.gitignore` `/dist` → `dist` change (no tracked `dist/*` files), and the `cwd` vs documented usage of `copy-pkg-assets.mjs` (functionally works since the script resolves against repoRoot).

layershifter:
Slicing the migration into phases doesn't pan out — closing in favor of the single migration in #774.

PR #995, opened 2026-06-20
docs(react): document the Next.js App Router SSR setup

vashek:
The SSR guide only covered the Pages Router (_document, one-shot). Add an App Router section using useServerInsertedHTML, and explain why the renderer must be cleared after each streaming flush: renderToStyleElements returns the renderer's entire accumulated CSS on every call, so without clearing, Next duplicates the full stylesheet into <body> on later flushes. Those stale copies override head-inserted styles after a client-side navigation, making makeStyles overrides lose to their makeResetStyles base. Also clarifies the existing heading as 'Next.js (Pages Router)'.

### DIFF
--- a/apps/website/docs/react/guides/ssr-usage.md
+++ b/apps/website/docs/react/guides/ssr-usage.md
@@ -6,7 +6,7 @@ sidebar_position: 4
 
 Griffel provides first class support for Server-Side Rendering.
 
-## Next.js
+## Next.js (Pages Router)
 
 ### Base setup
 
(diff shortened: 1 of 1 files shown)


vashek:

apps/website/docs/react/guides/ssr-usage.md
```diff
@@ -99,3 +99,68 @@ export default function Home() {
...
+import { useState } from 'react';
+
+export function GriffelRegistry({ children }: { children: React.ReactNode }) {
+  // 👇 one renderer per request, kept stable across re-renders
+  const [renderer] = useState(() => createDOMRenderer());
+
+  useServerInsertedHTML(() => {
+    const styles = renderToStyleElements(renderer);
+
+    // highlight-start
```
layershifter, line 125: @vashek Can you please check if code below works as expected?

```
  const [renderer] = React.useState(() => createDOMRenderer());
  const didRenderRef = React.useRef(false);

  useServerInsertedHTML(() => {
    if (didRenderRef.current) {
      return;
    }
    didRenderRef.current = true;
    return <>{renderToStyleElements(renderer)}</>;
  });
```

https://storybooks.fluentui.dev/react/?path=/docs/concepts-developer-server-side-rendering-next-js-appdir-setup--docs

---

If so, let's use it instead as it does not touch internals. Thanks

layershifter (approved):

PR #996, opened 2026-06-20
feat(core): warn in dev when SSR styles are rehydrated more than once

vashek:
renderToStyleElements returns the renderer's entire accumulated CSS on every call. In the Next.js App Router, useServerInsertedHTML runs once per streaming flush, so a setup that does not clear the renderer between flushes re-emits the full stylesheet and the copies are streamed into <body>. Those stale copies override styles inserted after a client-side navigation, silently breaking makeStyles overrides (they lose to their makeResetStyles base). The failure looks correct on a hard reload and only appears after a soft navigation, which makes it very hard to diagnose.

rehydrateRendererCache scans every server-rendered <style> element, so it is the natural place to detect this: in development, count CSS rules that appear in more than one element (duplicate rule text, not merely a repeated bucket — a bucket legitimately recurs across flushes with different rules) and emit one console.error pointing at the renderer.stylesheets reset fix. Zero production overhead. Also DRYs the previously thrice-repeated devtools guard into a cacheRule helper.

### DIFF
--- a/change/@griffel-core/warn-duplicate-rehydration-styles-20260620155350.json
+++ b/change/@griffel-core/warn-duplicate-rehydration-styles-20260620155350.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Warn in development when duplicate CSS rules are found during rehydration — a signal that server-rendered styles were flushed into the HTML more than once (e.g. the Next.js App Router calling renderToStyleElements on every flush without clearing the renderer)",
+  "packageName": "@griffel/core",
+  "email": "vaclav.dvorak@ysoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@griffel-core/warn-duplicate-rehydration-styles-20260620155350.json
+++ b/change/@griffel-core/warn-duplicate-rehydration-styles-20260620155350.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Warn in development when duplicate CSS rules are found during rehydration — a signal that server-rendered styles were flushed into the HTML more than once (e.g. the Next.js App Router calling renderToStyleElements on every flush without clearing the renderer)",
+  "comment": "chore: add development warnings for hydration issues in `rehydrateRendererCache`",
   "packageName": "@griffel/core",
   "email": "vaclav.dvorak@ysoft.com",
   "dependentChangeType": "patch"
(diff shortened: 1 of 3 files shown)


vashek:

packages/core/src/renderer/rehydrateRendererCache.ts
```diff
@@ -60,25 +79,40 @@ export function rehydrateRendererCache(
...
+          'In the Next.js App Router this happens when renderToStyleElements() is called on every',
+          'useServerInsertedHTML flush without clearing the renderer in between: each flush re-emits',
+          'the full stylesheet and the extra copies are streamed into <body>. Those stale copies can',
+          'override styles inserted after a client-side navigation, making makeStyles() overrides lose',
+          'to their makeResetStyles() base.',
+          '\n\n',
+          'Clear the renderer after each flush:',
+          '\n\n',
+          'useServerInsertedHTML(() => {\n  const styles = renderToStyleElements(renderer);\n  renderer.stylesheets = {};\n  return styles;\n});',
+          '\n\n',
```
layershifter:
```suggestion
```

Let's remove the actual recommendation from code.

packages/core/src/renderer/rehydrateRendererCache.ts
```diff
@@ -29,6 +29,29 @@ export function rehydrateRendererCache(
...
+    let duplicateRuleCount = 0;
+
+    const cacheRule = (cssRule: string, bucketName: StyleBucketName) => {
+      if (seenRules) {
+        if (seenRules.has(cssRule)) {
+          duplicateRuleCount++;
+        } else {
+          seenRules.add(cssRule);
+        }
+      }
```
layershifter:
```suggestion
    const seenRules: Set<string> | undefined = new Set();
    let duplicateRuleCount = 0;

    const cacheRule = (cssRule: string, bucketName: StyleBucketName) => {
      if (process.env.NODE_ENV !== 'production') {
        if (seenRules.has(cssRule)) {
          duplicateRuleCount++;
        } else {
          seenRules.add(cssRule);
        }
      }
```

nit: let's write it this way. While Terser is able to DCE the original code, I would prefer to keep instructions gated in more solid way. Thx!

change/@griffel-core/warn-duplicate-rehydration-styles-20260620155350.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Warn in development when duplicate CSS rules are found during rehydration — a signal that server-rendered styles were flushed into the HTML more than once (e.g. the Next.js App Router calling renderToStyleElements on every flush without clearing the renderer)",
```
layershifter:
```suggestion
  "comment": "chore: add development warnings for hydration issues in `rehydrateRendererCache`",
```

layershifter:
@vashek Can you please check comments? Thx!

layershifter (approved):

PR #1023, opened 2026-07-15
fix(sorting): keep @container rules after regular styles and @media

Yakubko:
### Sorting fix
We noticed @container rules were being emitted before regular styles, so the container styles got overridden and effectively ignored. Rules are now sorted bucket-first everywhere they're emitted, keeping @media / @container grouped and always after regular styles:

- @griffel/react (SSR renderToStyleElements) and @griffel/webpack-extraction-plugin / @griffel/webpack-plugin (sortCSSRules) now order by bucket first, then by the respective query comparator within the @media / @container buckets.
- @griffel/sort-css-queries: rewrote compareCSSQueries into a total order (bucket → axis → numeric bound), added min/max-height and and-band support, and a createCompareCSSQueries({ rootFontSize }) factory for rem/em conversion. Never throws.

**Tests & changesets**
Expanded coverage for the comparator, SSR ordering, and both webpack sortCSSRules. Beachball patch change files added for all five affected packages.

### DIFF
--- a/packages/react/src/renderToStyleElements.ts
+++ b/packages/react/src/renderToStyleElements.ts
@@ -12,14 +12,9 @@ import type { GriffelRenderer } from '@griffel/core';
  */
 export function renderToStyleElements(renderer: GriffelRenderer): ReactElement[] {
   const stylesheets = Object.values(renderer.stylesheets).sort((a, b) => {
-    // Primary: bucket order. This keeps "@media" / "@container" sheets grouped, separated from each
-    // other, and always placed after regular styles before ordering within a bucket by its condition.
-    // It must come first: a user-supplied comparator only understands its own condition strings and
-    // can't be trusted to order empty/other-bucket values, so gating by bucket avoids scattering
-    // "@container" sheets throughout the output.
-    const bucketDiff = styleBucketOrdering.indexOf(a.bucketName) - styleBucketOrdering.indexOf(b.bucketName);
-    if (bucketDiff !== 0) {
-      return bucketDiff;
+    const bucketNameDiff = styleBucketOrdering.indexOf(a.bucketName) - styleBucketOrdering.indexOf(b.bucketName);
+    if (bucketNameDiff !== 0) {
+      return bucketNameDiff;
     }
 
     // Within the "@media" bucket, order by media query.
--- a/packages/react/src/renderToStyleElements.ts
+++ b/packages/react/src/renderToStyleElements.ts
@@ -12,15 +12,39 @@ import type { GriffelRenderer } from '@griffel/core';
  */
 export function renderToStyleElements(renderer: GriffelRenderer): ReactElement[] {
   const stylesheets = Object.values(renderer.stylesheets).sort((a, b) => {
-    return (
-      renderer.compareContainerQueries(
+    // Primary: bucket order. This keeps "@media" / "@container" sheets grouped, separated from each
+    // other, and always placed after regular styles before ordering within a bucket by its condition.
+    // It must come first: a user-supplied comparator only understands its own condition strings and
+    // can't be trusted to order empty/other-bucket values, so gating by bucket avoids scattering
+    // "@container" sheets throughout the output.
+    const bucketDiff = styleBucketOrdering.indexOf(a.bucketName) - styleBucketOrdering.indexOf(b.bucketName);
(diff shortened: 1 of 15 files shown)


packages/webpack-plugin/src/utils/sortCSSRules.mts
```diff
@@ -58,12 +58,33 @@ function compareCSSRules(
...
-    compareContainerQueries(a.container, b.container) ||
-    compareMediaQueries(a.media, b.media) ||
-    styleBucketOrderingMap[a.styleBucketName] - styleBucketOrderingMap[b.styleBucketName] ||
-    a.priority - b.priority
-  );
+  // Primary: bucket order. This keeps "@media" / "@container" rules grouped, separated from each
+  // other, and always placed after regular styles before ordering within a bucket by its condition.
+  // It must come first: a user-supplied comparator only understands its own condition strings and
+  // can't be trusted to order empty/other-bucket values, so gating by bucket avoids scattering
+  // "@container" rules throughout the output.
```
bsunderhus:
```suggestion
```
Let's drop the comments

packages/webpack-plugin/src/utils/sortCSSRules.mts
```diff
@@ -58,12 +58,33 @@ function compareCSSRules(
...
+  const bucketDiff = styleBucketOrderingMap[a.styleBucketName] - styleBucketOrderingMap[b.styleBucketName];
```
bsunderhus: nit:

```suggestion
  const nameDiff = styleBucketOrderingMap[a.styleBucketName] - styleBucketOrderingMap[b.styleBucketName];
```

packages/react/src/renderToStyleElements.ts
```diff
@@ -12,15 +12,39 @@ import type { GriffelRenderer } from '@griffel/core';
  */
 export function renderToStyleElements(renderer: GriffelRenderer): ReactElement[] {
   const stylesheets = Object.values(renderer.stylesheets).sort((a, b) => {
-    return (
-      renderer.compareContainerQueries(
+    // Primary: bucket order. This keeps "@media" / "@container" sheets grouped, separated from each
+    // other, and always placed after regular styles before ordering within a bucket by its condition.
+    // It must come first: a user-supplied comparator only understands its own condition strings and
+    // can't be trusted to order empty/other-bucket values, so gating by bucket avoids scattering
+    // "@container" sheets throughout the output.
```
bsunderhus:
```suggestion
```
Drop the commens

packages/react/src/renderToStyleElements.ts
```diff
@@ -12,15 +12,39 @@ import type { GriffelRenderer } from '@griffel/core';
...
+    const bucketDiff = styleBucketOrdering.indexOf(a.bucketName) - styleBucketOrdering.indexOf(b.bucketName);
```
bsunderhus: nit: 
```suggestion
    const nameDiff= styleBucketOrdering.indexOf(a.bucketName) - styleBucketOrdering.indexOf(b.bucketName);
```

bsunderhus (commented):
You're stating:
> Fix @container / https://github.com/media ordering **&** make @griffel/transform tolerant of non-authoring input

Can those 2 changes be 2 different PRs? one for the container / media ordering, and another for the tolerance of the non-authoring input

packages/webpack-plugin/src/utils/sortCSSRules.mts
Yakubko: removed

packages/webpack-plugin/src/utils/sortCSSRules.mts
Yakubko: Changed to bucketNameDiff

packages/react/src/renderToStyleElements.ts
Yakubko: removed

packages/react/src/renderToStyleElements.ts
Yakubko: Changed to bucketNameDiff

packages/sort-css-queries/src/compareCSSQueries.ts
bsunderhus, line 1: Let's drop the excessive comments

bsunderhus (approved):

PR #1026, opened 2026-07-16
fix(transform): skip CommonJS files and multi-arg runtime calls inste…

Yakubko:
### @griffel/transform: skip non-authoring input instead of throwing
- CommonJS (non-ESM) files now pass through untouched instead of throwing, restoring pre-3.0.x behavior so CJS dependencies (e.g. @fluentui/react lib-commonjs) can flow through the loader.
- Multi-argument Griffel calls — the lower-level @griffel/core runtime form such as makeStyles(styles, insertionFactory) used internally by @griffel/react — are skipped instead of throwing, since they aren't the authoring API this transform rewrites.

Beachball patch change file added for @griffel/transform.

### DIFF
--- a/packages/transform/src/transformSync.mts
+++ b/packages/transform/src/transformSync.mts
@@ -265,9 +265,6 @@ export function transformSync(sourceCode: string, options: TransformOptions): Tr
         const functionKind = importedName as FunctionKinds;
 
         if (node.arguments.length !== 1) {
-          // Multi-argument calls are the lower-level @griffel/core runtime form
-          // (e.g. makeStyles(styles, insertionFactory)), not the authoring API
-          // this transform rewrites, so skip them.
           return;
         }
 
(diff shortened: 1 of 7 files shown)


packages/transform/src/transformSync.mts
```diff
@@ -265,9 +265,10 @@ export function transformSync(sourceCode: string, options: TransformOptions): Tr
         const functionKind = importedName as FunctionKinds;
 
         if (node.arguments.length !== 1) {
-          throw new Error(
-            `${functionKind}() function accepts only a single param, got ${node.arguments.length} in ${filename}`,
-          );
+          // Multi-argument calls are the lower-level @griffel/core runtime form
+          // (e.g. makeStyles(styles, insertionFactory)), not the authoring API
+          // this transform rewrites, so skip them.
```
bsunderhus:
```suggestion
```
Drop the comments

bsunderhus (approved):

PR #1067, opened 2026-07-30
refactor: share the transform resolver via `@griffel/css-extraction-utils`

layershifter:
## Overview

Follow up to #1064, which shared the CSS extraction utils between the Webpack & Vite plugins but left one duplicate behind.

`packages/webpack-plugin/src/resolver/` and `packages/vite-plugin/src/resolver/` held near identical copies of the resolver used for imports inside modules that are evaluated during the transform — ~35 of 58 lines were the same. Only the factory argument differed.

## Implementation

The shared core lands as `createTransformResolver()` in `@griffel/css-extraction-utils`:

```ts
export function createTransformResolver(options?: { alias?: NapiResolveOptions['alias'] }): TransformResolver;
```

It owns what was duplicated: the CJS-only package list (`tslib`, `@babel/runtime`, `@swc/helpers`), `RESOLVE_OPTIONS_DEFAULTS`, the CJS & ESM resolver pair and the `resolveModule` closure.

Each plugin keeps a thin wrapper, so **no public API changes**:

| Plugin | Public factory type | Unchanged |
| --- | --- | --- |
| `@griffel/webpack-plugin` | `(compilation: Compilation) => TransformResolver` | ✅ |
| `@griffel/vite-plugin` | `(options: ResolverFactoryOptions) => TransformResolver` | ✅ |

`ResolverFactoryOptions` is now an alias of the shared `TransformResolverOptions`, and both plugins still export `createResolverFactory` from their entry points. The Webpack wrapper keeps its commented out `resolveOptionsFromWebpackConfig` note, that decision is untouched here.

`oxc-resolver` is dropped from both plugins' dependencies — neither imports it anymore.

## Test plan

`createResolverFactory.test.mts` was the only suite covering this code and lived in `@griffel/webpack-plugin`. It is `git mv`'d to `@griffel/css-extraction-utils` and retargeted at the shared function, so the Webpack plugin no longer needs a `webpack` type import just to build a stub `Compilation`.

Two cases are added for `alias`. It was implemented in the Vite copy with **no coverage at all**, so this is net new:

- an aliased specifier resolves to the aliased path
- a non aliased specifier is unaffected

```sh
yarn nx run @griffel/css-extraction-utils:test
```

31 tests pass (8 in the resolver suite). `lint`, `type-check`, `build`, `test`, `prettier` and `syncpack` are green across all 24 projects.

### DIFF
--- a/change/@griffel-css-extraction-utils-shared-resolver.json
+++ b/change/@griffel-css-extraction-utils-shared-resolver.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "feat: add \"createTransformResolver()\" shared by bundler integrations",
+  "packageName": "@griffel/css-extraction-utils",
+  "email": "olfedias@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@griffel-vite-plugin-shared-resolver.json
+++ b/change/@griffel-vite-plugin-shared-resolver.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "chore: use the resolver shared via \"@griffel/css-extraction-utils\"",
+  "packageName": "@griffel/vite-plugin",
+  "email": "olfedias@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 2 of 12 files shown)


layershifter:
Abandoning this for now — the duplication between the two plugins' resolver factories isn't worth the extra shared surface at this point.

PR #1082, opened 2026-08-17
Added support for precompiled styles

tomas-vodrazka:


layershifter:
Heads up. You're likely don't want to do this, instead you would to move to unprocessed styles (https://storybooks.fluentui.dev/react/?path=/docs/concepts-developer-unprocessed-styles--docs) and work with partners to follow best practices.

